PR #49, opened 2018-03-12
Always use `conda info` to find prefix instead of environment

### LINKED ISSUE #48: Conda environment detection with CONDA_PREFIX is unreliable
We added a workaround to discover the conda environment when `$CONDA_PREFIX` is unavailable in:

https://github.com/ContinuumIO/intake/commit/73be37e2f86ffcc5f9a91cd4e81658c8550d1427

However, we've seen a few cases where the `$CONDA_PREFIX` has been set incorrectly, so we should probably only use the approach in the above commit.

martindurant:
Fixes #48

### DIFF
--- a/intake/catalog/default.py
+++ b/intake/catalog/default.py
@@ -51,9 +51,7 @@ def conda_prefix():
 def global_data_dir():
     """Return the global Intake catalog dir for the current environment"""
 
-    if CONDA_VAR in os.environ:
-        prefix = os.environ[CONDA_VAR]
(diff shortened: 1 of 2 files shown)


martindurant:
Still prefers virtualenv's environment variable, if it's set.

martindurant:
(test fail, because they rely on setting CONDA_PREFIX in the in-process environment explicitly)

philippjfr:
Can confirm this change works for me consistently.

martindurant:
OK!

PR #57, opened 2018-03-21
[WIP] xarray container

mmccarty:


intake/container/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+from .semistructured import SemiStructured
+
+container_map = {
+    'dataframe' : DataFrame,
+    'python' : SemiStructured,
+    'ndarray' : NdArray
+}
+
+def get_container_klass(container):
+    return container_map.get(container, BaseContainer)
```
mmccarty: Using a mapper here for the time being so we don't have to change the interface.

intake/container/__init__.py
```diff
@@ -0,0 +1,16 @@
+from .base import BaseContainer
+from .dataframe import DataFrame
+from .ndarray import NdArray
+from .semistructured import SemiStructured
+
+container_map = {
+    'dataframe' : DataFrame,
+    'python' : SemiStructured,
```
martindurant: This name is misleading: it is a sequence of arbitrary python objects; in the common case those objects are homogeneous, probably dicts, but that is not a requirement.

intake/container/__init__.py
martindurant: Totally fine with that. Later we may provide a "register" API, function or decorator.

intake/container/__init__.py
mmccarty: Yeah, I agree.  I struggled with the name here.  I certainly didn't want to use `Python`.  SemiStructure came from the docs.  In either case, the naming should be consistent.

mmccarty:
:+1: for integration test repo

PR #60, opened 2018-03-29
Style

martindurant:
Updates for PEP8 and other small things.

### DIFF
--- a/intake/catalog/default.py
+++ b/intake/catalog/default.py
@@ -6,17 +6,11 @@ import subprocess
 from .base import Catalog
 
 
-def make_empty_catalog():
-    # TODO: replace with Catalog() (or None or []) and implement in the constructor
-    empty = os.path.join(os.path.dirname(__file__), 'empty.yml')
-    return Catalog(empty)
-
-
 def load_user_catalog():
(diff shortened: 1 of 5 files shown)


martindurant:
(the kind of editing I can do in a conference room while jet-lagged)

intake/catalog/default.py
```diff
@@ -6,17 +6,11 @@
...
-    empty = os.path.join(os.path.dirname(__file__), 'empty.yml')
-    return Catalog(empty)
-
-
 def load_user_catalog():
     """Return a catalog for the platform-specific user Intake directory"""
     cat_dir = user_data_dir()
     if not os.path.isdir(cat_dir):
-        return make_empty_catalog()
+        return Catalog()
```
martindurant: Note that I'm thinking that `Catalog(cat_dir)` should error if the directory doesn't exist, but not if it doesn't contain any valid cat files. The "combo" Catalog should not error if any of the directories doesn't exist, so may end up as an empty cat, without the need to explicitly check the directories.

martindurant: Likewise, `Catalog(afile)` *should* error if the file doesn't exist.

PR #61, opened 2018-03-29
RFC: Sub cats

martindurant:
A simple implementation of catalogues that can point to other catalogs, and make them available as attributes for tab-completion. This allows for "index" cats, which can point to a combination of local and remote catalogs, and for hierarchical sources, rather than the flattening that happens when loading a directory of yamls. Also opens the door to data source engines, such as "all the known tables in postgresql" and other data providers.

Note that this implementation is not recommended as the final way to do it, just put here for conversation.

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -559,7 +559,9 @@ class CatalogParser(object):
 
         return dict(
             plugin_sources=self._parse_plugins(data),
-            data_sources=self._parse_data_sources(data))
+            data_sources=self._parse_data_sources(data),
+            catalogs=data.get('catalogs', {})
+        )
 
 
 class CatalogConfig(object):
@@ -605,6 +607,11 @@ class CatalogConfig(object):
             entry.find_plugin(self._plugins)
             self._entries[entry.name] = entry
 
+        self._cats = {}
+        for entry in cfg['catalogs']:
+            from intake import Catalog
+            self._cats[entry['name']] = Catalog(entry['url'])
+
     @property
     def name(self):
         return self._name
@@ -616,3 +623,7 @@ class CatalogConfig(object):
     @property
     def entries(self):
         return self._entries
+
+    @property
+    def cats(self):
+        return self._cats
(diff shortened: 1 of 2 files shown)


martindurant:
Suggesting that when a cat refers to another cat, failure to load it should not cause an error, i.e., errors in subcat should be swallowed, perhaps with a warning. 
Note that there is no check for circular references, which would presumably cause infinite recursion.

intake/catalog/local.py
```diff
@@ -578,3 +623,7 @@ def plugins(self):
     @property
     def entries(self):
         return self._entries
+
+    @property
+    def cats(self):
```
mmccarty: Can we just call them `catalogs`?

martindurant: Sure! I was thinking that the PR here gives a possible outline, but not actually the right way to do things, so I wanted a conversation around this.

mmccarty:
Another thought.  Would we ever want sub catalogs to have a reference to their parent?

martindurant:
> sub catalogs to have a reference to their parent

Maybe? Not sure what you'd do with it. I think I noted above that my sketch here perfectly allows circular cat references, which would be bad.

intake/catalog/local.py
martindurant: Done

mmccarty:
@martindurant Is there anything left to do here?  A apart from resolve conflicts.

martindurant:
Yes, I feel like what I've done here is too hacky to be the final approach. I'm not yet sure of the "right" way to do it, but could use eyes on it. I suspect that it would be better to remove the "state" objects and have each catalogue define its own refresh behaviour directly.

You are thinking of having a SQL-table factory in a catalogue?

martindurant:
You might argue that the issue around state objects is orthogonal, but the implementation here only allows for calling `Catalog()` with some string, i.e., a YAML file somewhere and *not* something like the SQL factory. It should probably have a "driver" field like any data source. Indeed, is it necessary to have the sub-cats in a separate YAML section, or could they be normal entries with driver "yaml_file", "directory", "sql_tables", etc., and the usual additional metadata. That would allow for the inclusion of parameters, so that, for instance, we can only list a set of SQL tables once the user credentials have been rendered from the environment.

PR #66, opened 2018-04-09
New parameter templates for local shell/env

martindurant:
The idea is to have some shell/env values pulled from the client
environment before sending to remote cat server.

In the case that the same cat is used locally, the new templates
will evaluate as before, with the difference that permissions are
not required (since the user already has arbitrary access by
definition).


martindurant:
@seibert , your thoughts on whether this is workable

- [ ] tests
- [ ] docs

intake/catalog/remote.py
```diff
@@ -26,30 +27,38 @@ def describe(self):
...
         if self._real_source is None:
             payload = dict(action='open',
                            name=self._entry_name,
                            parameters=self._user_parameters,
                            available_plugins=list(plugin_registry.keys()))
-            req = requests.post(self._url, data=msgpack.packb(payload, use_bin_type=True))
-            if req.status_code == 200:
+            req = requests.post(self._url, data=msgpack.packb(
+                payload, use_bin_type=True))
+            if req.ok:
```
martindurant: For anyone interested, this changed from code==200, since in theory the server should be able to do redirects, and perhaps at some later date administrative actions too.

seibert:
btw, needs to re-merge master after last PR merge

mmccarty:
LGTM! 👍

PR #67, opened 2018-04-10
Allow catalog to watch missing directories

### LINKED ISSUE #24: Enable catalog to watch a directory that does not exist
There are situations where a catalog will need to watch a directory that does not exist yet (such as the global environment or user directory).  A few changes to the DirectoryState class should make this possible.

jbcrail:
Fix #24

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -51,15 +51,15 @@ class DirectoryState(State):
         catalogs = []
 
         if os.path.isdir(self.observable):
-        self._last_files = []
-        for f in os.listdir(self.observable):
-            if f.endswith('.yml') or f.endswith('.yaml'):
-                path = os.path.join(self.observable, f)
-                try:
-                    catalogs.append(Catalog(path))
-                    self._last_files.append(path)
-                except Exception as e:
-                    logger.warning("%s: %s" % (str(e), f))
+            self._last_files = []
+            for f in os.listdir(self.observable):
+                if f.endswith('.yml') or f.endswith('.yaml'):
+                    path = os.path.join(self.observable, f)
+                    try:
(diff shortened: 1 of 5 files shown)


martindurant:
Good to see this coming. Would this be affected by #61?

martindurant:
OK good. Would appreciate your thoughts on that PR too, as we strengthen the API.

martindurant:
Sorry, needs merge from master.

martindurant:
(I assume this should no longer be WIP?)

jbcrail:
Yes, this is complete pending a review.

jbcrail:
Sounds good. We'll also have to replace `os.listdir`.

jbcrail:
Regarding bandwidth used by timestamp checks, I had thought about using the `E-Tag` or `If-Modified-Since` headers on the remote Intake server to reduce this traffic. We could use the same technique for S3 or similar URLs.

martindurant:
I don't think that would work, because it would be backend-specific, and still require a call to the storage service, or possibly several, each time the check is done. Basically, I imagine the check-interval being a per-catalogue configuration, which also allows not checking at all.

Something like

```
sources:
  ...
catalogs:
  - mycat:
    - destription: stored on s3
    - url: s3://bucket/cat.yaml
    - refresh: 0
```
for sub-cats, or

```
plugins:
  - ...
sources:
  - ...
config:
  - refresh: 120s
```

jbcrail:
Your proposed interval configuration looks good.

martindurant:
Perhaps that can go into a new issue for tracking.

I am starting to feel like there are enough disperate issues/topics in play simultaneously that we may need to formally move to a project board view.

PR #69, opened 2018-04-11
Update catalog docs with new api

ah-:
I'm just reading through the docs and noticed a few occasions where the examples don't work anymore.

The new versions are just my best guess and what seems to work, please update if there's a better way of doing this.

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -136,9 +136,9 @@ Then sources can be listed::
 
     list(cat)
 
-and data sources are loaded with ``get()``::
+and data sources are loaded via their name:
 
-    data = cat['entry_part1'].get(part='1')
+    data = cat.entry_part1(part='1')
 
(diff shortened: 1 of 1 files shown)


docs/source/catalog.rst
```diff
@@ -134,11 +134,11 @@ A Catalog can be loaded from a YAML file on the local filesystem by creating a C
 
 Then sources can be listed::
 
-    cat.list()
+    list(cat)
 
 and data sources are loaded with ``get()``::
 
-    data = cat.get('entry1_part', part='1')
+    data = cat['entry_part1'].get(part='1')
```
martindurant: Does this work? `cat.entry_part1(part='1')` is probably the simplest form.

martindurant:
Yes, unfortunately the API is still somewhat in flux...

docs/source/catalog.rst
ah-: That's even nicer, updated.

martindurant:
please merge from master, it will make the failure go away.

PR #76, opened 2018-04-15
Add storage options explicitly

martindurant:
NB: the general kwargs available before already made this possible, this PR is just making the option explicit, so we can refer to it in documentation.


seibert:
This looks good to me.

PR #82, opened 2018-04-25
Simple auth system

### LINKED ISSUE #80: Authorization plugins for intake server
In order to keep the core of Intake lightweight, while allowing for the customization of authorization policies for a catalog server, we should have a simple way to plugin a custom class to handle these decisions.  Note that the scope here is authorization, not authentication.  Authentication should basically always be handled by reverse proxy (like nginx) sitting in front of Intake which can populate (and strip) verified HTTP headers before passing the request to the Intake server.

Here are some proposed specs for discussion:

An authorization plugin class is a subclass of `intake.auth.AuthorizationPlugin` which requires the following methods to be implemented:

- `allow_connect(http_headers)`: Return `True` if the HTTP connection to the server should be allowed.  The `http_headers` dictionary contains all the headers from the HTTP request.  Reverse proxies can insert any information they like (such as authenticated user names or tokens) into the HTTP headers before passing along to Intake.  This function needs to be fast, as it is called for every request.
- `allow_access(http_headers, catalog_entry)`: Return `True` if the catalog entry (instance of `intake.catalog.entry.CatalogEntry`) should be visible to the client.  This is called for every catalog entry when the user requests a list of sources, and it is called once when the user initially tries to open a source.  The plugin can use any static metadata on the catalog entry to make the authorization decision, as well as the HTTP headers.  This allows use cases, like putting user and group names in the catalog entry metadata to control visibility.

To allow the plugin to be more easily instantiated with custom options, it will be activated via the planned catalog metadata section (see #75).  For example:
``` yaml
meta:
  version: 1
  authorization:
    plugin_class: intake_ldap_auth.AuthPlugin
    args:
      domain_controller: ldap.example.com
      dn: dc=exampledomain,dc=myorg
```
Only one authorization plugin can be active for a given server (if none is selected, the current behavior of "allow everything" continues).  If more than one is found, an error is raised and the server aborts.  In most cases, the auth plugin will be in a catalog file that is empty other than the `meta` section.

[edit: put the arguments to the class constructor into an args dict]

martindurant:
- The default auth allows everything
- The secret auth allows everything if the key is in the header
- This is meant mainly as a template for other, more complex auth systems

Fixes #80

### DIFF
--- a/intake/auth/base.py
+++ b/intake/auth/base.py
@@ -1,6 +1,6 @@
 
 
-class BasicAuth(object):
+class BasicAuthPlugin(object):
     """Base class for authorization
 
     This basic class allows all access.
@@ -10,9 +10,23 @@ class BasicAuth(object):
         self.args = args
 
     def allow_connect(self, header):
-        """Is the requests header given allowed to talk to the server"""
+        """Is the requests header given allowed to talk to the server
+
+        Parameters
+        ----------
+        header: dict
+            The HTTP header from the incoming request
+        """
         return True
 
     def allow_access(self, header, source):
-        """Is the given HTTP header allowed to access given data source"""
+        """Is the given HTTP header allowed to access given data source
+
+        Parameters
+        ----------
+        header: dict
+            The HTTP header from the incoming request
+        source: CatalogEntry
+            The data source the user wants to access.
+        """
         return True
--- a/intake/auth/base.py
+++ b/intake/auth/base.py
@@ -1,6 +1,6 @@
 
 
-class BasicAuthPlugin(object):
+class BaseAuth(object):
     """Base class for authorization
 
     This basic class allows all access.
--- a/intake/auth/base.py
+++ b/intake/auth/base.py
@@ -1,0 +1,18 @@
+
+
+class BasicAuth(object):
+    """Base class for authorization
+
+    This basic class allows all access.
+    """
+
+    def __init__(self, *args):
+        self.args = args
+
+    def allow_connect(self, header):
+        """Is the requests header given allowed to talk to the server"""
+        return True
+
+    def allow_access(self, header, source):
+        """Is the given HTTP header allowed to access given data source"""
+        return True
--- a/intake/auth/secret.py
+++ b/intake/auth/secret.py
@@ -1,11 +1,11 @@
 import logging
-from .base import BasicAuthPlugin
+from .base import BaseAuth
 import uuid
 
 logger = logging.getLogger('intake')
 
 
-class SecretAuth(BasicAuthPlugin):
+class SecretAuth(BaseAuth):
     """A very simple auth mechanism using a shared secret
 
     Parameters
--- a/intake/auth/secret.py
+++ b/intake/auth/secret.py
@@ -1,16 +1,18 @@
-
-from .base import BasicAuth
+import logging
+from .base import BasicAuthPlugin
 import uuid
 
+logger = logging.getLogger('intake')
 
-class SecretAuth(BasicAuth):
+
+class SecretAuth(BasicAuthPlugin):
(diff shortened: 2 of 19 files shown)


intake/auth/base.py
```diff
@@ -0,0 +1,18 @@
...
+    """Base class for authorization
+
+    This basic class allows all access.
+    """
+
+    def __init__(self, *args):
+        self.args = args
+
+    def allow_connect(self, header):
+        """Is the requests header given allowed to talk to the server"""
```
seibert: Can you indicate that header is a dictionary of HTTP headers in the docstring?

intake/auth/base.py
```diff
@@ -0,0 +1,18 @@
...
+        return True
+
+    def allow_access(self, header, source):
+        """Is the given HTTP header allowed to access given data source"""
```
seibert: Same here: header is dict of HTTP headers and source is an instance of CatalogEntry

intake/auth/secret.py
```diff
@@ -0,0 +1,34 @@
...
+    secret: str
+        The string that must be matched in the requests. If None, a random UUID
+        is generated.
+    key: str
+        Header entry in which to seek the secret
+    """
+
+    def __init__(self, secret=None, key='intake-secret'):
+        if secret is None:
+            secret = uuid.uuid1().hex
```
seibert: Does it make sense to autogenerate a secret key?  There's no way to even know what it is and communicate it to the client.  I'd rather raise an exception for not setting the `secret` argument.

intake/auth/base.py
```diff
@@ -0,0 +1,18 @@
+
+
+class BasicAuth(object):
```
seibert: I'd rename this to `BaseAuth` or `BaseAuthPlugin`.  `BasicAuth` is likely to be confused with HTTP Basic Authentication.

intake/auth/secret.py
martindurant: I am thinking about it like the jupyter token, something that could be printed to the console or stored at runtime. I would suggest actually that there is a log statement here to record the new random value.

intake/cli/server/server.py
```diff
@@ -65,15 +70,23 @@ def _make_cache_callback(self, callback, idle_time):
...
+            sources = []
+            for _, name, source in self.catalog.walk():
+                if self.auth.allow_access(head, source):
+                    info = source.describe()
+                    info['name'] = name
+                    sources.append(info)
+
+            server_info = dict(version='0.0.1', sources=sources)
+        else:
+            server_info = dict(version='0.0.1', sources=[])
```
seibert: If `allow_connect` fails, we should return 403 (same as failure of `allow_access` below), otherwise the client won't be able to tell the difference between overall authorization failure and an empty remote catalog.

seibert:
This looks good!  Aside from individual comments above, I think this just needs a test to verify that configuration loading works.  (Sets port, selected auth plugin is actually enabled, etc.)

intake/auth/secret.py
seibert: Ah, I see.  That makes more sense, although I don't think the usage model is that similar to Jupyter.  The intake server is likely to be started by a different person than the client, but I'm fine with either failing or picking a secret and logging it to the console.

seibert:
One super minor nit about "Base" vs. "Basic", and I also realized we need a new documentation page on activating and creating auth plugins.

martindurant:
I think we need to have a documentation sprint in general.
Also, we now have configuration; what about cat metadata?

seibert:
Yeah, now is a good time to add the minimal catalog metadata section I think.

I'm taking a pass through the existing docs and will be opening a PR for those updates.

PR #87, opened 2018-05-04
Allow remote catalogs

martindurant:
- the URL for the intake server is now intake://

- general URLs like s3:// are now accepted

- Catalog has gained storage_options, which either are requests options
  for the server, or the usual options for the backend. This is, in
  practice, how you can provide auth headers for the server

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -18,7 +18,7 @@ Intake catalogs are described with YAML files.  Here is an example:
 
     sources:
       example:
-        description: test  
+        description: test
         driver: random
         args: {}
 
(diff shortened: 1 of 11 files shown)


martindurant:
Try `cat = intake.Catalog('s3://MDtemp/us_crime.yaml')` !

Note that this breaks one test, because I skipped checking timestamps on files. We could get a dasky unique version token using `f.fs.ukey(f.path)` where `f` is an `OpenFile`.

martindurant:
Once again, if the idea is approved, I'll make tests and update docs.

seibert:
My main concern is that HTTP vs HTTPS communication with the Intake server can't be indicated if`intake://` is the scheme.  Maybe `intake+https://example.com:5000` for remote catalog servers vs `https://example.com/catalog.yaml` for a remote catalog file?

martindurant:
Yes, that is something to sort out. It could just be another parameter ('scheme=').

martindurant:
Note that I found there is no up-to-date check available on HTTP, which would mean infinite-reloads, so instead we have no reloads ever.

rabernat:
What is the difference between an intake server and a remote catalog file?

seibert:
Functionally, nothing if you don't need the Intake server to:

* Enforce any access control to specific catalog entries (see auth plugins)
* Stream data to clients because they lack the required data plugins, or because they actually can't reach the data source directly.
* Obtain credentials or other things by running scripts locally on the server and substituting them into the catalog

For completely open shared catalog use cases, posting a catalog file on a file server (HTTP/S3/etc) is the most straightforward option once we merge this PR.

martindurant:
We have not yet done this, but it should be possible to open netCDF data on a server and stream them to clients without those clients having direct access to the file-system hosting the files. There are indeed dedicated netCDF servers that do only this job, but we would do it rather simply, and I expect at a performance cost relative to dedicated services.

PR #88, opened 2018-05-05
Configuration docs

martindurant:
This section could become its own page eventually, as we add things we want to configure.


seibert:
Merging this so I can refer to these docs in the auth plugin docs.

PR #90, opened 2018-05-07
Client auth

seibert:
Most server-side authorization schemes (implemented by the auth plugins) will require matching client side code that can populate the HTTP headers of requests to the Intake server with some information.  This PR implements a minimal plugin API that queries a client auth plugin for extra HTTP headers prior to each call.

One subtle issue is that to enable serialization of data sources to remote workers, we have to freeze these headers when the RemoteDataSource is created, as we can't rely on the client auth plugin to work properly on the workers.  (There are likely use cases where this will be insufficient, but to resolve it we first need to decide on a reliable way to run Dask tasks back on the client, even when the client is not running a worker process in the Dask cluster.)

To do:
* [x] Integration test demonstrating that secret auth / client auth plugins cooperate properly
* ~~Client plugin discovery and configuration~~ (Not needed in this first pass)
* [x] Documentation

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -190,3 +190,8 @@ Whether a particular catalog entry supports direct or proxied access is determin
 - ``force``: Force all clients to access the data directly.  If they do not have the required plugin, an exception will be raised.
 
 Note that when the client is loading a data source via direct access, the catalog server will need to send the plugin arguments to the client.  Do not include sensitive credentials in a data source that allows direct access.
+
+Client Authorization Plugins
+''''''''''''''''''''''''''''
+
+Intake servers can check if clients are authorized to access the catalog as a whole, or individual catalog entries.  Typically a matched pair of server-side plugin (called an "auth plugin") and a client-side plugin (called a "client auth plugin) need to be enabled for authorization checks to work.  This feature is still in early development, so please `open a Github issue <https://github.com/ContinuumIO/intake/issues/new>`_ to discuss your use case before creating a plugin.
--- a/intake/catalog/tests/test_auth_integration.py
+++ b/intake/catalog/tests/test_auth_integration.py
@@ -1,0 +1,73 @@
+import os
+import os.path
+import shutil
+import tempfile
+import time
+
+import pytest
+
+from .util import assert_items_equal
+from intake.catalog import Catalog
+
+from intake.auth.secret import SecretClientAuth
+
+TMP_DIR = tempfile.mkdtemp()
+CONF_DIR = os.path.join(TMP_DIR, 'conf')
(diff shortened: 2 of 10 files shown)


martindurant:
Is the intent to specify an auth class in configuration (like you can do for the server), or always to pass a configured *instance* to the call to Catalog? Can we do both?

seibert:
Yes, the idea is to allow both setting the `auth` kwarg on `Catalog` and also having a default auth set in a configuration file.

seibert:
Although I'm not sure if having a global config file is the right usage model, so maybe just explicit `auth` for now...

seibert:
OK, I've put in some placeholder documentation.  I think I want to add more documentation as part of a larger docs overhaul once we finish the other things in flight.

intake/catalog/tests/test_auth_integration.py
```diff
@@ -0,0 +1,73 @@
...
+import pytest
+
+from .util import assert_items_equal
+from intake.catalog import Catalog
+
+from intake.auth.secret import SecretClientAuth
+
+TMP_DIR = tempfile.mkdtemp()
+CONF_DIR = os.path.join(TMP_DIR, 'conf')
+os.mkdir(CONF_DIR)
```
martindurant: Should be context or fixture, I think

intake/catalog/tests/test_auth_integration.py
```diff
@@ -0,0 +1,73 @@
...
+def teardown_module(module):
+    shutil.rmtree(TMP_DIR)
+
+
+@pytest.fixture
+def intake_server_with_auth(intake_server):
+    fullname = os.path.join(TMP_DIR, YAML_FILENAME)
+
+    with open(fullname, 'w') as f:
+        f.write('''
```
martindurant: Maybe we should include MemoryFS to avoid having to write all these temporary files... or be able to directly pass a string to Catalog(), without assuming it's a path to open.

docs/source/catalog.rst
```diff
@@ -190,3 +190,8 @@ Whether a particular catalog entry supports direct or proxied access is determin
 - ``force``: Force all clients to access the data directly.  If they do not have the required plugin, an exception will be raised.
 
 Note that when the client is loading a data source via direct access, the catalog server will need to send the plugin arguments to the client.  Do not include sensitive credentials in a data source that allows direct access.
+
+Client Authorization Plugins
+''''''''''''''''''''''''''''
+
```
martindurant: Should include the path to the simple auths we have and/or a very brief description of what they are expected to provide (a headers method).

seibert:
OK, going to merge this PR and start another one with some follow up fixes

PR #93, opened 2018-05-14
Some fixes from profiling in py2

martindurant:
This is enough to fix tests on py2 - I had thought there was more.


martindurant:
@seibert , I think this is now worth a quick once-over

seibert:
Looks good!

PR #94, opened 2018-05-15
small misc changes

martindurant:
- fast conda find
- time module imports
- munge non-standard plugin names for open_*

### DIFF
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -1,3 +1,4 @@
+import re
 import warnings
 
 from . import source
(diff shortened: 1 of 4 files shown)


intake/__init__.py
```diff
@@ -24,6 +26,9 @@ def make_open_functions():
     """From the current state of ``registry``, create open_* functions"""
     for plugin_name, plugin in registry.items():
         func_name = 'open_' + plugin_name
+        if not isidentifier(func_name):
+            # primitive name normalization
+            func_name = re.sub('[-=~^&|@+]', '_', func_name)
```
seibert: Can we also emit a warning that we've done this renaming?  Plugins should generally use identifier-friendly names and not rely on a specific name normalization scheme.

seibert: OK, I'm happy to do the check in the base class.

PR #96, opened 2018-05-16
Auth plugin improvements

seibert:
* Pass catalog object to allow_access so that global attributes can be pull from the metadata section
* Document how to write and use auth plugins
* Add some cross reference links in the docs


martindurant:
Yes, I'm OK with all of this.
We will probably want to clarify later, based on user feedback, but merging now.

PR #104, opened 2018-05-28
Remove the data source browser page

seibert:
The HTTP-based data source browser had minimal functionality, and can't easily be made to work with auth plugins anyway.  Data source browsing should be done as a JupyterLab extension instead.


martindurant:
Agree totally, plus the jupyter-lab browser should also be able to work for local-only and built-in cats.

intake/catalog/tests/conftest.py
```diff
@@ -23,7 +23,7 @@ def ping_server(url, swallow_exception):
         else:
             raise e
 
-    return r.status_code == 200
+    return r.status_code in (200, 403)  # allow forbidden as well
```
martindurant: I thought it would still be possible to "see" the server even when we have no permissions to access any of the data - it looks like an empty cat. Is this wrong?

seibert: There are two checks done by the auth plugins: one to determine if the connection is allowed, and one to determine if the data source is available to the client.  If the first check is denied, automatic 403 on every request.  If the second one fails, then the catalog listing will be empty, and attempting to open the source will return the 403

PR #105, opened 2018-05-29
Descent into source parameters for expansion

### LINKED ISSUE #103: nested defaults
If a source has parameters which include nested structures, e.g., `storage_options`, which is always a dictionary, there is no way to make a part of that structure a user parameter.

Could be fixed by recursively descending into standard argument containers and checking for jinja-like strings at all levels.

martindurant:
Now can apply user parameters to deeper structures to be passed
as arguments

Fixes #103


seibert:
Looks good.

Issue #107, opened 2018-06-10
problem under pip


stonebig:
I don't catch why but this instruction doesn't work, yet `ruamel.yaml` is installed (Python-3.7, same error on Python-3.6))
````
from ruamel_yaml.constructor import DuplicateKeyError
````
I get a
````
C:\WinP\bd37\bu\winp64-3.7.x.1\python-3.7.0b5.amd64\lib\site-packages\intake\catalog\local.py in <module>()
      7 import yaml
      8 
----> 9 from ruamel_yaml.constructor import DuplicateKeyError
     10 
     11 from jinja2 import Template

ModuleNotFoundError: No module named 'ruamel_yaml'
````
any clue ?

stonebig:
!pip list 
````
Package                  Version      
------------------------ -------------
adodbapi                 2.6.0.7      
alabaster                0.7.10       
algopy                   0.5.7        
altair                   2.1.0        
altair-widgets           0.1.2        
appdirs                  1.4.3        
asciitree                0.3.3        
asteval                  0.9.12       
astroid                  1.6.5        
astroML                  0.3          
atomicwrites             1.1.5        
attrs                    18.1.0       
Babel                    2.6.0        
backcall                 0.1.0        
backports-abc            0.5          
baresql                  0.7.4        
bcolz                    1.2.1        
beautifulsoup4           4.6.0        
blaze                    0.10.1       
bleach                   2.1.3        
blosc                    1.5.1        
bloscpack                0.11.0       
bokeh                    0.12.16      
Bottleneck               1.2.1        
bqplot                   0.10.5       
brewer2mpl               1.4.1        
Brotli                   1.0.4        
certifi                  2018.4.16    
cffi                     1.11.5       
cftime                   1.0.0        
chardet                  3.0.4        
click                    6.7          
cloudpickle              0.5.3        
clrmagic                 0.0.1a2      
colorama                 0.3.9        
colorcet                 0.9.1        
comtypes                 1.1.4        
cvxopt                   1.2.0        
cvxpy                    1.0.5        
cx-Freeze                5.1.1        
cycler                   0.10.0       
cyordereddict            1.0.0        
Cython                   0.28.3       
cytoolz                  0.9.0.1      
dask                     0.17.5       
dask-searchcv            0.2.0        
datashader               0.6.6        
datashape                0.5.2        
db.py                    0.5.3        
decorator                4.3.0        
dill                     0.2.7.1      
distributed              1.21.8       
[...]

PR #108, opened 2018-06-10
make it work for normal pypi / Python-3 users (using cgohlke wheel)

stonebig:
would solve https://github.com/ContinuumIO/intake/issues/107,  I suppose

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -6,7 +6,10 @@ import os.path
 import runpy
 import yaml
 
-from ruamel.yaml.constructor import DuplicateKeyError
+try:
+    from ruamel_yaml.constructor import DuplicateKeyError
+except:
+    from ruamel.yaml.constructor import DuplicateKeyError
 
 from jinja2 import Template
 import six
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -6,7 +6,7 @@ import os.path
 import runpy
 import yaml
 
-from ruamel_yaml.constructor import DuplicateKeyError
+from ruamel.yaml.constructor import DuplicateKeyError
 
 from jinja2 import Template
 import six


martindurant:
Thank you, @stonebig 
So I understand, this is to get around a difference in naming convention between conda and pypi's version of ruamel? Does this completely fix #107 , or was some of the issue still outstanding?

stonebig:
it resolves the package naming convention, but:
-  then I fell on the conda prefix https://github.com/ContinuumIO/intake/issues/107#issuecomment-396059722 , workarounded per another try/expect.
- then I failed to find any uc-crime (can you beleive that ?) 
 https://github.com/ContinuumIO/intake/issues/107#issuecomment-396060088 

At that point, I gave up. I suppose Intake is not for pip.

seibert:
This is something we can fix.  Thanks for raising the issue and we can build upon this PR.

philippjfr:
>then I failed to find any us-crime (can you believe that ?)

😄It should be available from the intake channel.

martindurant:
> It should be available from the intake channel.

we do not have a distribution for this in a pure pypi package. I'm afraid that we have been perhaps too opinionated in preferring conda as the distribution engine.

seibert:
Well, pip and PyPI are for Python packages, not for arbitrary package-able artifacts.  There isn't really a natural way to map "data packages" to pip any more than C++ libraries, so I think our opinion in this case is somewhat justified.  :)

However, for ease of running examples, we can rely more on the ability to load catalog and data files over HTTP to remove the need for conda installing things.

chobeat:
any news on this issue?

chobeat:
Is this version already on Pypi?

martindurant:
> Is this version already on PyPI?

It is, but out of date https://pypi.org/project/intake
I'll push a new release. @seibert , can you give me permissions?

martindurant:
-EDIT-
updated on PyPI. Note that that I don't think any of the plugins have been released, and some of them have rather complicated dependencies, so please request releases on the plugins that matter to you.

chobeat:
So far I'm just working with CSV. I will let you know when I will start working with other formats. Thank you

martindurant:
@chobeat , ok cool. Please let me know if install and functionality is OK for you now.
(can close this and #107 , if so)

seibert:
the easier and most useful plugins to do would be:
- intake-parquet
- intake-sql
- intake-xarray

chobeat:
Ok, it worked and now the documentation finally matches the behavior I see on my machine. I had to update a call to `open_csv` but the import works flawlessly

martindurant:
Excellent, thank.

PR #120, opened 2018-07-17
Ripout state

martindurant:


martindurant:
Replaces #61

martindurant:
@seibert , you may find this interesting. Single file and multi-file cats now actual data-source plugins with type "catalog".

martindurant:
So, states are gone, plugins are gone and containers are gone. This would require bumping versions on all existing plugins.

intake/catalog/local.py
```diff
@@ -120,21 +118,19 @@ def __init__(self, name, description, driver, direct_access, args,
...
-    def find_plugin(self, registry):
-        if self._driver in registry:
-            self._plugin = registry[self._driver]
-        else:
-            self._plugin = global_registry[self._driver]
-
     def describe(self):
+        if isinstance(self._plugin.container, (list, tuple)):
+            import pdb
+            pdb.set_trace()
```
mmccarty: Don't forget to remove

intake/catalog/local.py
```diff
@@ -120,21 +118,19 @@ def __init__(self, name, description, driver, direct_access, args,
...
+            print('####', self._plugin)
```
mmccarty: Perhaps we should have an issue to come back and replace `print` with a logger?

intake/catalog/local.py
```diff
@@ -326,18 +261,25 @@ def _parse_plugins(self, data):
                 self.error("keys 'module' and 'dir' both exist (select only "
                            "one)", plugin_source)
             elif 'module' in plugin_source:
-                obj = self._parse_plugin(plugin_source, 'module')
-                if obj:
-                    sources.append(obj)
+                for k, v in load_plugins_from_module(
```
mmccarty: For this block and the one below, it feels like there's enough here to warrant a separate function (`register_module` and `register_dir`).

intake/container/base.py
```diff
@@ -1,21 +1,81 @@
...
-    @staticmethod
-    def decode(bytestr):
-        raise NotImplementedError
+    def _get_source_id(self):
+        if self._source_id is None:
+            payload = dict(action='open', name=self.name,
+                           parameters=self.parameters)
+            req = requests.post(self.url, data=msgpack.packb(
+                payload, use_bin_type=True), **self.headers)
+            if req.status_code == 200:
```
mmccarty: Do we need to raise an exception here?  We do that above.

intake/container/base.py
```diff
@@ -1,21 +1,81 @@
...
+        self.npartitions = response['npartitions']
+        self.metadata = response['metadata']
+        self._schema = Schema(datashape=None, dtype=self.dtype,
+                              shape=self.shape,
+                              npartitions=self.npartitions,
+                              extra_metadata=self.metadata)
+        self._source_id = response['source_id']
+
+    def _get_partition(self, i):
+        pass
```
mmccarty: todo?

intake/source/base.py
```diff
@@ -139,14 +99,7 @@ def discover(self):
...
         """Load entire dataset into a container and return it"""
-        self._load_metadata()
-
-        parts = [self._get_partition(i) for i in range(self.npartitions)]
-
-        return self._merge(parts)
-
-    def _merge(self, parts):
-        return get_container_klass(self.container).merge(parts)
+        pass
```
mmccarty: todo?

intake/__init__.py
```diff
@@ -60,6 +69,44 @@ def output_notebook(inline=True, logo=False):
...
+                else:
+                    # single filename in a list
+                    driver = 'yaml_file_cat'
+            elif isinstance(uri, str):
+                # single URL
+                if uri.startswith('intake:'):
+                    # server
+                    driver = 'intake_remote'
+                else:
+                    if uri.endswith('.yml') or uri.endswith('.yaml'):
```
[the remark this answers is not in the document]
martindurant: Always forget about that

intake/catalog/local.py
martindurant: Doesn't actually get called. Was caused by one of the datasources having a rogue comma, making a string into a tuple.

intake/container/base.py
martindurant: I prefer no-op over virtual methods or NotImplemented, just for simplicity.

intake/catalog/local.py
martindurant: done

intake/container/base.py
martindurant: done

intake/catalog/local.py
martindurant: done

intake/catalog/local.py
martindurant: done

intake/__init__.py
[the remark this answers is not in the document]
martindurant: done

intake/container/base.py
mmccarty: Perhaps a comment then to explain why this is not implemented.

martindurant: Do you personally prefer NotImplemented? That's what's done in the DataSource base class.

mmccarty: It depends.  I think the noop is fine if a noop is the default.  If you are depending on the method to actually do something, then I would raise NotImplemented.

martindurant: done

intake/source/base.py
martindurant: done

martindurant:
Relented... those methods should never be directly called, all containers should override them.

Also added a nested test :)

martindurant:
@seibert , remind me: if I merge this, it will build a <dev> package, and only tagging will result in a <main> package, correct?

martindurant:
(the documentation will need updating before release, especially the "how to make a plugin", because we no longer need `Plugin`; also all existing plugins would need to be changed and released very shortly after intake, and depending on the new version)

seibert:
This repo uploads every merge to master with a build number that increments for each commit since the most recent tag (basically `git describe`).  That is useful for testing, but we should be uploading them with a `dev` label.

martindurant:
What would it take to get the dev/main label behaviour? I am *not* ready to start a release process!

seibert:
I think it would require changing the anaconda upload line to set the label (and probably copying the bash if statement from one of the plugins to branch for tagged and untagged commits).

PR #125, opened 2018-07-31
Add npy source and remote array

### LINKED ISSUE #112: implement arrays from server
intake-xarray already implements a Source which has container type “xarray", which works totally fine in the local case, because it overrides all of the base methods (not just the internal ones). It has peculiarities, such as the schema.dtype being the xarray object itself, to_dask() and read_iter() returning exactly the same array again, and read() also returning the same thing, but with the data members materialised in memory. That’s all fine, if extremely ad-hoc.

The trouble comes with the remote source (or proxy), where the container name is used to determine the behaviour in packing up the source definition, both implicit in Source.base, and explicit in the container classes (xarray doesn’t have one). I want the container types to be extensible too, so that when a source has data not fitting the existing types (keras model??), the code to make it work should be in the plugin package, not intake main. To achieve this, the source would need to be able to serialise all information about itself in it’s schema, which also specifies the set of chunks than can be loaded, and specify how to assemble the calls into a dask and in-memory version. So that’s at least two helper functions per container, and a registry in Intake. Calls to the get-chunk would pass through a full spec to get a given piece, which for arrays would be a tuple of ints, for xarray datasets a tuple of (name, int, int…).

On the final point, I started off already in the mindset that dask should be the de-facto target and have become only more firm in this view. In this model, get_partition would be a call to bag.to_delayed()[i].compute(), dataframe.get_partition(i).compute(), Delayed((arr.name, int, int), arr.dask) for bag, dataframe and array, resp. I would note that dask already has xarray as a first-class citizen, and even has a serialiaser for keras built in. If we can provide a natural home for dask read_* functions, not within the dask codebase, that would make everyone happy.

PS: dask arrays now have the blocks interface to simplify grabbing chunks for streaming.

martindurant:
Remote streaming of array data

Fixes #112 

cc @mmccarty


intake/source/npy.py
```diff
@@ -0,0 +1,108 @@
...
+            if self.shape is None:
+                arr = NumpyAccess(files[0])
+                self.shape = arr.shape
+                self.dtype = arr.dtype
+                arrs = [arr] + [NumpyAccess(f, self.shape, self.dtype)
+                                for f in files[1:]]
+            else:
+                arrs = [NumpyAccess(f, self.shape, self.dtype)
+                        for f in files]
+            self.chunks = (self._chunks, ) + (-1, ) * (len(self.shape) - 1)
```
mmccarty: Curious, why not set this in the `__init__` method and get rid of `self._chunks`?  Do we need the distinction between `self.chunks` and `self._chunks`?  Perhaps this is a convention I'm unaware of.

intake/container/ndarray.py
```diff
@@ -1,7 +1,58 @@
-from .base import RemoteSource
+import itertools
+from .base import RemoteSource, Schema, get_partition
 
 
 class RemoteArray(RemoteSource):
-    # TODO
-    def __init__(self, *args, **kwargs):
-        raise NotImplementedError
+    """nd-array on an Intake server"""
```
mmccarty: Where/how is this tested?

mmccarty: I see it now, after looking at the individual commits.

intake/source/npy.py
martindurant: I wanted to store the original input explicitly - this seems like a good idea for hygiene and serialization in general.
`self.chunks` is a computed value, and depends on the array shape, which may have been provided, but also may have come from the first of the files. Intake should perform as little work on DataSource instantiation as possible.

Issue #95, opened 2018-05-16
Develop a jupyter-lab simple data browser pane


martindurant:
Idea from https://github.com/pangeo-data/pangeo/issues/39#issuecomment-389519782

seibert:
This looks like a great example to follow: https://github.com/eWaterCycle/jupyterlab_thredds

johngroves:
I have an initial JupyterLab extension working with a remote catalog, I can open a PR if it is helpful. The extension sends a request to the /v1/info endpoint, presents a list of sources, and provides code for loading the data. 

To do this, I've had to Dockerize the intake server and implement an nginx proxy to avoid CORS issues. Is there another endpoint for catalog discovery?


![intake_data](https://user-images.githubusercontent.com/10951009/42913880-31ba7744-8aac-11e8-9c83-f19d65c9fed7.gif)

martindurant:
This is a great start, thank you!
Yes, please open a PR, so we can discuss.
I assume that in your panel, the URI to poke for datasets is pre-determined? By working with the server HTTP endpoint, there is no python logic at all, correct?

seibert:
This is very exciting!  We had imagined this having a Python kernel behind it (which is what the threadds extension does), so that you would not have to deal with CORS issues, and it would also allow the browser to see local catalogs on the filesystem (including the default catalog in `intake.cat`).

To echo Martin's suggestion: Feel free to open an WIP PR and we can discuss further.

martindurant:
Please don't hesitate to get in direct touch with us if you need some help here.

johngroves:
@martindurant Apologies for dropping off here. Is anyone working on a Jupyterlab extension, or is the new GUI providing the desired functionality?

martindurant:
The new GUI can be used in a notebook or standalone, but no one is working specifically on a j-lab pane. Having such a pane, perhaps a separate process or perhaps sharing state with a kernel running one of the notebooks, could be useful.

PR #130, opened 2018-08-10
Add data browser

martindurant:
One implementation for #95 

Styling, and how to call this at the top level TBD.


martindurant:
screen shot 2018-08-10 at 10 49 46

ericdill:
Ah this is pretty cool! What are your thoughts on having `gui.item.discover()` get called on selection and show in a new textarea?

martindurant:
cc @ericdill @seibert @pzwang

martindurant:
@ericdill , yes the thought had occurred; probably `describe_open()`, instead, since `discover()` can require resources. 
I didn't show the file selector that opens if you click the folder icon, so it would need a little planning to have it not too busy looking.

martindurant:
What I really like here, is the small number of lines it took to make this happen compared to the threadds widget, and that this needs no compilation or server extension installation. Obviously it doesn't preclude a more sophisticated jupyterlab widget.

martindurant:
Now includes info box for currently-selected item, and the behaviour of the file selector is nicer (has a close button which cancels the operation).

rabernat:
Is there a way to put this into a jupyterlab right panel (like https://github.com/eWaterCycle/jupyterlab_thredds) rather than a notebook widget?

Or would that be a separate project?

I started playing around with a jupyter lab extension to browse an intake catalog.

rabernat:
Ok thanks.

I think the extension would have a lot of value for the pangeo community. I'm going to try to work on it some more. It might not be that hard.

martindurant:
I would be **all behind** having the option of either :)  In fact, it'd a nice example of community-wide involvement. If you need anything from me to make things easier, just ask.

rabernat:
The option we are pursing with @ian-r-rose is to do a notebook extension only (pure javascript) which loads and parses a catalog file directly.

What would be most useful is to have a firm spec for the catalog.yaml file. Currently we have an example, but that's not quite the same as a spec...

martindurant:
Since ipywidgets is in the pangeo image already, can try the GUI I made with
```
conda install -c intake intake-xarray=0.1.1  -y # update from what is in the image now
conda install -c intake data-us-states  -y # so there is an example "builtin" dataset
pip install -U --no-deps git+https://github.com/ContinuumIO/intake  # master version; think better with nodeps
```
and in a notebook
```
import intake
intake.gui
```
You could enter "http://raw.githubusercontent.com/pangeo-data/pangeo/master/gce/catalog.yaml" in the URL to show the standard datasets, or we could pre-populate.

martindurant:
screen shot 2018-08-15 at 16 43 30

(I prefer dark :) )

martindurant:
^ this is for pangeo eyes

seibert:
I worry that a pure Javascript extension for reading catalog files will be limited to only catalog YAML files served from a webserver (which is likely your use case) but be unable to deal with catalog files on local/network filesystems or connecting to an Intake server.

A YAML schema is something that could be done.  In the early days of Intake, we tried using a schema to validate catalog files, but had a hard time producing good error messages, so it was abandoned for a hand-written validator.  The schema could be brought back, but would have to be maintained in parallel.

In #95, @johngroves demoed a Javascript-only JupyterLab extension that connected to an Intake server, and mentioned he was interested in adding a Python kernel that would handle everything directly.

martindurant:
If all you wanted to do was get a list of data sources and container/description for each, then JS-only could work (and should be able to read from HTTP or local file, right?), but intake does allow for nested catalogs and catalogs that only evaluate on demand, e.g., tables in a database source. Still, the simple version of the sources pane might still be useful.
For this simplified case, the spec is simply
```
...
sources:
    key:
        description: text
        driver: text
```
Where `key` is the name of the dataset, `description` is free text and `driver` is only of the plugin names.

seibert:
Ah right, I forgot that the JupyterLab server can act as a proxy to send files from the local filesystem to the browser (only within the starting directory, like Jupyter Notebook?).

martindurant:
(or, I should say, the pane can choose to show as little of the information in the YAML file as is convenient - it could simply show a structured view of the file, but probably the two fields I show are the most important, and other ones are not necessarily guaranteed)

PR #135, opened 2018-08-17
improvements to path munging for cache_paths.

mmccarty:
I noticed a few rough edges when using http remote files.  This PR addresses them.

### DIFF
--- a/intake/source/cache.py
+++ b/intake/source/cache.py
@@ -8,9 +8,19 @@ import logging
 import os
 import shutil
 
+from dask.bytes.utils import infer_storage_options
 from intake.config import conf
 
 logger = logging.getLogger('intake')
+
+
+def strip_http(path):
+    storage_option = infer_storage_options(path)
+
+    protocol = storage_option['protocol']
+    if protocol in ('http', 'https'):
+        return path.replace("{}://".format(protocol), '')
+    return path
 
 
 class FileCache(object):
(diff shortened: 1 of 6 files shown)


intake/source/cache.py
```diff
@@ -8,11 +8,21 @@
...
 
 logger = logging.getLogger('intake')
 
 
+def strip_http(path):
+    storage_option = infer_storage_options(path)
+
+    protocol = storage_option['protocol']
+    if protocol in ('http', 'https'):
+        return path.replace("{}://".format(protocol), '')
```
martindurant: Hm, right... so most FSs remove the protocol, but http doesn't?
This needs a comment in the code at least.

intake/source/cache.py
```diff
@@ -44,12 +54,17 @@ def _ensure_cache_dir(self):
 
     def _munge_path(self, cache_subdir, urlpath):
         import re
+
+        regex = os.path.normpath(strip_http(self._spec['regex']))
+        path = os.path.normpath(strip_http(urlpath))
```
martindurant: Is `normpath` necessary? What failed without it? Worry that only local FS it makes no difference, but remote FS might require original key without munge (or is this only for local files in the cache?)

intake/source/cache.py
```diff
@@ -106,8 +121,14 @@ def load(self, urlpath):
             cache_path = file_out.path
             cache_paths.append(cache_path)
 
+            # If `_munge_path` did not find a match we want to avoid writing to the urlpath.
+            if cache_path == urlpath:
+                continue
+
             if not os.path.isfile(cache_path):
-                logger.info("Caching file {} from urlpath {}".format(file_in.path, urlpath))
+                logger.info("Caching file: {}".format(file_in.path))
```
martindurant: Change these to debug and add docs for how to set logging level.

intake/source/cache.py
mmccarty: That's right.  I'll add a comment.

intake/source/cache.py
mmccarty: I'm using it here to remove trailing slashes. :-)

intake/source/cache.py
mmccarty: 👍

intake/source/cache.py
mmccarty: Also, notice I'm calling `normpath` on the striped `regex` which http is the only case that will require the protocol in the regex to match properly.  However, you're right about the `urlpath`.  I'll fix that.

mmccarty:
Any further comments? Or can we merge this one?

PR #138, opened 2018-08-21
Added progress bar for cache downloads.

mmccarty:
progress_bar

Since making this gif, I've added clearing the output after each file has downloaded.

### DIFF
--- a/intake/source/cache.py
+++ b/intake/source/cache.py
@@ -120,7 +120,6 @@ class FileCache(object):
 
         from dask.bytes import open_files
         from tqdm import tqdm_notebook
-        from IPython.display import clear_output
 
         self._ensure_cache_dir()
         subdir = self._hash(urlpath)
@@ -144,7 +143,7 @@ class FileCache(object):
                 file_size = file_in.fs.size(file_in.path)
                 percent_per_block = 100 * self.blocksize / file_size
                 print("Caching {}".format(file_in.path))
-                with tqdm_notebook(total=100) as pbar:
+                with tqdm_notebook(total=100, leave=False) as pbar:
                     with file_in as f1:
                         with file_out as f2:
                             data = True
@@ -152,7 +151,6 @@ class FileCache(object):
                                 data = f1.read(self.blocksize)
                                 f2.write(data)
                                 pbar.update(int(percent_per_block))
-                clear_output()
 
 
         return cache_paths
--- a/intake/source/cache.py
+++ b/intake/source/cache.py
@@ -119,7 +119,7 @@ class FileCache(object):
             return [urlpath]
 
         from dask.bytes import open_files
-        from tqdm import tqdm_notebook
+        from tqdm.autonotebook import tqdm
 
         self._ensure_cache_dir()
         subdir = self._hash(urlpath)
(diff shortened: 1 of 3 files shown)


intake/source/cache.py
```diff
@@ -139,13 +141,20 @@ def load(self, urlpath):
...
+                print("Caching {}".format(file_in.path))
+                with tqdm_notebook(total=100) as pbar:
+                    with file_in as f1:
+                        with file_out as f2:
+                            data = True
+                            while data:
+                                data = f1.read(self.blocksize)
+                                f2.write(data)
+                                pbar.update(int(percent_per_block))
+                clear_output()
```
martindurant: I assume this would remove everything from the output area of the cell - and it looks like this happens after each file is done.

intake/source/cache.py
```diff
@@ -139,13 +141,20 @@ def load(self, urlpath):
...
                 self._log_metadata(urlpath, file_in.path, cache_path)
 
-                with file_in as f1:
-                    with file_out as f2:
-                        data = True
-                        while data:
-                            #TODO: print out progress
-                            data = f1.read(self.blocksize)
-                            f2.write(data)
+                file_size = file_in.fs.size(file_in.path)
```
martindurant: This will fail for HTTP

intake/source/cache.py
martindurant: I should say, it *may* fail for HTTP

intake/source/cache.py
mmccarty: No, HTTP implements `.size()`

martindurant: It does, but the server may well not supply the size when asked. If you look at the implementation, the behaviour is to try to cope with unknown sizes in a number of cases (particularly small files), but there is no general solution: HTTP is not FTP.

mmccarty: Or are you talking about the [ValueError](https://github.com/dask/dask/blob/master/dask/bytes/http.py#L312)?  I can catch that.

mmccarty: So, we ask for it and if the server doesn't provide it then we just don't show the progress bar.

martindurant: Yes.
Also the `.raise_for_status()` which can return any number of `requests` exceptions such as "I don't implement HEAD"; at the moment, those lower-level errors prevent any read, but they should be treated the same.

intake/source/cache.py
```diff
@@ -140,18 +140,23 @@ def load(self, urlpath):
...
+                with tqdm_notebook(total=100, leave=False, disable=pbar_disabled) as pbar:
                     with file_in as f1:
                         with file_out as f2:
                             data = True
                             while data:
                                 data = f1.read(self.blocksize)
                                 f2.write(data)
-                                pbar.update(int(percent_per_block))
-
+                                pbar.update(int(progress_block))
```
martindurant: `progress_block` undefined in case of ValueError above?

martindurant:
I think the failure is because of not including the new dependency.

martindurant:
Typo in the error message :)
""IntProgress not found. Please update juputer and ipywidgets."

martindurant:
OK, is good to me. All done?

mmccarty:
Yeah, all done.

PR #139, opened 2018-08-21
Parallelise caching, with updated progressbar

martindurant:
@mmccarty - not yet right, but a start

### DIFF
--- a/intake/source/cache.py
+++ b/intake/source/cache.py
@@ -147,7 +147,7 @@ class FileCache(object):
                 self._log_metadata(urlpath, file_in.path, cache_path)
                 ddown = dask.delayed(_download)
                 out.append(ddown(file_in, file_out, self.blocksize))
-        dask.compute(*out, scheduler='threads')
+        dask.compute(*out)
 
         return cache_paths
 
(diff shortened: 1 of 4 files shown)


martindurant:
out

intake/source/cache.py
```diff
@@ -31,6 +31,9 @@ def sanitize_path(path):
     return path
 
 
+display = {}
```
mmccarty: Why is ``display`` a dict key with ints instead of an array?

martindurant: Should probably be a set? I was using it differently before, but the values are no longer useful.

martindurant:
out

intake/source/cache.py
mmccarty: Yeah, based on the context it which it is used, I think a set would work better.

mmccarty:
Looks good!  Ready to merge?

martindurant:
Do you think the display parameter should be in the config and/or environment parameter and/or cache spec?

mmccarty:
You could add it to the config and env.  I wouldn't put it in the cache spec.

PR #141, opened 2018-08-22
Passing storage_options to caching.

mmccarty:


intake/source/base.py
```diff
@@ -79,7 +79,8 @@ def __init__(self, metadata=None):
         # default data
         self.metadata = metadata or {}
         if isinstance(self.metadata, dict):
-            self.cache = make_caches(self.name, self.metadata.get('cache'))
+            storage_options = self._captured_init_kwargs.get('storage_options', {})
```
mmccarty: Is this the right place to fetch `storage_options`?

mmccarty:
cc @jsignell

intake/source/base.py
martindurant: Calling it `storage_options` is a convention rather than spec, so i have a feeling it might be something that plugins should do, depending on what arguments they use, and passed in `load()`

mmccarty: Okay, even tho this is less desirable, let's go with it for now.  I can open an issue about the need to add arguments without the need to break the DataSource init API.  Sound okay?

martindurant: ok

mmccarty:
Issue #142 created.

PR #143, opened 2018-08-24
Initalizing file_size value when data source server does not return it.

mmccarty:


mmccarty:
@martindurant I'll probably merge this quickly, unless you have objections.

PR #145, opened 2018-08-28
Add config CLI functions

martindurant:


martindurant:
Testing should involve calling `intake` as a subprocess and capturing the output?

To add: cache functions (which should all be callable from code too): clear, list keys, list files for key, remove key, disc usage.

intake/source/cache.py
```diff
@@ -323,24 +334,27 @@ def _make_files(self, urlpath, **kwargs):
...
             [os.path.join(d, os.path.basename(f.path))], 'wb',
                                 **self._storage_options)[0]
              for f in files_in]
-        super(CompressedCache, self)._load(files_in, files_out, urlpath)
-        files_in = [f.path for f in files_out]
-        subdir = self._hash(urlpath)
-        return files_in, os.path.join(self._cache_dir, subdir)
+        super(CompressedCache, self)._load(files_in, files_out, urlpath,
+                                           meta=False)
+        # files_in = [f.path for f in files_out]
```
mmccarty: Probably should remove?

PR #155, opened 2018-09-07
Allow for cache_path="catdir" to put cached files alongside cats

### LINKED ISSUE #151: Configure Intake to cache files in the current directory
From @jbednar 

In most cases I'd rather have the files associated with the notebook or project where I'm working, not dumped into an ever-increasing global cache heap.
I would be nice to have a simple flag for configuring intake (something `intake.local_cache=True`) to store cached items in a subdirectory like `.intake_cache/` in the current directory.

martindurant:
Fixes #151 

If `cache_dir="catdir"`, resultant directory will be placed under the directory containing the catalog file.

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -208,8 +208,12 @@ Example:
     args:
       urlpath: '{{ CATALOG_DIR }}/cache_data/states.csv'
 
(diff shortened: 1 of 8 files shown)


martindurant:
@seibert , the idea was either/or, that it would be suitable to have *some* data sources cached to particular locations (e.g., the catalog directory) but not others.

PR #157, opened 2018-09-11
Filename parsing for csvs

jsignell:
Same concept as https://github.com/ContinuumIO/intake-xarray/pull/16 but for csvs. Moves the utils logic out of intake-xarray and into intake proper.


jsignell:
Rendered notebook at https://anaconda.org/jsignell/model_precip_data

martindurant:
Is this reimplementing the functionality recently merged into dask, or merely making use of it?

jsignell:
> Is this reimplementing the functionality recently merged into dask, or merely making use of it?

Making use of it. And expanding on it. That reminds me this would require latest released dask.

martindurant:
Seems to pass anyway :)

jsignell:
Travis is getting a new enough version of dask:
https://travis-ci.org/ContinuumIO/intake/builds/427278384#L1067

martindurant:
Also https://github.com/martindurant/filesystem_spec/pull/13

intake/source/csv.py
```diff
@@ -31,23 +32,47 @@ def __init__(self, urlpath, csv_kwargs=None, metadata=None,
...
+        self._urlpath = pattern_to_glob(urlpath)
+        self._pattern = urlpath
         self._storage_options = storage_options
         self._csv_kwargs = csv_kwargs or {}
         self._dataframe = None
 
         super(CSVSource, self).__init__(metadata=metadata)
 
+    def _get_filname_args(self, include_path_column):
+        col = include_path_column
```
martindurant: I don't really get what's happening here. Perhaps it would be best to include a description of what can happen in urlpath, and at least one example, the in docstring in init.

intake/source/utils.py
```diff
@@ -0,0 +1,87 @@
+def _get_parts_of_format_string(resolved_string, literal_texts, format_specs):
```
martindurant: I gather the intent of putting the functions here would be that they could be generally useful for other file-to-dataframe plugins. Therefore, I think it's worth-while documenting the functions here, and including a section in making-plugins.html next to Remote Data.

jsignell:
>Ought there to be an updated version constraint on dask, or can we make a nice error if dask is too old, and this functionality is requested

Yes totally. Also what should happen if the user tries to use a this template style for an unsupported backend? Should it raise a `Not Implemented` error? And if so how?

martindurant:
A kwarg to the two plugins would be fine; I wouldn't put it in the global config. Of the remaining plugins, I don't know how many would qualify - perhaps avro only?
There is scope for having conventions here, like the storage_options kwarg - and the more prominent the conventions are, the easier our lives will be.

intake/source/utils.py
jsignell: do you think some docs should also go in catalog? And if so where?

martindurant: Probably not, it's not general to catalogues, only used for a couple of specific plugins.
We could consider separate "spec" and "conventions" pages to describe what is required and what is likely to be found in catalog descriptions.

docs/source/making-plugins.rst
```diff
@@ -246,3 +246,60 @@ To use an ``OpenFile`` object, make it concrete by using a context:
...
+                container='dataframe',
+                metadata=metadata
+            )
+
+        def _get_schema(self):
+            # get list of file paths
+            for path in file_paths:
+                # read in the file
+                values_by_field = reverse_format(self.pattern, path)
+                # add these fields and values to the data
```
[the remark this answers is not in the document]
jsignell: In that case should I make a helper function that takes a pattern and a list of paths and returns something of the form: `[{ field: {path: value}}, {field: {path: value}}]`?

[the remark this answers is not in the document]
martindurant: Sure sounds reasonable, and very amenable to clean testing.

martindurant:
@jsignell , anything left to do here?

jsignell:
> anything left to do here?

If you are happy with it, I'm happy with it. There aren't any docs on usage other than the example though. Also not sure about best practices for including example data, but these csvs are pretty small.

martindurant:
I'm happy with it. Once you've written the blog, that can be linked in the docs too. I imagine we'll need to explicitly point people to this when they ask for similar functionality, because it's not something that most will have thought of.

jsignell:
👍

PR #160, opened 2018-09-13
Raise is attempt to open cat file which cannot be read

### LINKED ISSUE #159: Inaccessible file path for data catalog fails silently
When trying to load a catalog yml file hosted S3, missing credentials will silently lead to an empty catalog. So the user executes,

```
In [1]: from intake import Catalog

In [2]: cat = Catalog("s3://foo/bar.yml")

In [3]: cat.yaml()
Out[3]: "sources:\n  bar:\n    args:\n      path: s3://foo/bar.yml\n    description: ''\n    driver: yaml_files_cat\n    metadata: {}\n"

```
and it's not obvious why there are no sources. 

For the case of s3 specifically, a quick check looks like:
```
def verify_credentials(cat_uri):
    """
    FIXME copied from data-layer-api
    The intake catalog will appear empty when trying to read a catalog yml
    file on s3 if the credentials are not supplied correctly.
    """
    if "s3:" not in cat_uri:
        return True

    fs = s3fs.S3FileSystem()
    cat_exists = fs.exists(cat_uri)
    if not cat_exists:
        err_msg = "Catalog file not found at {}. Check AWS credentials.".format(cat_uri)
        raise EnvironmentError(err_msg)
```

There are probably issues related to this as well if trying to open a local yml file in a directory that doesn't exist.

martindurant:
Does not affect reading a *directory*, which will still return an
empty cat if does not exist, and will update automatically as before

Fixes #159

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -494,11 +494,8 @@ class YAMLFileCatalog(Catalog):
             self.path))[0].replace('.', '_')
         self._dir = get_dir(self.path)
 
-        try:
-            with file_open as f:
-                text = f.read().decode()
-        except (IOError, OSError):
-            return
+        with file_open as f:
+            text = f.read().decode()
         if "!template " in text:
             logger.warning("Use of '!template' deprecated - fixing")
             text = text.replace('!template ', '')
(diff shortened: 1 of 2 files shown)


martindurant:
@jmabry , @seibert , thoughts?

jmabry:
@martindurant I checked out your branch and tested the missing AWS credentials scenario described in the issue.  I now get a `FileNotFoundError` which makes sense.  Glad to see it fixed without needing any new args to Catalog.  

A few comments explaining some of the logic in `_load()` could make it easier for newcomers like myself to contribute back to this part of the codebase or otherwise troubleshoot. You could also think about factoring out some of the operations.  Not trying to complain, more just asking for some better guideposts so I can be more helpful in the future. Thanks for the fast resolution to this issue!

martindurant:
Thanks, @jmabry .
Generally agree that the code can be better structured and explained. You are welcome to contribute, if you are able, but I will try to keep this in mind for the future.

PR #162, opened 2018-09-17
Fixing urlpath = list of files for csv plugin

jsignell:
I introduced a bug as part of #157 that this fixes and adds a test for.

### DIFF
--- a/intake/source/utils.py
+++ b/intake/source/utils.py
@@ -47,9 +47,9 @@ def reverse_formats(format_string, resolved_strings):
     Reverse the string method format for a list of strings.
 
     Given format_string and resolved_strings, for each resolved string
-    find arguments that would give 
-    ``format_string.format(**arguments) == resolved_string``. 
-    
+    find arguments that would give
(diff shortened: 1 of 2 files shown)


intake/source/utils.py
```diff
@@ -210,6 +212,9 @@ def path_to_glob(path):
 
     fmt = Formatter()
 
+    if not isinstance(path, str):
+        return path
```
martindurant: Do you think it's worth being more careful here, checking for specific types (tuple, list), that it's iterable, or something else?

jsignell: I don't just because this function is called right away on the urlpath and is only useful for converting from something like `data{field}.csv` to `data*.csv`. In any plugin using this path as a pattern approach, we would check if `path_to_glob(urlpath) == urlpath` to decide if there were any pattern fields in the path. So by just returning path we ensure that the path won't be treated as a pattern.

martindurant: OK, I can agree with that.

PR #163, opened 2018-09-19
Collision alleviation

### LINKED ISSUE #161: Plugin name collisions
First off thanks so much for your awesome work on Intake!

Today I've been developing an [Intake Plugin for Iris](https://github.com/informatics-lab/intake-iris) and noticed that if I create a `netcdf` plugin it collides with the xarray plugin and Intake only loads one of them at import time. 

image

Luckily for my testing iris comes before xarray alphabetically so I have been able to test my plugin without removing the xarray one. 

Given that both iris and xarray can open netcdf datasets I think this choice should probably be down to the user as to which one they want it to be loaded with. It would be great to update intake to be able to load multiple plugins for the same datatype and then allow the user to pass a kwarg somewhere to specify which one they want to use at a particular moment.

martindurant:
Fixes #161 (eventually)

Initially, this is just documentation of the spec that catalogues might have. How does this seem as a solution?

(Note that I have not made the "alias" type mentioned in the issue, I view that as separate and could be introduced in a separate PR)

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -135,6 +135,67 @@ available for use in template strings, which can be used to alter the arguments 
 a data source might accept a "postal_code" argument which is used to alter a database query, or select a particular
 group within a file.  Users set parameters with keyword arguments to the ``get()`` method on the catalog object.
 
+Driver Selection
+''''''''''''''''
+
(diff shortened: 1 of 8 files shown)


jacobtomlinson:
This is awesome!

PR #168, opened 2018-09-27
Add INTAKE_PATH default catalog loading

### LINKED ISSUE #167: Add a path style lookup for catalog locations
Currently intake checks your user dir and conda dir for manifest files to put into the default `cat`. 

It would be great if this behaved like a path and you could tack extra ones on with environment variables. 

For example we want to add some extra manifests, we don't want to package these with conda and put them in the conda path and we want them to be available to everyone. So it would be great if we could add an extra location for manifests using an environment variable to point intake towards a network share or something.

I propose something like `INTAKE_PATH` which could be optionally set and would follow the format of `/some/path/:/some/other/path/:/somewhere/else/again/`. Then during the default loading it could check that each of these exist and load any catalogs inside as part of `intake.catalog.default.load_combo_catalog()`.

jacobtomlinson:
Closes #167.

This PR updates the default catalog loader to check for any directories specified by the `INTAKE_PATH` environment variable. This behaves like a regular path where each directory should be colon separated. If each directory exists it will be included in the loading of default catalog files.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -154,7 +154,7 @@ installed in user-specific location.
 
 Adding Data Source Packages using the Intake path
 -------------------------------------------------
-Intake checks the environment variable `INTAKE_PATH` for a colon separated list of paths to search for catalog files.
+Intake checks the environment variable `INTAKE_PATH` for a colon separated list of paths (semicolon on windows) to search for catalog files.
 When you import ``intake`` we will see data from catalog files in the path appear as part of a global catalog
 called ``intake.cat``.
 
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -151,6 +151,14 @@ Conda installs the catalog file in this package to ``$CONDA_PREFIX/share/intake/
 The global catalog is a union of all catalogs installed in the conda/virtualenv environment and also any catalogs
 installed in user-specific location.
 
+
+Adding Data Source Packages using the Intake path
+-------------------------------------------------
+Intake checks the environment variable `INTAKE_PATH` for a colon separated list of paths to search for catalog files.
+When you import ``intake`` we will see data from catalog files in the path appear as part of a global catalog
+called ``intake.cat``.
+
+
 Using the GUI
 -------------
 
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -154,8 +154,8 @@ installed in user-specific location.
 
 Adding Data Source Packages using the Intake path
 -------------------------------------------------
-Intake checks the environment variable `INTAKE_PATH` for a colon separated list of paths (semicolon on windows) to search for catalog files.
-When you import ``intake`` we will see data from catalog files in the path appear as part of a global catalog
+Intake checks the Intake config file for `catalog_path` or the environment variable `INTAKE_PATH` for a colon separated list of paths (semicolon on windows) to search for catalog files.
+When you import ``intake`` we will see all entries from all of the catalogues referenced as part of a global catalog
 called ``intake.cat``.
 
 
--- a/intake/catalog/default.py
+++ b/intake/catalog/default.py
@@ -98,8 +98,10 @@ def load_combo_catalog():
         cat_dirs.append(global_dir + '/*.yaml')
         cat_dirs.append(global_dir + '/*.yml')
     for path_dir in conf.get('catalog_path', []):
-        if os.path.isdir(path_dir):
+        if os.path.isdir(path_dir) and path_dir != '':
             cat_dirs.append(path_dir + '/*.yaml')
             cat_dirs.append(path_dir + '/*.yml')
+
+    print(cat_dirs)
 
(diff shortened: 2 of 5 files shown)


martindurant:
This sounds like a very reasonable thing to want to do.

I have a couple of worries here: 
- colons make poor separators on windows, where a colon may be part of a path. Perhaps the syntax needs to be a JSON list?
- I would add this as a config file attribute, which may be the simplest way for a user/sysadmin to pass the information in.
- should we allow remote directories or specific remote catalogue files here? I think that, as things stand, remotes would be loaded (except that the paths always contain colons).

seibert:
For ease of use, I think we have to use the path conventions on each platform: colons for Mac/Linux and semicolons on Windows.  I think remote directories might be too confusing for an environment variable named `*_PATH`.

martindurant:
By the way, the failure in travis is related to the package upload after testing is done, apparently the secure token isn't being resolved. I am tempted to remove the upload altogether and handle releases manually. Conda-forge should certainly be the long-term aim, but that has higher overheads per release, which I expect to still be frequent.

jacobtomlinson:
I've added windows semicolon support and added the option to the config as `catalog_path` but have overridden from the environment still.

docs/source/quickstart.rst
```diff
@@ -151,9 +151,17 @@ Conda installs the catalog file in this package to ``$CONDA_PREFIX/share/intake/
 The global catalog is a union of all catalogs installed in the conda/virtualenv environment and also any catalogs
 installed in user-specific location.
 
+
+Adding Data Source Packages using the Intake path
+-------------------------------------------------
+Intake checks the environment variable `INTAKE_PATH` for a colon separated list of paths (semicolon on windows) to search for catalog files.
```
martindurant: Can you please update this to specify the config file option, and that the paths should be *local* (for now). Actually, should the config variant strictly follow the same syntax, or could it have normal YAML syntax, in which case it could easily be remote paths too?
In this one specific case, do you think that the environment variable should override, or is there an argument that it should only add? Normal BASH path lists do override, even if the typical use case includes the list as defined so far.

docs/source/quickstart.rst
```diff
@@ -151,9 +151,17 @@ Conda installs the catalog file in this package to ``$CONDA_PREFIX/share/intake/
...
+When you import ``intake`` we will see data from catalog files in the path appear as part of a global catalog
```
martindurant: All entries from all of the catalogues referenced

intake/tests/test_top_level.py
```diff
@@ -47,11 +56,18 @@ def test_user_catalog(user_catalog):
     assert set(cat) >= set(['ex1', 'ex2'])
 
 
+def test_path_catalog(tmp_path_catalog):
+    intake.config.conf['catalog_path'] = os.path.join(tempfile.gettempdir(), 'intake')
+    cat = intake.load_combo_catalog()
+    time.sleep(2) # wait 2 seconds for catalog to refresh
```
martindurant: This shouldn't be necessary, it refreshes on instantiation.

docs/source/quickstart.rst
jacobtomlinson: I was thinking the same thing. The config option could be a YAML list, and when the path env var gets unpacked we then turn it into a list. I do think that the env var should override the config.

intake/tests/test_top_level.py
jacobtomlinson: Ah ok! I was just copying the other test.

martindurant:
Well, it seems I was wrong about removing the sleep, since now one of the tests fails. I'll look into why that is later, but you can just reinstate the sleep for now.
Also, feel free to remove the package upload step - either completely, or with an extra condition on the availability of the token.

jacobtomlinson:
Done and done!

jacobtomlinson:
@martindurant I'm not sure I understand what all these test error are. Do you have any thoughts?

intake/catalog/default.py
```diff
@@ -96,5 +97,11 @@ def load_combo_catalog():
     if os.path.isdir(global_dir):
         cat_dirs.append(global_dir + '/*.yaml')
         cat_dirs.append(global_dir + '/*.yml')
+    for path_dir in conf.get('catalog_path', []):
+        if os.path.isdir(path_dir) and path_dir != '':
+            cat_dirs.append(path_dir + '/*.yaml')
+            cat_dirs.append(path_dir + '/*.yml')
+
+    print(cat_dirs)
```
martindurant: Leftover

martindurant:
No, I can''t think of anything obvious. Perhaps remove that print, and see if it's just "one of those things"?

jacobtomlinson:
I had forgotten about the print. Hopefully it is something as daft as that!

martindurant:
OK, it's good.
I can merge this now, if you are done.

jacobtomlinson:
Yes please!

martindurant:
Done. Thanks for this.

PR #172, opened 2018-09-30
Add search interface to catalogues

### LINKED ISSUE #169: Search functionality
I intend to make a search method on a catalogue, and corresponding search box in the GUI. This issue is to gather any opinions before I start.

I am thinking to start with simple text search to begin with
```
any([word in entry.describe() for word in search.split()])
```

A search function would take a list of catalogues, and default to the builtin ``intake.cat``, and the user would need to consider what depth to search when there is a cat hierarchy.

martindurant:
Fixes #169 

The POC here has the following functionality, as a demonstration:

`cat.search("some text")` will return a new catalogue, whose entries are the entries of `cat`, where either "some" or "text" appears in any of the fields returned by `.describe()`.

Questions:
- could have also included `.describe_open()` to check, as absolute arguments and metadata usually end up here
- is it right to have the search return a new, limited catalogue
- this is a simple implementation, and won't change in the case that the cat it is based on reloads. Seems OK to have to search again. To enable reload, there would need to be a SearchedCatalog class.
- how should this be incorporated into the GUI?


martindurant:
Also, we don't "discover" the entries, which would take some time, and means that we can't search in the details of the data sources, such as column names.

martindurant:
screen shot 2018-10-01 at 16 01 31

martindurant:
Surprises of the search functionality (at the moment): 
- searching is within the currently selected cat
- if your search finds nothing, there is no warning
- if you search again on something you searched before, the entry in the table is overwritten.

martindurant:
Additional note on the interface in general: the (-) button does remove cats, as it should. It also can remove "builtin", which you have no way to get back if you delete it. I think I would call this fully-functional at this point, since I don't yet know what behaviour users would want concerning this and the questions above.

PR #175, opened 2018-10-02
Added the ability to set width on strings

jsignell:
This is needed in the context of string fields where there is no separator between fields. See added tests.

### DIFF
--- a/intake/source/tests/test_utils.py
+++ b/intake/source/tests/test_utils.py
@@ -26,6 +26,8 @@ def test_path_to_glob(pattern, expected):
     {'year': 2016, 'month': 2, 'day': 1}),
     ('{year:.4}/{month:.2}/{day:.2}.csv', '2016/2/01.csv',
     {'year': '2016', 'month': '2', 'day': '01'}),
+    ('{year:.4}{month:.2}{day:.2}.csv', '20160201.csv',
+    {'year': '2016', 'month': '02', 'day': '01'}),
     ('SRLCCTabularDat/Ecoregions_{emissions}_Precip_{model}.csv',
      '/user/examples/SRLCCTabularDat/Ecoregions_a1b_Precip_ECHAM5-MPI.csv',
      {'emissions':'a1b', 'model': 'ECHAM5-MPI'}),
@@ -56,7 +58,7 @@ def test_reverse_format_errors():
                                             "if no separator between fields.")):
         reverse_format(pattern, resolved)
 
-    pattern = '{month:.2}{day:2}{year:.4}'
+    pattern = '{month:.2d}{day:2d}{year:.4d}'
     resolved = '20012001'
     with pytest.raises(ValueError, message="Format specifier must have a set width"):
         reverse_format(pattern, resolved)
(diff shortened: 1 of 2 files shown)


intake/source/tests/test_utils.py
```diff
@@ -56,7 +58,7 @@ def test_reverse_format_errors():
                                             "if no separator between fields.")):
         reverse_format(pattern, resolved)
 
-    pattern = '{month:.2}{day:2}{year:.4}'
+    pattern = '{month:.2d}{day:2d}{year:.4d}'
```
martindurant: What is the right pattern here?

jsignell: I was trying to only allow fixed width strings, but I can't figure out how to make them fixed width.

martindurant: If I follow, ".2d" means two decimal places, but total width unknown; so the right thing here would be `'{month:2d}{day:2d}{year:4d}'` (no decimals; these are all integers anyway; could be 'i' instead of 'd'). However, such formatting is generally "minimum" width, so I'm not sure there's a general solution.

jsignell: Yeah that is minimum width and the string version of minimum width is `'{a:.2d}'`

```python
>>>'{:.2}'.format('l')
'l'
>>>'{:.2}'.format('loc')
'lo'
>>>'{:02d}'.format(4)
'04'
>>>'{:02d}'.format(400)
'400'
```

jsignell: so... it isn't explicit how to split the string if there is no separator. I was thinking I'd just treat both `'{:02d}'` and `'{:.2}'` as having mandatory len 2.

jsignell: I'll treat `'{:2}'` as having mandatory len 2 as well even though in practice it has no impact on the string formatting.

martindurant: Regex will let you have patterns like "at least as long as", "with x chars after the decimal", etc. and then see if the whole pattern, including the fixed-with parts scans. It really feels like this kinf of matching must have already been done elsewhere.
Reminds me: we may want to also allow real regexes, which are probably more flexible, but only match with string outputs.

jsignell: The main reason why I did string formatting rather than regex is to make the fields name-able

jsignell: hmm maybe something like: https://github.com/r1chardj0n3s/parse

RutgerK: FYI: Trollsift does something similar: https://github.com/pytroll/trollsift
It's used for example by SatPy to find/parse filenames as specified in a config file, look for an example at the `file_patterns` part in https://github.com/pytroll/satpy/blob/master/satpy/etc/readers/abi_l1b.yaml.

martindurant: I see that satpy also has a loader/argument YAML specification in that directory and many entries. What do you think the overlap with Intake can be? I'm thinking the files could be directly parsed into Intake catalogues without modification - would that be useful?

martindurant: @jsignell , trollsift is available on conda-forge, so I would just try it, to see if their parser meets all of the filename requirements.

jsignell:
Closing in favor of replacing parser with external dependency: https://github.com/ContinuumIO/intake/issues/181

Issue #170, opened 2018-09-28
Plugins depending on other datasets


martindurant:
This idea is to allow for a data source type, where one or more of the inputs are data-sets within the same catalogue. Probably, this could be achieved simplest by having the catalogue object itself attached to any source created from it (so this would not apply to `open_*` methods, unless you explicitly pass a catalogue).

The plugin, being ordinary python code, if free to process the input as it sees fit, but probably the name of the data-set(s) to work on are explicit arguments. I expect that any derrived data-set class only applied to one input container type, but there's no reason that the output container type need be the same.

The simplest form of this plugin would be "alias", mentioned in https://github.com/ContinuumIO/intake/issues/161 , where the single input would map to data-sets in the same catalogue, which would be passed through without processing.

(This is NOT general data pipelining, the functions to be applied are fixed in the plugin code, cannot be specified in the catalogue spec)

PR #176, opened 2018-10-02
Derived data source plugins

martindurant:
- [x] Pass catalog to data sources
- [x] Added AliasSource, which is a simple pass-through and the first demonstration
of a derived dataset plugin.
- [x] Tests
- [x] Docs


martindurant:
cc @seibert

martindurant:
cc @niallrobinson ; @jacobtomlinson was interested in the idea of a Alias data source type.

martindurant:
Fixes #170

PR #179, opened 2018-10-03
Making dask fields from pattern known categoricals

### LINKED ISSUE #178: In CSV plugin, `path_as_pattern` field columns should be known categories
In the path column on dask the paths are known categories, which is why they are parse-able before the data are read. It would be good if the `path_as_pattern` fields were also known in this same way. It is a little tricky because the mapping from paths to unique pattern-field values is not necessarily 1:1 so we can't just use the same cat.codes. I guess what we should do is declare a new category index with new codes  and field values and then map path codes to field codes. That should work.

jsignell:
Fixes #178 and adds test

### DIFF
--- a/intake/source/csv.py
+++ b/intake/source/csv.py
@@ -55,6 +55,8 @@ class CSVSource(base.DataSource, base.PatternMixin):
         """Get a column of values for each field in pattern
         """
         try:
+            # CategoricalDtype allows specifying known categories when
+            # creating objects. It was added in pandas 0.21.0.
             from pandas.api.types import CategoricalDtype
             _HAS_CDT = True
         except ImportError:
--- a/intake/source/csv.py
+++ b/intake/source/csv.py
@@ -54,12 +54,19 @@ class CSVSource(base.DataSource, base.PatternMixin):
     def _set_pattern_columns(self, path_column):
         """Get a column of values for each field in pattern
         """
+        try:
+            from pandas.api.types import CategoricalDtype
+            _HAS_CDT = True
+        except ImportError:
+            _HAS_CDT = False
+
         col = self._dataframe[path_column]
         paths = col.cat.categories
 
         column_by_field = {field:
-            col.cat.codes.map(dict(enumerate(values))).astype("category") for
-            field, values in reverse_formats(self.pattern, paths).items()
+            col.cat.codes.map(dict(enumerate(values))).astype(
+                "category" if not _HAS_CDT else CategoricalDtype(set(values))
+            ) for field, values in reverse_formats(self.pattern, paths).items()
         }
         self._dataframe = self._dataframe.assign(**column_by_field)
 
--- a/intake/source/tests/test_csv.py
+++ b/intake/source/tests/test_csv.py
@@ -98,6 +98,10 @@ def test_read_chunked(sample1_datasource, data_filenames):
 
 
 def check_read_pattern_output(source):
+    da = source.to_dask()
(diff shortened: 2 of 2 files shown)


intake/source/csv.py
```diff
@@ -54,12 +54,19 @@ def __init__(self, urlpath, csv_kwargs=None, metadata=None,
...
+
         col = self._dataframe[path_column]
         paths = col.cat.categories
 
         column_by_field = {field:
-            col.cat.codes.map(dict(enumerate(values))).astype("category") for
-            field, values in reverse_formats(self.pattern, paths).items()
+            col.cat.codes.map(dict(enumerate(values))).astype(
+                "category" if not _HAS_CDT else CategoricalDtype(set(values))
+            ) for field, values in reverse_formats(self.pattern, paths).items()
```
martindurant: So the original type loaded by Dask is still string, but we convert to get categoricals at the next stage?

intake/source/tests/test_csv.py
```diff
@@ -98,6 +98,10 @@ def test_read_chunked(sample1_datasource, data_filenames):
 
 
 def check_read_pattern_output(source):
+    da = source.to_dask()
+    assert da.num.cat.known is True
```
martindurant: Presumably we can check the specific values of `da.num.cat` (the metadata), and also grab a partition and (`.get_partition(0).compute()`) and check that it also has the right categories.

intake/source/csv.py
```diff
@@ -54,12 +54,19 @@ def __init__(self, urlpath, csv_kwargs=None, metadata=None,
     def _set_pattern_columns(self, path_column):
         """Get a column of values for each field in pattern
         """
+        try:
+            from pandas.api.types import CategoricalDtype
+            _HAS_CDT = True
```
martindurant: This is a Pandas version thing? Maybe nice to have a clarifying comment.

jsignell: Right it is >=0.21.0 https://pandas.pydata.org/pandas-docs/stable/generated/pandas.api.types.CategoricalDtype.html#pandas.api.types.CategoricalDtype

intake/source/tests/test_csv.py
jsignell: yeah we can. We check all that below though right?

intake/source/csv.py
jsignell: no sure I understand the distinction.

intake/source/tests/test_csv.py
martindurant: No, below we effectively do `.compute()` and then check everything on the resulting dataframe. I'm suggesting we should also check the `df` metadata before compute, and the contents of one of the partitions. It should be the same, of course.

intake/source/csv.py
martindurant: I think I was misunderstanding what happens here. Is the following right:
upon load, dask applies the filename pieces to each chunk, to create categorical columns which always have just a  single label within each partition; then in this step we change the labels to real values based on the parsing. Is that right? In that case, everything is fine.

martindurant:
Closes #178

intake/source/csv.py
jsignell: fixed

intake/source/tests/test_csv.py
jsignell: done

jsignell:
Ok I think this is ready

Issue #185, opened 2018-10-24
import error


timothydmorton:
On the way to trying to reproduce locally an issue I was just about to raise, I get the following after conda installing intake and trying to import:
```
Python 3.6.5 |Anaconda, Inc.| (default, Apr 26 2018, 08:42:37)
Type "copyright", "credits" or "license" for more information.

IPython 5.6.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import intake
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-924f1ec0a922> in <module>()
----> 1 import intake

/Users/tdm/miniconda3/lib/python3.6/site-packages/intake/__init__.py in <module>()
     12 from .source import registry
     13 from .source.discovery import autodiscover
---> 14 from .gui import DataBrowser
     15
     16 # Populate list of autodetected plugins

/Users/tdm/miniconda3/lib/python3.6/site-packages/intake/gui/__init__.py in <module>()
      1 try:
      2     import ipywidgets
----> 3     from .widgets import DataBrowser
      4
      5 except ImportError:

/Users/tdm/miniconda3/lib/python3.6/site-packages/intake/gui/widgets.py in <module>()
      8 here = os.path.abspath(os.path.dirname(__file__))
      9 logo_file = os.path.join(here, 'logo.png')
---> 10 logo = widgets.Box([widgets.Image.from_file(logo_file)],
     11                    layout=widgets.Layout(width='46px', height='35px',
     12                                          overflow_y='visible',

AttributeError: type object 'Image' has no attribute 'from_file'```

timothydmorton:
OK, looks like my `ipywidgets` was out of date.  Should probably update the intake dependencies accordingly:
```ipywidgets:         7.1.0-py36_0 --> 7.4.2-py36_0```

PR #186, opened 2018-10-24
Put ipywidgets dependency in requirements

timothydmorton:
I had `ipywidgets` v7.1 and ran into the import error in #185. Upgrading to 7.4 fixed it, and I confirmed that at least `import intake` works with  >= 7.2.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ six
 tornado >= 4.5.1
 dask[complete]
 holoviews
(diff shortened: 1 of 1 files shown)


martindurant:
Fixes #185 

Thank you, @timothydmorton . This wasn't done initially, since I didn't want to require ipywidgets, but Intake has quite a heavy set of dependencies, this is pretty light by comparison.

PR #188, opened 2018-10-24
Fix for missing plugin situation

### LINKED ISSUE #187: get_plugin_class('blahblah') fails silently
I tried to intake a parquet dataset before having installed `intake-parquet`, got an error, and traced it to `get_plugin_class` failing silently and returning `None`.  Or, for example:
```
In [1]: from intake.source import get_plugin_class

In [2]: get_plugin_class('blahblah')

In [3]:```

### PR DESCRIPTION
**Pull request by martindurant**
Fixes #187

The intended behaviour is that the entries should load OK, but that
the plugin attribute contains None, so that an attempt to *access*
the entry as a data source fails, but other entries are unaffected.

The message you get is like 
```
~/code/intake/intake/catalog/local.py in get(self, **user_parameters)
    235         if len(self._plugin) == 0:
    236             raise ValueError('No plugins loaded for this entry: %s'
--> 237                              % self._driver)
    238         elif isinstance(self._plugin, list):
    239             plugin = self._plugin[0]

ValueError: No plugins loaded for this entry: parquet
```
(although Intake has no way of knowing which conda package might contain the named driver)

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -158,8 +158,9 @@ class LocalCatalogEntry(CatalogEntry):
         self._catalog_dir = catalog_dir
         self._catalog = catalog
         if isinstance(driver, str):
-            self._plugin = [get_plugin_class(driver)]
-            containers = set([self._plugin[0].container])
+            dr = get_plugin_class(driver)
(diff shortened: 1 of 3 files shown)


martindurant:
Thanks for the report, @timothydmorton

Issue #198, opened 2018-11-21
intake server documentation


delgadom:
The documentation references the ability to set up an intake server, but does not point to any resources about how to set this up. do docs on server setup exist somewhere?

martindurant:
The documentation is [here](https://intake.readthedocs.io/en/latest/tools.html#intake-server), perhaps a little hidden and certainly could be expanded. 

Probably the protocol itself should be described in detail, though a glance at [the implementation](https://github.com/ContinuumIO/intake/blob/master/intake/cli/server/server.py#L20) shows it is pretty simple. As I have been telling people, the provided server could be seen as a prototype/example, and if the concept proves useful, I expect more server implementations to appear.

martindurant:
@delgadom , I am interested, though: what is your use case?

delgadom:
Oh thank you @martindurant! Yes I had completely missed this in reading through the docs. Maybe some internal links from other pages would be helpful, if not a whole section (not sure if you think this warrants this).

Still getting spun up on this (evaluating using this for our research team), but are there types of storage formats available to the server besides yml? Where can one find information about server configuration? Is a source dive necessary here?

Also if there are any example deployments you think are particularly helpful a link would be great. E.g. is there a pangeo deployment that could serve as a particularly useful intake server deployment?

I'd be happy to add to the docs as I work through these questions if it would be helpful.

delgadom:
Re: our use case, we have two pangeo-style kubernetes clusters with 10 or so active users. We split off of the core pangeo development process because we needed some core features really early on that weren't supported and have been too busy to rebase to the current deployments, but intend to rejoin the flock if possible soon. Looking to see if intake would be a good data catalogue & versioning system for that deployment, and perhaps also for our other computing resources (30 active users spread across the Berkeley high performance computing center, a standalone Linux server at Berkeley, and some nodes at the open science data cloud at UChicago). We do climate economic impact research, so our data is pretty typical geoscience data with 1000s of frequently accessed netcdf, zarr, CSV, parquet files, along with other less standard file types (e.g. stats .dat and Matlab .mat files).

martindurant:
Agree, that it should have its own section and certainly better internal linking. For exampe, "Intake Server" appears in the glossary, but doesn't include the link I gave here.

> are there types of storage formats available to the server besides yml

[...]

PR #200, opened 2018-11-28
Add pagination support to server and RemoteCatalog

danielballan:
Thanks for this excellent project!

I have a Catalog with ~10k entries. (It happens to be backed by MongoDB, not YAML.) To make this scalable, I need to avoid pulling down all the entries from the server. This PR adds support for optional pagination parameters to the server's `GET /info` route and adds a new `GET /source` method to get the 'info' about one source. Respectively, these two server changes allow the `RemoteSource` to fetch entry info in batches during iteration (only pulling down as many as are used) and to fetch any specific entry's info by name (random access) when called on by `__getitem__`.

I have added some pagination unit tests. I have also interactively tested a client running on the `master` branch against a server running on this branch and vice versa. Each side gracefully falls back to non-paginated behavior when its peer does not support pagination.

In commit b1dabc4 I removed a redundant lookup but also made a slight API change: the Catalog metadata must now be accessed by attribute, not by `__getitem__`. Two small changes to existing tests were necessary to adjust for this. I believe this is an improvement. It is nice that entries are accessible by attribute, but I believe the inverse should _not_ be supported, lest the line between the library's core interface and its entries become too blurry and lead to confusing Tracebacks. If this change is controversial I am happy to rip it out into separate PR for further discussion.

Thanks in advance for taking a look.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -198,8 +198,7 @@ class Catalog(DataSource):
 
 class RemoteCatalog(Catalog):
     """The state of a remote Intake server"""
-
-    def __init__(self, url, http_args={}, **kwargs):
+    def __init__(self, url, http_args={}, page_size=100, **kwargs):
         """Connect to remote Intake Server as a catalog
 
         Parameters
@@ -209,12 +208,16 @@ class RemoteCatalog(Catalog):
         http_args: dict
             Arguments to add to HTTP calls, including "ssl" (True/False) for
             secure connections.
+        page_size : int, optional
+            The number of entries fetched at a time during iteration.
+            Default is 100.
         kwargs: may include catalog name, metadata, source ID (if known) and
             auth instance.
         """
         self.http_args = http_args
         self.http_args.update(kwargs.get('storage_options', {}))
         self.http_args['headers'] = self.http_args.get('headers', {})
+        self.page_size = page_size
         self._source_id = kwargs.get('source_id', None)
         if self._source_id is None:
             secure = http_args.pop('ssl', False)
@@ -230,11 +233,10 @@ class RemoteCatalog(Catalog):
             self.info_url = url.replace('v1/source', 'v1/info')
         self.auth = kwargs.get('auth', None)  # instance of BaseClientAuth
 
-        PAGE_SIZE = 100
-
         def fetch_page(page_number):
             logger.debug("Request page %d of entries", page_number)
-            params = {'page[number]': page_number, 'page[size]': PAGE_SIZE}
+            params = {'page[number]': page_number,
+                      'page[size]': self.page_size}
             response = requests.get(self.info_url, params=params,
                                     **self._get_http_args())
             if response.status_code != 200:
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -160,6 +160,15 @@ class Catalog(DataSource):
         """Return an iterator over catalog entries."""
         return iter(self._get_entries())
 
+    def __contains__(self, key):
+        # Avoid iterating through all entries.
+        try:
+            self[key]
+        except KeyError:
+            return False
+        else:
+            return True
+
     def __dir__(self):
         return list(self)
 
@@ -175,17 +184,16 @@ class Catalog(DataSource):
         
         Can also use attribute syntax, like ``cat.entry_name``.
         """
-        try:
+        if key in self._entries:
             return self._entries[key]
-        except KeyError:
-            out = self
-            for k in key.split('.'):
-                if isinstance(out, CatalogEntry):
-                    out = out()  # default parameters
-                if not isinstance(out, Catalog):
-                    raise ValueError("Attempt to recurse into non-catalog")
-                out = getattr(out, k)
-            return out
+        out = self
+        for k in key.split('.'):
+            if isinstance(out, CatalogEntry):
+                out = out()  # default parameters
+            if not isinstance(out, Catalog):
+                raise ValueError("Attempt to recurse into non-catalog")
+            out = getattr(out, k)
+        return out
 
     def discover(self):
         return {"container": 'catalog', 'shape': None,
@@ -303,6 +311,15 @@ class RemoteCatalog(Catalog):
             def __iter__(self):
                 for key in self.keys():
                     yield key
+
+            def __contains__(self, key):
+                # Avoid iterating through all entries.
+                try:
+                    self[key]
+                except KeyError:
+                    return False
+                else:
+                    return True
 
             def items(self):
                 for item in six.iteritems(self._page_cache):
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -268,6 +268,9 @@ class RemoteCatalog(Catalog):
                 auth=self.auth,
                 http_args=self.http_args, **info['source'])
 
+        def server_can_paginate():
+            return self._server_can_paginate
+
         class Entries:
             """Mock enough of the dict interface.
 
@@ -292,6 +295,11 @@ class RemoteCatalog(Catalog):
             def items(self):
                 for item in six.iteritems(self._page_cache):
                     yield item
+                if not server_can_paginate():
+                    # We are talking to an older server, before pagination
+                    # support was added. It would have already sent us all its
+                    # entries, so we are done.
+                    return
                 # Fetch more entries from the server.
                 while True:
                     page = fetch_page(self._highest_page_fetched + 1)
@@ -324,6 +332,7 @@ class RemoteCatalog(Catalog):
                         return source
 
         self._entries = Entries()
+        self._server_can_paginate = True
         super(RemoteCatalog, self).__init__(self, **kwargs)
 
     def _get_http_args(self):
@@ -362,10 +371,13 @@ class RemoteCatalog(Catalog):
         # If the server respects the pagination parameters, info['sources']
         # should be empty. But if we have an old server, it will contain all
         # the entries and we should cache them now.
-        self._entries._page_cache.update({source['name']: RemoteCatalogEntry(
-                url=self.source_url,
-                getenv=self.getenv,
-                getshell=self.getshell,
-                auth=self.auth,
-                http_args=self.http_args, **source)
-                for source in info['sources']})
+        if info['sources']:
+            self._server_can_paginate = False
+            self._entries._page_cache.update(
+                {source['name']: RemoteCatalogEntry(
+                    url=self.source_url,
+                    getenv=self.getenv,
+                    getshell=self.getshell,
+                    auth=self.auth,
+                    http_args=self.http_args, **source)
+                 for source in info['sources']})
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -251,7 +251,7 @@ class RemoteCatalog(Catalog):
             # and our own more direct context.
             try:
                 response.raise_for_status()
-            except HTTPError:
+            except requests.HTTPError:
                 raise RemoteCatalogError(
                     "Failed to fetch page of entries.")
             info = msgpack.unpackb(response.content, encoding='utf-8')
@@ -273,7 +273,7 @@ class RemoteCatalog(Catalog):
                 raise KeyError(name)
             try:
                 response.raise_for_status()
-            except HTTPError:
+            except requests.HTTPError:
                 raise RemoteCatalogError(
                     "Failed to fetch entry {!r}.".format(name))
             info = msgpack.unpackb(response.content, encoding='utf-8')
@@ -409,7 +409,7 @@ class RemoteCatalog(Catalog):
                                 **self._get_http_args())
         try:
             response.raise_for_status()
-        except HTTPError:
+        except requests.HTTPError:
             raise RemoteCatalogError(
                 "Failed to fetch metadata {!r}.".format(name))
         info = msgpack.unpackb(response.content, encoding='utf-8')
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -10,7 +10,7 @@ from requests.compat import urljoin, urlparse
 from ..auth.base import BaseClientAuth
 from .entry import CatalogEntry
 from .remote import RemoteCatalogEntry
-from .utils import flatten, reload_on_change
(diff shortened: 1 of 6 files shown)


martindurant:
I will look at this in the afternoon. A more capable server is certainly a project goal, and pagination/info a great start for this.

At an immediate hunch, I feel like a discussion on attribute versus item lookup is a separate conversation.

danielballan:
OK, agreed. I will rip out the item-lookup change and the associated test changes in b1dabc4 after your review lands.

martindurant:
xref #198 - changes like this should (eventually) be well documented, giving the specific set of end-points and their behaviour on a dedicated page.

intake/catalog/base.py
```diff
@@ -237,17 +374,35 @@ def _load(self):
...
-            auth=self.auth,
-            http_args=self.http_args, **s)
-            for s in info['sources']}
+        self._entries.reset()
+        # If the server respects the pagination parameters, info['sources']
+        # should be empty. But if we have an old server, it will contain all
+        # the entries and we should cache them now.
+        if info['sources']:
+            self._server_can_paginate = False
+            self._entries._page_cache.update(
```
martindurant: why not just set the dictionary as before in this case?

intake/catalog/tests/test_local.py
```diff
@@ -75,7 +75,7 @@ def test_source_plugin_config(catalog1):
 
 def test_metadata(catalog1):
     assert hasattr(catalog1, 'metadata')
-    assert catalog1['metadata']['test'] is True
+    assert catalog1.metadata['test'] is True
```
martindurant: I feel like the change may break things for too many people. I note that for a *local* catalog, both methods still work. Also, it is inkeeping with the idea that `in` normally relates to what you can do with getitem. This is inspired by the usage of both in Pandas.

intake/cli/server/__main__.py
```diff
@@ -27,6 +27,8 @@ def main(argv=None):
     parser = argparse.ArgumentParser(description='Intake Catalog Server')
     parser.add_argument('-p', '--port', type=int, default=conf['port'],
                         help='port number for server to listen on')
+    parser.add_argument('--list-entries', action='store_true',
```
martindurant: Good idea!

Also, flatten yes/no may be a good one.

intake/cli/server/server.py
```diff
@@ -84,13 +85,25 @@ def initialize(self, cache, catalog, auth):
 
     def get(self):
         head = self.request.headers
+        page_size = self.get_argument('page[size]', None)
```
martindurant: was wondering why you use `page[size]` as opposed to `page_size` or the object `page: {size:, number/offset"}`

intake/cli/server/server.py
```diff
@@ -84,13 +85,25 @@ def initialize(self, cache, catalog, auth):
...
+            if page_size is None:
+                # Return all the results in one page. This is important for
+                # back-compat with clients that predate pagination. It may
+                # also be useful to keep things simple for clients that do not
+                # need pagination.
+                start = stop = None
+            else:
+                start = (int(page_number) - 1) * int(page_size)
+                stop = int(page_number) * int(page_size)
+            page = itertools.islice(cat.walk(depth=1).items(), start, stop)
```
martindurant: same as `iter(cat)` here instead of `walk`?

intake/cli/server/server.py
```diff
@@ -160,6 +173,31 @@ def initialize(self, catalog, cache, auth):
         self._cache = cache
         self.auth = auth
 
+    def get(self):
+        head = self.request.headers
```
martindurant: Doc: this is for direct access to an *entry* by name for random access where the whole catalog has not first been listed (e.g., in the case of pagination).
Correct?

intake/cli/server/server.py
```diff
@@ -160,6 +173,31 @@ def initialize(self, catalog, cache, auth):
         self._cache = cache
         self.auth = auth
 
+    def get(self):
```
martindurant: There is an argument that this belongs under the info/ endpoint:
- this location: because it's about a specific source
- in info: because we are only getting information

intake/catalog/base.py
```diff
@@ -73,7 +75,11 @@ def __init__(self, *args, **kwargs):
             args = args[0]
         self.args = args
         self.updated = time.time()
-        self._entries = {}
+        # Allow the subclass to have set self._entries to some other dict-like
+        # object *before* this is run so that the last thing to happen is
+        # force_reload().
+        if getattr(self, '_entries', None) is None:
```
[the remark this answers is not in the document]
danielballan: This is necessary because of the way `__getitem__` and `__getattr__` mix. See for example:

```
In [49]: type(catalog)                                                                                                 
Out[49]: intake.catalog.base.RemoteCatalog

In [50]: hasattr(catalog, '_doesnotexit')                                                                              
Out[50]: True
```

I would definitely prefer to change this, but that's out of scope for this PR.

intake/catalog/base.py
```diff
@@ -169,16 +175,20 @@ def __getitem__(self, key):
         
         Can also use attribute syntax, like ``cat.entry_name``.
         """
-        if key in self._entries:
+        try:
             return self._entries[key]
```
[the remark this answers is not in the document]
danielballan: Good thought. I also like that implementing `__contains__` helps the _user_ avoid accidentally pulling down the entire catalog if they use `... in catalog` in user code.

intake/cli/server/__main__.py
danielballan: Thanks. I haven't quite wrapped my brain around how that would go so I am punting this to [a new issue](https://github.com/ContinuumIO/intake/issues/202). Feel free to consider it "assigned" to me.

intake/catalog/base.py
```diff
@@ -222,10 +235,134 @@ def __init__(self, url, http_args={}, **kwargs):
             self.source_url = url
             self.info_url = url.replace('v1/source', 'v1/info')
         self.auth = kwargs.get('auth', None)  # instance of BaseClientAuth
+
+        def fetch_page(page_number):
+            logger.debug("Request page %d of entries", page_number)
+            params = {'page[number]': page_number,
```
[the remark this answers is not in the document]
danielballan: I agree that would be more flexible. Will change.

intake/catalog/base.py
```diff
@@ -237,17 +374,35 @@ def _load(self):
         if self._source_id is not None:
             headers['source_id'] = self._source_id
         http_args['headers'] = headers
-        response = requests.get(self.info_url, **http_args)
+        return http_args
+
+    def _load(self):
+        """Fetch metadata from remote. Entries are fetched lazily."""
+        # This will not immediately fetch any sources (entries). It will lazily
+        # fetch sources from the server in paginated blocks when this Catalog
```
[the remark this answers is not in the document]
danielballan: Yes. I believe this is the only way to keep iteration order, which seemed to me worth keeping.

intake/cli/server/server.py
danielballan: I use the key (`name`) and the value (`source`) below. Iterating just gives me the keys. Is there a better choice than `walk`?

intake/cli/server/server.py
danielballan: Yeah, I actually put it in `/info` first and then moved it here. Feels like a tossup to me -- happy to follow your judgement here.

intake/catalog/base.py
danielballan: Updating the existing cache feels safer than blowing away the whole object and replacing it. But I can appreciate the simplicity of that approach, and could be persuaded.

intake/cli/server/server.py
danielballan: Yes, correct. Docstring added. Thank you for flagging this.

intake/catalog/base.py
```diff
@@ -222,10 +235,134 @@ def __init__(self, url, http_args={}, **kwargs):
...
+
+        def fetch_by_name(name):
+            logger.debug("Requesting info about entry named '%s'", name)
+            params = {'name': name}
+            response = requests.get(self.source_url, params=params,
+                                    **self._get_http_args())
+            if response.status_code == 404:
+                raise KeyError(name)
+            if response.status_code != 200:
+                raise Exception('%s: status code %d' % (response.url,
```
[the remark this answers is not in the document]
danielballan: My favorite pattern here is to `raise_for_status()` and then catch that and raise a more context-specific error, which produces a nice chained exception in Python 3 and something still useful in Python 2. Done.

danielballan:
I force-pushed to blow away the one commit that had affected `__getitem__` -> `__getattr__` and had altered some existing tests. I added some new commits to address review comments and left some follow-up questions. Tomorrow I plan to address:

- [ ] moving locally-defined functions and classes out into module
- [ ] changing pagination from number/size to offset/limit formulation

intake/catalog/base.py
```diff
@@ -222,10 +235,134 @@ def __init__(self, url, http_args={}, **kwargs):
...
+                http_args=self.http_args, **info['source'])
+
+        def server_can_paginate():
+            # a closure for use in entries, because we won't know the truth of
+            # this until after _load() is called
+            return self._server_can_paginate
+
+        page_size = self._page_size
+
+        class Entries:
```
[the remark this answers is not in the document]
danielballan: Ah, re-reading your comment --

> Might any other sort of catalog want to implement the paging idea, e.g., if reading from another data service?

-- I now see that these may be well worth reusing. A strong case for avoiding the rather eccentric choice I made of hiding them in a local definition!

intake/catalog/base.py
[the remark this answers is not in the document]
danielballan: This unexpected behavior is still an issue, but I did settle on a more elegant way to let subclasses customize `self._entries` in 9de555f7086a98f8ca3830ee82f07cdf0981f95a.

intake/catalog/base.py
```diff
@@ -188,11 +206,86 @@ def _close(self):
...
+        "Clear caches to force a reload."
+        self._page_cache.clear()
+        self._direct_lookup_cache.clear()
+        self._page_offset = 0
+
+    def __iter__(self):
+        for key in self.keys():
+            yield key
+
+    def __contains__(self, key):
```
martindurant: Does this mean that the `__contains__` for the Catalog class can just call `if in` on its entries dict object?

intake/catalog/base.py
```diff
@@ -188,11 +206,86 @@ def _close(self):
...
+            yield item
+        if self._catalog.page_size is None:
+            # We are not paginating, either because the user set page_size=None
+            # or the server is a version of intake before pagination parameters
+            # were added.
+            return
+        # Fetch more entries from the server.
+        while True:
+            page = self._catalog.fetch_page(self._page_offset)
+            self._page_cache.update(page)
```
martindurant: Do we ever want to purge from the head of the cache?

intake/catalog/base.py
```diff
@@ -224,8 +321,66 @@ def __init__(self, url, http_args={}, **kwargs):
...
+    def page_size(self):
+        return self._page_size
+
+    def fetch_page(self, page_offset):
+        logger.debug("Request page entries %d-%d",
+                     page_offset, page_offset + self._page_size)
+        params = {'page_offset': page_offset,
+                  'page_size': self._page_size}
+        response = requests.get(self.info_url, params=params,
+                                **self._get_http_args())
```
martindurant: Maybe check keys of `_get_http_args()`, and merge `params` key, if it exists.

intake/cli/server/server.py
```diff
@@ -160,6 +173,49 @@ def initialize(self, catalog, cache, auth):
...
+        """
+        head = self.request.headers
+        name = self.get_argument('name')
+        if self.auth.allow_connect(head):
+            if 'source_id' in head:
+                cat = self._cache.get(head['source_id'])
+            else:
+                cat = self._catalog
+            try:
+                # This would ideally be cat[name] and not access the interal
```
martindurant: OK, so this goes away following the getitem discussion :)

intake/catalog/base.py
danielballan: Good question, @martindurant. Let me think this through. The purpose of `_page_cache` is to expedite subsequent calls to `items()`. Since those _always_ start from the top -- there is no API to start in the middle -- it makes sense to hold on to the early pages no matter what.

```
iterator1 = cat.items()
# Iterate through some items.
iterator2 = cat.items()
# If have dumped the first page, we'll have to pull it again.
```

We could impose a limit on the number of pages and/entries to cache, as in a new parameter `page_cache_size` alongside `page_size` to make it possible to iterate through unbounded catalogs. I think that's a good idea. But I don't think purging from the head is ever optimal, which is fortunate from the point of view of simplicity.

Does that seem like a good approach to you?

Tangentially related:
* I have a `search` branch based off of this branch that supports search on remote catalogs (and nested catalogs therein) which will also be useful for navigating through large catalogs, addressing that concern about "there is no API to start in the middle" mentioned above. 
* As you likely already noticed, the _entire_ cache is occasionally purged, of course, whenever the Catalog reloads.

danielballan:
Rebased on master and left follow-up comment. Looks like 2ec29c6 broke a reload test, not yet sure why....

martindurant:
Passes before 2ec29c6 but fails after?

danielballan:
Correct. Very repeatable locally. The version before 2ec29c6 calls `__getitem__` and triggers [this logic](https://github.com/ContinuumIO/intake/pull/200/files#diff-e28c1f323132aa5d6916dd9d8b2e0326R180) which might be doing something subtle.

danielballan:
I suppose we want to keep it the long way because `Catalog.__getitem__` is checking nested-ness (dots in names), which `Entries` is not, and if that comes back with a result we want `__contains__` to return `True`. OK with you if I rebase 2ec29c6 out?

danielballan:
:+1: Only remaining question is the one in line above then. Thanks.

intake/catalog/base.py
martindurant: Happy to leave it. Perhaps at some point someone will want a purely iterative approach, or partitioned entry list to process with dask... but no need for now.

danielballan:
Fire away! :-D

PR #204, opened 2018-12-03
Add flatten option to server (default True)

### LINKED ISSUE #202: Add a 'flatten' argument to the server CLI
A good suggestion lifted from [this review comment](https://github.com/ContinuumIO/intake/pull/200#discussion_r237260506).

martindurant:
Fixes #202

### DIFF
--- a/intake/cli/server/__main__.py
+++ b/intake/cli/server/__main__.py
@@ -19,7 +19,7 @@ def call_exit_on_sigterm(signal, frame):
 
 def main(argv=None):
     from intake.config import conf
-    from intake import Catalog
+    from intake import open_catalog
 
     if argv is None:
         argv = sys.argv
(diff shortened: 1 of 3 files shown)


danielballan:
Out of curiosity, why flatten by default (or at all)?

martindurant:
Not sure, I didn't write the default, would be happy to change it.

danielballan:
In my case, flattening would create many name collisions, so `False` seems a safer default from my perspective, but I'll be content as long as there is a switch available.

martindurant:
Note that flattening only refers to passing multiple files (or glob) to the intake-server command - is that your situation, @danielballan ?

danielballan:
Oh I see. No. Carry on, sorry for the noise.

PR #206, opened 2018-12-05
Make GUI instance only get created on demand

martindurant:
Fixes https://github.com/ContinuumIO/intake/issues/183#issuecomment-439730993

@LightnerAndrew , sorry for the delay - does this work for you?

### DIFF
--- a/intake/gui/__init__.py
+++ b/intake/gui/__init__.py
@@ -8,3 +8,32 @@ except ImportError:
         def __repr__(self):
             raise RuntimeError("Please install ipywidgets to use the Data "
                                "Browser")
+except Exception as e:
+
+    class DataBrowser(object):
+        def __repr__(self):
+            raise RuntimeError("Initialisation of GUI failed, even though"
+                               "ipywidgets is installed. Please update it"
+                               "to a more recent version.")
+
+
+class InstanceMaker(object):
+
+    def __init__(self, cls, *args, **kwargs):
+        self._cls = cls
+        self._args = args
+        self.kwargs = kwargs
+        self._instance = None
+
+    def _instantiate(self):
+        if self._instance is None:
+            print("instantiate")
+            self._instance = self._cls(*self._args, **self.kwargs)
+
+    def __getattr__(self, attr, *args, **kwargs):
+        self._instantiate()
+        return getattr(self._instance, attr, *args, **kwargs)
+
+    def __getitem__(self, item):
+        self._instantiate()
+        return self._instance[item]
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -12,7 +12,7 @@ from .catalog import local
 from .catalog.default import load_combo_catalog
 from .source import registry
 from .source.discovery import autodiscover
-from .gui import DataBrowser
+from .gui import DataBrowser, InstanceMaker
 
 # Populate list of autodetected plugins
 registry.update(autodiscover())
@@ -136,4 +136,4 @@ def open_catalog(uri=None, **kwargs):
 
 
 Catalog = open_catalog
-gui = DataBrowser()
+gui = InstanceMaker(DataBrowser)
--- a/intake/cli/server/tests/test_server.py
+++ b/intake/cli/server/tests/test_server.py
@@ -219,7 +219,6 @@ class TestServerV1Source(TestServerV1Base):
             self.server._cache.peek(source_id)
 
 
-
 @pytest.fixture()
 def multi_server(tmpdir):
     fn1 = os.path.join(tmpdir, 'cat1.yaml')


martindurant:
cc @jbednar for a review, if you have time.

jsignell:
Does this get rid of the dependency on ipywidgets on import?

jsignell:
I think the version string is wrong in requirements which might be why people are getting tripped up in the first place - or not getting the error sooner. Possibly there should be an intake[core] and an intake[gui] env or something?

martindurant:
Still attempts to import ipywidgets, but doesn't instantiate anything. Should give decent message in place of GUI for most situations where people try to use it but it fails. Hopefully the situations like @LightnerAndrew 's where the GUI caused a crash, people can simply not use it. Also, I am imagining this as the first step of having the GUI backend picked by a config key.

Could implement no importing right now, I suppose.

martindurant:
@bryevdv , any opinion on how best to allow for multiple GUI backends here?

jsignell:
screen shot 2018-12-05 at 10 30 21 am

martindurant:
That would be OK, I suppose - sorry for the debug text there.

jsignell:
The only option for tab complete is giving me is `kwargs`?

martindurant:
Tab complete doesn't make much sense for something whose only purpose is to be displayed as a GUI; but I could implement some of the basic methods like dir and str to make it behave a little nicer. First I want to make sure this is a reasonable approach.

jsignell:
:+1: Now when I do `intake.gui` I get a good and helpful error message - albeit missing a space:

```python-traceback
RuntimeError                              Traceback (most recent call last)
~/conda/envs/intake-dev/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

~/conda/envs/intake-dev/lib/python3.6/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    400                         if cls is not object \
    401                                 and callable(cls.__dict__.get('__repr__')):
--> 402                             return _repr_pprint(obj, self, cycle)
    403 
    404             return _default_pprint(obj, self, cycle)

~/conda/envs/intake-dev/lib/python3.6/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    695     """A pprint that just redirects to the normal repr function."""
    696     # Find newlines and replace them with p.break_()
--> 697     output = repr(obj)
    698     for idx,output_line in enumerate(output.splitlines()):
    699         if idx:

~/intake/intake/gui/__init__.py in __repr__(self)
     40     def __repr__(self):
     41         self._instantiate()
---> 42         return repr(self._instance)
     43 
     44     def __dir__(self):

~/intake/intake/gui/__init__.py in __repr__(self)
     13     class DataBrowser(object):
     14         def __repr__(self):
---> 15             raise RuntimeError("Initialisation of GUI failed, even though"
     16                                "ipywidgets is installed. Please update it"
     17                                "to a more recent version.")

RuntimeError: Initialisation of GUI failed, even thoughipywidgets is installed. Please update itto a more recent version.
```

intake/gui/__init__.py
```diff
@@ -8,3 +8,39 @@ class DataBrowser(object):
         def __repr__(self):
             raise RuntimeError("Please install ipywidgets to use the Data "
                                "Browser")
+except Exception as e:
+
+    class DataBrowser(object):
+        def __repr__(self):
+            raise RuntimeError("Initialisation of GUI failed, even though"
+                               "ipywidgets is installed. Please update it"
```
jsignell: need trailing spaces

bryevdv:
> any opinion on how best to allow for multiple GUI backends here?

I'm not sure I understand the question in the context of this PR. As I interpret it, having multiple GUI backends seems like a large maintenance burden with no payoff. I'd try to find one GUI backend to use, and go with that. Offhand, Panel/Bokeh seems like a good choice to explore in the future, since it can work both in and outside Jupyter. 

All that is orthogonal to making GUI support optional, which seems more to be the issue for this PR?

bryevdv:
BTW being new to this codebase I only just now appreciated that the issue discussed here is a result of allowing an attribute `intake.gui` rather than requiring a function call e.g. `intake.gui()`. In general I would simply advocate for the latter, i.e. making users call a function. However in cases where the former is required, I'd point to the ["class module trick" described by Guido](https://mail.python.org/pipermail/python-ideas/2012-May/014969.html) as a way to avoid having to define the extra `InstanceGetter` class. Bokeh uses this method for the [`bokeh.palettes` module](https://github.com/bokeh/bokeh/blob/master/bokeh/palettes.py#L1382-L1387)

Also it occurs to me that perhaps you were asking about multiple GUI backends re: multiple versions of ipywidgets? I think that is far more hassle than its worth, and would simply communicate your ipywidgets version requirements (whatever they are) clearly.

PR #207, opened 2018-12-05
Bryanv/docs tidy

bryevdv:
@martindurant I was going to first make a PR for plotting docs, but I'd actually like to discuss some of the goals there first. In any case as I was just going through all the docs and code structure to get familiar I thought I would fix up a few things as I went along. 

### Add copyright blocks to all source files

I think you will want ultimately want these from an organizational hygiene POV, and speaking from experience with Bokeh, you will be grateful to do this sort of thing earlier, when the project is smaller. However, I am happy to remove this commit if it is not desired. 

### Sphinx warnings

Most caused by duplicate targets and can be avoided with anonymous links. There is one remaining warning, but I think the best solution there would be to actually document the `open_X` functions individually and explicitly, but that's would be up to you to decide. 

### Minor Edits

Small corrections for some typos and adjustments to grammatical structure.

### DIFF
--- a/docs/source/data-packages.rst
+++ b/docs/source/data-packages.rst
@@ -4,7 +4,7 @@ Making Data Packages
 Combined with the `Conda Package Manger <https://conda.io/docs/>`_, Intake makes it possible to create *data packages*
 which can be installed and upgraded just like software packages.  This offers several advantages:
 
-  * Distributing datasets becomes as easy as installing software
+  * Distributing datasets becomes as easy ``conda install``
   * Data packages can be versioned, improving reproducibility in some cases
   * Data packages can depend on the libraries required for reading
   * Data packages can be self-describing using Intake catalog files
@@ -20,10 +20,13 @@ The steps involved in creating a data package are:
 
 1. Identifying a dataset, which can be accessed via a URL or included directly as one or more files in the package.
 
-2. Creating a package containing an intake catalog file, meta.yaml (description of the data, version, requirements,
-   etc.) and a script to copy the data.
+2. Creating a package containing:
 
-3. Building the package using the command conda build.
+   * an intake catalog file
+   * a ``meta.yaml`` file (description of the data, version, requirements, etc.)
+   * a script to copy the data
+
+3. Building the package using the command ``conda build``.
 
 4. Uploading the package to a package repository such as `Anaconda Cloud <https://anaconda.org>`_ or your own private
    repository.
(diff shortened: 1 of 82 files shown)


docs/source/data-packages.rst
```diff
@@ -4,7 +4,7 @@ Making Data Packages
 Combined with the `Conda Package Manger <https://conda.io/docs/>`_, Intake makes it possible to create *data packages*
 which can be installed and upgraded just like software packages.  This offers several advantages:
 
-  * Distributing datasets becomes as easy as installing software
+  * Distributing datasets becomes as easy ``conda install``
```
bryevdv: I chuckled when I read the original -- if someone says a task is "as easy as installing software" I'd assume it's very difficult :)

docs/source/examples.rst
```diff
@@ -43,22 +43,22 @@ More specific examples of Intake functionality
...
-      [`Static <https://github.com/jsignell/intake-blog/blob/master/path-as-pattern/landsat.ipynb>`_]
-      [`Executable <https://mybinder.org/v2/gh/jsignell/intake-blog/master?filepath=path-as-pattern%2Flandsat.ipynb>`_]
+      [`Static <https://github.com/jsignell/intake-blog/blob/master/path-as-pattern/landsat.ipynb>`__]
+      [`Executable <https://mybinder.org/v2/gh/jsignell/intake-blog/master?filepath=path-as-pattern%2Flandsat.ipynb>`__]
 
     - How to set up pattern parsing
-      [`Static <https://github.com/jsignell/intake-blog/blob/master/path-as-pattern/csv.ipynb>`_]
-      [`Executable <https://mybinder.org/v2/gh/jsignell/intake-blog/master?filepath=path-as-pattern%2Fcsv.ipynb>`_]
+      [`Static <https://github.com/jsignell/intake-blog/blob/master/path-as-pattern/csv.ipynb>`__]
+      [`Executable <https://mybinder.org/v2/gh/jsignell/intake-blog/master?filepath=path-as-pattern%2Fcsv.ipynb>`__]
```
bryevdv: double trailing underscores create anonymous links (no targets, hence no duplicate targets)

martindurant:
This is all excellent, thank you.
A conflict seems to have come in, presumably from what I was doing, sorry.

bryevdv:
It looks like `import collections` was added in master but I don't actually see `collections` used anywhere?

martindurant:
If tests pass without it, feel free to remove.

bryevdv:
FYI I get one failing test when I tried to run locally, including trying on master. I just ran `py.test` at the top level. Is there any documentation on dev setup/running tests by hand?

martindurant:
> Is there any documentation on dev setup/running tests by hand

That's probably a good idea, and no not yet.
My versions of things are moderately up to date; the set of installs used by Travis (same as the conda requirements) would be a good start. To test the same as Travis, you can run `conda build`, which I don't normally do in the interests of time.

martindurant:
Seems like it was required after all :)
```
   def __init__(self, catalog):
        self._catalog = catalog
>       self._page_cache = collections.OrderedDict()
E       NameError: name 'collections' is not defined
```

PR #208, opened 2018-12-05
Support search() on remote catalogs

danielballan:
Submitted for a first-pass review. Needs tests. There is some work in progress on making a nested catalog...I'm not exactly sure how to formulate that in YAML and wouldn't mind a nudge in the right direction.

This supports:

1. "Progressive" search: search on a Catalog that is itself a set of search results.
2. Searching on a Catalog that is nested inside another Catalog

See an example illustrating both [at this section of a notebook](https://nbviewer.jupyter.org/github/danielballan/intake-bluesky/blob/master/demo.ipynb#Progressive-Search). The Catalog in that example is backed by MongoDB, and so its `search()` method expects a dict (to be precise, a dict that is a valid Mongo Query) but in general this PR accommodates any single argument to `search()`, including the `text` search that is implemented in intake's own Catalogs.

This PR implements progressive search by caching the chain of queries accumulated over chained searched and iteratively applying them each time. This could be made more efficient: one could imagine adding an optional `optimized_multi_search()` method that expects a _list_ of queries and does internal optimization. I'd be happy to include that in this PR, but I'll hold off until we agree that the basic outline is good.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -381,7 +381,8 @@ class RemoteCatalog(Catalog):
         logger.debug("Request page entries %d-%d",
                      page_offset, page_offset + self._page_size)
         params = {'page_offset': page_offset,
-                  'page_size': self._page_size}
+                  'page_size': self._page_size,
+                  'query': json.dumps(self._query)}
         http_args = self._get_http_args(params)
         response = requests.get(self.info_url, **http_args)
         # Produce a chained exception with both the underlying HTTPError
@@ -406,7 +407,7 @@ class RemoteCatalog(Catalog):
 
     def fetch_by_name(self, name):
         logger.debug("Requesting info about entry named '%s'", name)
-        params = {'name': name}
+        params = {'name': name, 'query': json.dumps(self._query)}
         http_args = self._get_http_args(params)
         response = requests.get(self.source_url, **http_args)
         if response.status_code == 404:
@@ -459,7 +460,7 @@ class RemoteCatalog(Catalog):
 
         if self.page_size is None:
             # Fetch all source info.
-            params = {}
+            params = {'query': json.dumps(self._query)}
         else:
             # Just fetch the metadata now; fetch source info later in pages.
             params = {'page_offset': 0, 'page_size': 0}
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -5,7 +5,6 @@
 # The full license is in the LICENSE file, distributed with this software.
 #-----------------------------------------------------------------------------
 
-import base64
 import collections
 import copy
 import logging
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -353,6 +353,8 @@ class RemoteCatalog(Catalog):
         if not url.endswith('/'):
             url = url + '/'
         self.url = url
+        self.info_url = urljoin(url, 'v1/info')
+        self.source_url = urljoin(url, 'v1/source')
         self.http_args = http_args
         self.http_args.update(kwargs.get('storage_options', {}))
         self.http_args['headers'] = self.http_args.get('headers', {})
@@ -361,14 +363,10 @@ class RemoteCatalog(Catalog):
         self._query = tuple(query or ())
         self._source_id = kwargs.get('source_id', None)
         if self._source_id is None:
-            self.info_url = urljoin(url, 'v1/info')
-            self.source_url = urljoin(url, 'v1/source')
             self.name = urlparse(url).netloc.replace(
                 '.', '_').replace(':', '_')
         else:
             self.name = kwargs['name']
-            self.info_url = urljoin(url, 'v1/info')
-            self.source_url = urljoin(url, 'v1/source')
         self.auth = kwargs.get('auth', None)  # instance of BaseClientAuth
         super(RemoteCatalog, self).__init__(self, **kwargs)
 
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -357,7 +357,7 @@ class RemoteCatalog(Catalog):
         self.http_args['headers'] = self.http_args.get('headers', {})
         self._page_size = page_size
         # Convert to (immutable) tuple just to avoid accidental mutation.
-        self._query = tuple(query or (, ))
+        self._query = tuple(query or ())
         self._source_id = kwargs.get('source_id', None)
         if self._source_id is None:
             self.info_url = urljoin(url, 'v1/info')
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -346,7 +346,12 @@ class RemoteCatalog(Catalog):
             parameters to pass to remote backend file-system. Ignored for
             normal local files.
         """
-        self._url = url
+        secure = http_args.pop('ssl', False)
+        scheme = 'https' if secure else 'http'
+        url = url.replace('intake', scheme)
+        if not url.endswith('/'):
+            url = url + '/'
+        self.url = url
         self.http_args = http_args
         self.http_args.update(kwargs.get('storage_options', {}))
         self.http_args['headers'] = self.http_args.get('headers', {})
@@ -356,17 +361,14 @@ class RemoteCatalog(Catalog):
         self._query = tuple(query)
         self._source_id = kwargs.get('source_id', None)
         if self._source_id is None:
-            secure = http_args.pop('ssl', False)
-            scheme = 'https' if secure else 'http'
-            base_url = url.replace('intake', scheme) + '/'
-            self.info_url = urljoin(base_url, 'v1/info')
-            self.source_url = urljoin(base_url, 'v1/source')
-            self.name = urlparse(base_url).netloc.replace(
+            self.info_url = urljoin(url, 'v1/info')
+            self.source_url = urljoin(url, 'v1/source')
+            self.name = urlparse(url).netloc.replace(
                 '.', '_').replace(':', '_')
         else:
             self.name = kwargs['name']
-            self.source_url = url
-            self.info_url = url.replace('v1/source', 'v1/info')
+            self.info_url = urljoin(url, 'v1/info')
+            self.source_url = urljoin(url, 'v1/source')
         self.auth = kwargs.get('auth', None)  # instance of BaseClientAuth
         super(RemoteCatalog, self).__init__(self, **kwargs)
 
@@ -395,7 +397,7 @@ class RemoteCatalog(Catalog):
                 "".format(page_offset, page_offset + self._page_size))
         info = msgpack.unpackb(response.content, **unpack_kwargs)
         page = {source['name']: RemoteCatalogEntry(
-            url=self.source_url,
+            url=self.url,
             getenv=self.getenv,
             getshell=self.getshell,
             auth=self.auth,
@@ -419,7 +421,7 @@ class RemoteCatalog(Catalog):
                 "Failed to fetch entry {!r}.".format(name))
         info = msgpack.unpackb(response.content, **unpack_kwargs)
         return RemoteCatalogEntry(
-            url=self.source_url,
+            url=self.url,
             getenv=self.getenv,
             getshell=self.getshell,
             auth=self.auth,
@@ -484,7 +486,7 @@ class RemoteCatalog(Catalog):
             self._page_size = None
             self._entries._page_cache.update(
                 {source['name']: RemoteCatalogEntry(
-                    url=self.source_url,
+                    url=self.url,
                     getenv=self.getenv,
                     getshell=self.getshell,
                     auth=self.auth,
@@ -494,6 +496,6 @@ class RemoteCatalog(Catalog):
     def search(self, query):
         # Return a new RemoteCatalog like this one in all respects except with
         # this query appended to its list of queries.
-        return RemoteCatalog(url=self._url,
+        return RemoteCatalog(url=self.url,
                              http_args=self.http_args,
                              query=self._query + (query,))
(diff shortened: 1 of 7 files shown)


intake/cli/sample/nested_catalog_example.yml
```diff
@@ -0,0 +1,2 @@
+sources:
+    - us_states
```
danielballan: WIP

intake/catalog/base.py
```diff
@@ -298,26 +302,36 @@ def __init__(self, url, http_args={}, page_size=None, **kwargs):
...
             The number of entries fetched at a time during iteration.
             Default is None (no pagination; fetch all entries in bulk).
+        query : Iterable or None
+            List from a sequence of progressive searches. Each element is
+            expected to be a valid argument to Catalog.search() for this
+            Catalog.
         kwargs: may include catalog name, metadata, source ID (if known) and
             auth instance.
         """
+        secure = http_args.pop('ssl', False)
```
danielballan: See commit message on https://github.com/ContinuumIO/intake/pull/208/commits/5a66628bead4eaa7910db3f7ed5dad5ee10af84a for description of why this change is scope for this PR.

martindurant:
On a separate note, the "None" as the name of search-output catalogs is a little poor. Very list it should say "search results".

intake/cli/server/server.py
```diff
@@ -102,15 +104,25 @@ def get(self):
             else:
                 start = int(page_offset)
                 stop = int(page_offset) + int(page_size)
-            page = itertools.islice(cat.walk(depth=1).items(), start, stop)
+            if query_list:
+                # This could be a search of a search of a serach (etc.).
+                # Progressively apply each search and then page through the
+                # final results.
+                for query in query_list:
+                    cat = cat.search(query)
```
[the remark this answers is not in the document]
danielballan: This is a good idea. Let me see if I can sort out how that would work.

intake/catalog/base.py
```diff
@@ -434,9 +449,16 @@ def _load(self):
             self._page_size = None
             self._entries._page_cache.update(
                 {source['name']: RemoteCatalogEntry(
-                    url=self.source_url,
+                    url=self.url,
```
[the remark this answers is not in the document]
danielballan: All of the `__init__`s now expect the base URL instead of a URL that has `/v1/source` or `/v1/info` in it. It makes composition simpler if we always expect the base URL.

martindurant:
Sorry for the conflict, quite a few PRs landed. I am trying to tie things down before I leave on holiday.

danielballan:
No worries!

danielballan:
I will pick this back up after the holiday break. Thanks again for the review.

intake/catalog/base.py
```diff
@@ -298,26 +302,36 @@ def __init__(self, url, http_args={}, page_size=None, **kwargs):
         page_size : int, optional
             The number of entries fetched at a time during iteration.
             Default is None (no pagination; fetch all entries in bulk).
+        query : Iterable or None
```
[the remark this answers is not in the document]
danielballan: Yes, client-side search is also valuable. Client-side searching doesn't need to stash and pass through state of "the queries that got us to this point" because it can just materialize the entire resultant (local) Catalog. On a RemoteCatalog, I assume that we need the server (in my application's case, MongoDB) to parse the parameters to `search(...)`, so we have to aggregate them as state carried along with the `RemoteCatalog`.

martindurant:
```
E     File "/home/travis/build/ContinuumIO/intake/intake/catalog/base.py", line 363
E       self._query = tuple(query or (, ))
```

`(,)` -> `()` :)

martindurant:
@danielballan , is there more to do here, or is this ready for review?

martindurant:
At least in py3, str(query) will give a consistent output, e.g., https://github.com/martindurant/filesystem_spec/blob/master/fsspec/utils.py#L215

The main point of the ID, though, is that the client can continue to refer to the same search results on the server without re-querying. I'm not too worried (but perhaps you are) about different clients sharing the results of the same query.

Adding more to the API sounds a bit overkill here.

danielballan:
> At least in py3, str(query) will give a consistent output. [...] Adding more to the API sounds a bit overkill here.

Agreed that adding special API for this is not desirable and agreed that `str` is sufficient for the cases we know about so far as long as dicts are ordered (3.5+). That's good enough for my use case, so I'll go with that.

> I'm not too worried (but perhaps you are) about different clients sharing the results of the same query.

I don't think that will be a concern for me either.

danielballan:
This is now in decent shape, ready for another round of review. Thanks for your help untangling my thoughts about tokenization.

intake/catalog/base.py
```diff
@@ -298,26 +302,36 @@ def __init__(self, url, http_args={}, page_size=None, **kwargs):
...
+            List from a sequence of progressive searches. Each element is
+            expected to be a valid argument to Catalog.search() for this
```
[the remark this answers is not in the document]
danielballan: In order to accommodate multiple args/kwargs (rather than just one arg, `query`, in my initial PR) `query` now expects an even _more_ persnickety structure. If you prefer, we could remove the `query` arg from `__init__` and add a new classmethod, `RemoteCatalog.from_query`, to be called by `RemoteCatalog.serach()` and perhaps by sophisticated users.

intake/catalog/tests/test_remote_integration.py
```diff
@@ -290,3 +290,18 @@ def test_getitem_and_getattr(intake_server):
     assert catalog.arr is catalog['arr']
     assert isinstance(catalog.arr, RemoteCatalogEntry)
     assert isinstance(catalog.metadata, (dict, type(None)))
+
+
+def test_search(intake_server):
+    catalog = Catalog(intake_server)
+
+    results = catalog.search('text', depth=1)
```
martindurant: I may have missed it: is this doing the normal walk/text search, or is everything passed to the server now?

intake/cli/server/server.py
```diff
@@ -86,17 +87,21 @@ class ServerInfoHandler(tornado.web.RequestHandler):
...
 
     def get(self):
         head = self.request.headers
         page_size = self.get_argument('page_size', None)
         page_offset = self.get_argument('page_offset', 0)
+        query_list = json.loads(self.get_argument('query', '[]'))
         if self.auth.allow_connect(head):
             if 'source_id' in head:
                 cat = self.cache.get(head['source_id'])
+                query_key = (head['source_id'],)
```
martindurant: Don't we just operate on `cat`? (or is this only for making the cat have a nice name)

intake/cli/server/server.py
```diff
@@ -107,15 +112,33 @@ def get(self):
             else:
                 start = int(page_offset)
                 stop = int(page_offset) + int(page_size)
-            page = itertools.islice(cat.walk(depth=1).items(), start, stop)
+            if query_list:
+                # This could be a search of a search of a serach (etc.).
+                # Progressively apply each search and then page through the
+                # final results.
+                for args, kwargs in query_list:
```
martindurant: I wonder if, for the mongo case, there is something to be done to merge the queries/dicts rather than chaining them in this loop. Forgive me if we already talked about the possibility

intake/cli/server/server.py
```diff
@@ -107,15 +112,33 @@ def get(self):
...
+                    # The results of each search are cached.
+                    query_key += (str((args, kwargs)),)
+                    try:
+                        source_id = self.searches[query_key]
```
martindurant: Really, we want an LRU cache?

intake/catalog/base.py
```diff
@@ -450,7 +465,7 @@ def _load(self):
 
         if self.page_size is None:
             # Fetch all source info.
-            params = {}
+            params = {'query': json.dumps(self._query)}
```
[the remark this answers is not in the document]
danielballan: Fair point. Fixed.

intake/catalog/base.py
```diff
@@ -474,9 +489,25 @@ def _load(self):
...
+    def search(self, *args, **kwargs):
+        # Return a new RemoteCatalog like this one in all respects except with
+        # this query appended to its list of queries.
+        return RemoteCatalog(
+            url=self.url,
+            http_args=self.http_args,
+            query=self._query + ((args, kwargs),),
+            # Give a name like <This Catalog>.search(...) with args and kwargs
+            # filled in.
+            name=("{!r}.search(".format(self) +
```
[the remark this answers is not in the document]
danielballan: I like that this leaves open the possibility of an eval-able repr, but I'll punt that to a separate PR and set name to `""` for now.

intake/cli/server/server.py
danielballan: Yeah, in the PR description (a long time ago now haha!) I pitched:

> This PR implements progressive search by caching the chain of queries accumulated over chained searched and iteratively applying them each time. This could be made more efficient: one could imagine adding an optional `optimized_multi_search()` method that expects a list of queries and does internal optimization. I'd be happy to include that in this PR, but I'll hold off until we agree that the basic outline is good.

In the Mongo case, that would be:

```py
def optimized_compound_search(self, *queries):
    return self.serach({'$and': queries})
```

I think it should be a separate method so that we don't have to alter the signature of `search(...)` to accommodate it, which I think would compromise the usability of `search(...)`. I am imagining that the server would check for this method, use it if it's implemented and fall back on the iterative approach if it is not. What do you think?

intake/catalog/base.py
```diff
@@ -397,7 +412,7 @@ def fetch_page(self, page_offset):
 
     def fetch_by_name(self, name):
         logger.debug("Requesting info about entry named '%s'", name)
-        params = {'name': name}
+        params = {'name': name, 'query': json.dumps(self._query)}
```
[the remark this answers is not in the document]
danielballan: I don't recall what my reasoning was originally. Seems better to stick with msgpack; I'll make the change.

intake/catalog/base.py
```diff
@@ -206,14 +208,17 @@ def __getitem__(self, key):
...
         out = self
-        if '.' in key:
-            for k in key.split('.'):
-                if isinstance(out, CatalogEntry):
-                    out = out()  # default parameters
-                if not isinstance(out, Catalog):
-                    raise ValueError("Attempt to recurse into non-catalog")
-                out = out[k]
-            return out
+        if isinstance(key, str):
```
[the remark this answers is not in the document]
danielballan: As noted in #229, I am in favor of mergin #229 first and rebasing this onto that.

intake/catalog/tests/test_remote_integration.py
danielballan: Because `catalog` is a `RemoteCatalog`, everything is passed to the server now. Previously -- that is, on current `master`, `RemoteCatalog` does not implement the `search` method at all. The `search` behavior of a local Catalog is unchanged.

intake/cli/server/server.py
danielballan: I'm not sure I understand the question. This is for creating a unique (internal) key to cache search results to be generated below. The key needs to capture both the search query itself and some identification of which (sub)catalog the search was performed on. This line does that second part.

intake/catalog/base.py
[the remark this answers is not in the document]
danielballan: Now I recall my reasoning -- will expand in comment outside of line notes....

intake/cli/server/server.py
martindurant: I don't think another route or method is a good thing, I agree with you. Consolidation of query terms may be done by the can on the server if it supports it, and the client should know nothing about it.

Still I have something I think is a misunderstanding - maybe we can talk it through live.

Let cat be a local reference to a bluesky catalog on a server, backed by mongo.

The user performs `cat2 = cat.search(...)`. The arguments are passed to the server, it executes the search and stores a new cat instance in the cache, returning an ID to the client. cat2 now has one entry in its query list, and a remote ID to pull entries from.

The user performs `cat3 = cat2.search(...)`. The arguments passed to the server include both the ID of the cat in the server cache and two elements of query, the previous one and the one added in this step. On the server side, which of the following occurs:
- use ID to fetch cat2 from the cache, and do cat2.search(query_list[-1]); noting that cat2 remembers the query that created it
- do cat.search(query_list[0]).search(query_list[1])
- do cat.search(query list), which is the equivalent of the previous option, but with the possibility of an optimised query in the future.

intake/cli/server/server.py
danielballan: Better yet, I think we can reuse the existing cache expiration mechanism and simplify this a bit. Done in 2bafa8ad1ecb7c1762226f2e923c150e4fe06596.

intake/cli/server/server.py
danielballan: The concern I have with that approach is that it sometimes passes `search(one_query)` and other times passes `search(query_list)`, which means `search` has to be smart enough to distinguish between them, and in general it might not be clear. Moreover, the "query" might not just one item; it could involve the use of kwargs like `depth`. That is the problem I am aiming to solve with a separate, dev-oriented method that explicitly takes a list of `(args, kwargs)` pairs.

intake/catalog/base.py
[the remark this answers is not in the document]
danielballan: Not sure I understand the question. Let's remember to clarify this one on the call.

danielballan:
This is ready for another round of review. Key points:
* Search queries are submitted through a new action in the existing POST route.
* No custom (JSON, base64) serialization -- just the usuage msgpack stuff.
* No "query list" necessary, and no query-caching on the client side.
* The discussion around `optimized_compound_search` is now not relevant/motivated as hoped.
* I fixed an unrelated bug (with test) in 6862f69 that I happened to hit while writing search tests.

martindurant:
I think everything here is ready to go?
I would appreciate additions to docs/source/server.rst , but that can come in a separate PR, if you prefer.

danielballan:
Yes this is ready to go. I pushed a commit to update the server documentation; thanks for the reminder.

martindurant:
Excellent, thank you.

PR #210, opened 2018-12-06
Fix getitem

### LINKED ISSUE #203: For discussion: getattr -> getitem is good, but the inverse is confusing
Intake goes further than pandas in blurring the line between attribute access and item access. I would like explore the possibility of moving to behavior more like pandas.

To review pandas behavior:

```py
In [7]: import pandas                                                                                                                                         

In [8]: df = pandas.DataFrame({'a': [1, 2]})                                                                                                                  

In [9]: df['a']                                                                                                                                               
Out[9]: 
0    1
1    2
Name: a, dtype: int64

In [10]: df.a  # a convenient alias                                                            
Out[10]: 
0    1
1    2
Name: a, dtype: int64

In [11]: df.sum  # This is part of the DataFrame API, not a column.                                                                                           
Out[11]: 
<bound method DataFrame.sum of    a
0  1
1  2>

In [12]: df['sum']  # This is a KeyError.
<snip>
KeyError: 'sum'
```

So, to summarize, `__getattr__` falls back to `__getitem__`, but not the other way around. Intake maps in both directions:

```py
In [1]: import intake                                                                                                                                         

In [2]: cat = intake.open_catalog('intake/cli/sample/us_states.yml')                                                                                          

In [3]: cat['states']                                                                                                                                         
Out[3]: <Catalog Entry: states>

In [4]: cat.states  # a convenient alias, a la pandas                                                                                                         
Out[4]: <Catalog Entry: states>

In [5]: cat.metadata  # this part of the Category API, not an entry                                                                                           
Out[5]: {}

In [6]: cat['metadata']  # I would not expect this to work, but it does.                                                                                      
Out[6]: {}
```

Anecdotally, it took me a day or two with intake to recognize that this was happening, and in my week or two working with the library it's made some debugging more subtle.

danielballan:
Closes #203 

Please take a close look at the tests to verify that we agree on what the desired behavior is. Thanks!

P.S. @martindurant I also figured out the problem with `__contains__` that was stumping me in #200: we just need `_get_entries()` instead of `_entries` in order to trigger a reload. Now `__contains__` is simplified.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -187,7 +187,12 @@ class Catalog(DataSource):
 
     def __getattr__(self, item):
         if not item.startswith('_'):
-            return self._get_entry(item)
+            # Fall back to __getitem__.
+            try:
+                return self._get_entry(item)
+            except KeyError:
+                raise AttributeError(item)
+        raise AttributeError(item)
 
     def __getitem__(self, key):
         """Return a catalog entry by name.
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -111,6 +111,7 @@ class Catalog(DataSource):
         # default version for pre-v1 files
         return self.metadata.get('version', 1)
 
+    @reload_on_change
     def search(self, text, depth=2):
         words = text.lower().split()
         cat = Catalog(name=self.name + "_search",
@@ -172,12 +173,7 @@ class Catalog(DataSource):
 
     def __contains__(self, key):
         # Avoid iterating through all entries.
-        try:
-            self[key]
-        except KeyError:
-            return False
-        else:
-            return True
+        return key in self._get_entries()  # triggers reload_on_change
 
     def __dir__(self):
         return list(self)
@@ -189,7 +185,7 @@ class Catalog(DataSource):
         if not item.startswith('_'):
             # Fall back to __getitem__.
             try:
-                return self._get_entry(item)
+                return self._get_entry(item)  # triggers reload_on_change
             except KeyError:
                 raise AttributeError(item)
         raise AttributeError(item)
@@ -199,16 +195,19 @@ class Catalog(DataSource):
 
         Can also use attribute syntax, like ``cat.entry_name``.
         """
-        if key in self._entries:
+        if key in self._get_entries():  # triggers reload_on_change
             return self._entries[key]
         out = self
-        for k in key.split('.'):
-            if isinstance(out, CatalogEntry):
-                out = out()  # default parameters
-            if not isinstance(out, Catalog):
-                raise ValueError("Attempt to recurse into non-catalog")
-            out = getattr(out, k)
-        return out
+        if '.' in key:
+            for k in key.split('.'):
+                if isinstance(out, CatalogEntry):
+                    out = out()  # default parameters
+                if not isinstance(out, Catalog):
+                    raise ValueError("Attempt to recurse into non-catalog")
+                out = out[k]
+            return out
+        else:
+            raise KeyError(key)
 
     def discover(self):
         return {"container": 'catalog', 'shape': None,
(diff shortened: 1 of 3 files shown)


intake/catalog/base.py
```diff
@@ -187,23 +183,31 @@ def __repr__(self):
 
     def __getattr__(self, item):
         if not item.startswith('_'):
-            return self._get_entry(item)
+            # Fall back to __getitem__.
+            try:
+                return self._get_entry(item)  # triggers reload_on_change
```
martindurant: So it is actually the same as `self[item]`, given that an attribute can't contain "."?

intake/catalog/tests/test_local.py
```diff
@@ -82,7 +82,7 @@ def test_source_plugin_config(catalog1):
 
 def test_metadata(catalog1):
     assert hasattr(catalog1, 'metadata')
-    assert catalog1['metadata']['test'] is True
```
martindurant: Ah, so the point was that "metadata" was always supposed to be an attribute only...

intake/catalog/tests/test_local.py
```diff
@@ -431,3 +431,24 @@ def test_no_plugins():
     s = cat.tables7
     with pytest.raises(ValueError):
         s()
+
+
+def test_getitem_and_getattr():
+    from intake.source.csv import CSVSource
+    fn = abspath('multi_plugins.yaml')
```
martindurant: Should use `os.join(dirpath(__file__), 'multi_plugins.yaml')` ?

intake/catalog/tests/test_local.py
```diff
@@ -431,3 +431,24 @@ def test_no_plugins():
...
+    catalog = open_catalog(fn)
+    catalog['tables0']
+    with pytest.raises(KeyError):
+        catalog['doesnotexist']
+    with pytest.raises(KeyError):
+        catalog['_doesnotexist']
+    with pytest.raises(KeyError):
+        # This exists as an *attribute* but not as an item.
+        catalog['metadata']
+    catalog.tables0  # alias to catalog['tables0']
```
martindurant:
```
assert catalog.tables0 is catalog['tables0']
assert isinstance(catalog.tables0, LocalCatalogEntry)
assert isinstance(catalog.metadata, (dict, NoneType))
```

intake/catalog/tests/test_local.py
```diff
@@ -431,3 +431,24 @@ def test_no_plugins():
...
+    catalog.metadata  # a normal attribute
+    with pytest.raises(AttributeError):
+        catalog.doesnotexit
+    with pytest.raises(AttributeError):
+        catalog._doesnotexit
+    catalog.metadata
```
martindurant: unneeded repeat

intake/catalog/tests/test_remote_integration.py
```diff
@@ -268,3 +268,22 @@ def test_pagination(intake_server):
...
+    catalog = Catalog(intake_server)
+    catalog['arr']
+    with pytest.raises(KeyError):
+        catalog['doesnotexist']
+    with pytest.raises(KeyError):
+        catalog['_doesnotexist']
+    with pytest.raises(KeyError):
+        # This exists as an *attribute* but not as an item.
+        catalog['metadata']
+    catalog.arr  # alias to catalog['arr']
```
martindurant: as above

intake/catalog/tests/test_local.py
danielballan: Sure. I blindly copied this from another test. Will update all occurrences.

danielballan: Oh actually, that's what `abspath` does. (It's locally defined convenience function.)

intake/catalog/base.py
danielballan: Good point, will simplify.

danielballan:
Ready for another round of review.

martindurant:
All good, thank you!
This is definitely an improvement.

danielballan:
Sure thing, thanks for being receptive to this change.

PR #211, opened 2018-12-06
Reduce build data duplication

bryevdv:
@martindurant This PR does this following (it may be simpler to look at each commit individually, they are all standalone):

* Alphabetizes the contents of `requirements.txt`

  Slightly helpful to have

* Removes `pytest` from `requirements.txt`

  Not sure why this should be here, but let me know if this commit should be reverted

* pulls "about" data fields from `setup.py`

  This is accomplished by using `{% set data = load_setup_py_data() %}` with the templating facility conda-build provides for `meta.yaml` files

* pulls entry point definitions from `setup.py`

  Same mechanism as "about" fields

* use `requirements.txt` for runtime deps everywhere

  This will need your judgment. After some experiences with botched/out of sync releases for Bokeh, I fall heavily on the side of "not repeating configuration". However, the situation for intake is slightly complicated relative to Bokeh, due to:

  * the `ruamal.yaml` vs `ruamel_yaml` mess. It's special cased in the `meta.yaml` in a somewhat ugly way. However, if you will be moving to using conda-forge, or if the main repo ever changes to agree with pypi and conda-forge naming, then this special case could be removed.

  * `dask[complete]` changed to `dask >=0.17.0` Would need to verify that other dependencies cover the things you actually need, or add them if not. I assume npy and csv readers are covered by `numpy` and `pandas` packages but this should be confirmed.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,3 +1,5 @@
+{% set data = load_setup_py_data() %}
+
 package:
    name: intake
    version: {{ environ.get('GIT_DESCRIBE_TAG', 'unknown') }}
@@ -42,7 +44,7 @@ test:
     - py.test --pyargs -v intake
 
 about:
-  home: https://github.com/ContinuumIO/intake
-  license: BSD
+  home: {{ data['url'] }}
+  license: {{ data['license'] }}
   license_file: LICENSE
-  summary: Data ingest catalog system
+  summary: {{ data['description'] }}
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -23,16 +23,16 @@ requirements:
   run:
     - python
     - numpy
-    - pandas
-    - msgpack-python
-    - msgpack-numpy
-    - requests
-    - tornado
-    - jinja2
+
+    {% for dep in data['install_requires'] %}
+    {% if "ruamel.yaml" not in dep %}
+    - {{ dep.lower() }}
+    {% endif %}
+    {% endfor %}
+
+    # There is an unfortunate name difference for this package in the main
+    # conda repo. This can be removed when/if ruamel.yaml becomes available
     - ruamel_yaml >=0.15.0
-    - dask >=0.17.0
-    - python-snappy
-    - appdirs
 
 test:
   requires:
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -10,8 +10,9 @@ source:
 build:
   number: {{ environ.get('GIT_DESCRIBE_NUMBER', 0) }}
   entry_points:
-    - intake-server = intake.cli.server.__main__:main
-    - intake = intake.cli.client.__main__:main
+    {% for entry in data['entry_points']['console_scripts'] %}
+    - {{ entry }}
+    {% endfor %}
   script: python setup.py install --single-version-externally-managed --record=record.txt
   noarch: python
 
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -2,7 +2,7 @@
 
 package:
    name: intake
-   version: {{ environ.get('GIT_DESCRIBE_TAG', 'unknown') }}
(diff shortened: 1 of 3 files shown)


bryevdv:
Additionally, the `meta.yaml` package version could also pull from `setup.py` via `data['version']`. That would result in conda-build using the versioneer version at all times. That's what Bokeh does, but was not sure if it's something you'd want. If you think that's advisable I will add it.

martindurant:
> not sure if it's something you'd want

It is good hygiene, I agree with you.

martindurant:
I should ask you to make these changes to all of the repos I'm involved in... :)
In Intake-land alone, there are quite a few.

bryevdv:
> I should ask you to make these changes to all of the repos I'm involved in... :)

I do find a peace of mind from tending to details like this. 

> It is good hygiene, I agree with you.

I've added a new commit to do this. FYI local builds now build packages like

> /Users/bryanv/anaconda/conda-bld/noarch/intake-0.3.0+46.g086889d.dirty-py_46.tar.bz2

but will report the tag version, when there is a tag (via versioneer) so should yield identical results when releasing, assuming releases happen exactly on tags.

requirements.txt
```diff
@@ -1,15 +1,15 @@
+# This file is also processed by conda-build, so keep spaces as "pkg >=1.2"
+appdirs
+dask >=0.17.0
+holoviews
+ipywidgets >=7.2
```
martindurant: Following the discussions, we can drop this requirement.

requirements.txt
```diff
@@ -1,15 +1,15 @@
...
-ruamel.yaml >= 0.15.0
 requests
-appdirs
+ruamel.yaml >=0.15.0
 six
-tornado >= 4.5.1
-dask[complete]
-holoviews
-ipywidgets >= 7.2
+tornado >=4.5.1
```
martindurant: I suppose this is strictly optional too - the server won't work without it, but everything else will. Let's defer for now, it would require at least a good message on failure in the server main.

requirements.txt
```diff
@@ -1,15 +1,15 @@
...
+appdirs
+dask >=0.17.0
+holoviews
+ipywidgets >=7.2
 jinja2
 msgpack-numpy
 msgpack-python
 numpy
 pandas
-pytest
```
martindurant: I suppose there could be a test-requires.txt, but this is already covered in meta.yaml.

conda/meta.yaml
```diff
@@ -1,15 +1,18 @@
+{% set data = load_setup_py_data() %}
```
martindurant: Never knew about it, very handy.

requirements.txt
bryevdv: Yes at least for Bokeh we only really support test/development from conda-based envs. If you want to be more broad than that, a `test-requires.txt` is an option, or else it could just be made a runtime dependency all the time.

conda/meta.yaml
bryevdv: I know about it (and it exists) thanks to @msarahan

requirements.txt
martindurant: Let's leave for now. Documentation on dev/test should be written at some point.

requirements.txt
bryevdv: done in https://github.com/ContinuumIO/intake/pull/211/commits/54fe34544672e8fb61e839ecdbfd73177a78c4b3

requirements.txt
bryevdv: Just to make sure I understand, leave things as they currently are, without `pytest` ?

martindurant: remove pytest

bryevdv: removed in https://github.com/ContinuumIO/intake/pull/211/commits/6fcc5a8dc225d7d7cc6de4a11ed2798476e84f1b

PR #212, opened 2018-12-06
Streamline Travis config

bryevdv:
@martindurant another PR for you to consider. This moves the actual code that Travis runs out from `.travis.yml` to real bash scripts. This allows them to be edited with proper syntax highlighting an indentation, and also as the project grows you may want to add build stages or a more complex matrix. This keeps things manageable. 

I also copied over the "Emergency Pin" code block from Bokeh. This can be used to quickly get tests working again in cases where some new version of a test dependency completely breaks something overnight. I think this is slightly less likely here, as you currently have far fewer test dependencies, but if you ever do need it, at least it is there ready to use.

### DIFF
--- a/scripts/ci/build.sh
+++ b/scripts/ci/build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -e # exit on error
+
+source ${HOME}/miniconda3/bin/activate root
+
+# Workaround for Travis-CI bug #2: https://github.com/travis-ci/travis-ci/issues/7773
+if ! [ "$TRAVIS_OS_NAME" = "linux" ]; then
+    EXTRA_OPTS="--no-test"
+fi
+
+echo "Building conda package."
+conda build -c defaults -c conda-forge ${EXTRA_OPTS:-} ./conda
+
+# If tagged, upload package to main channel
+if [ -n "$TRAVIS_TAG" ]; then
+    echo "Uploading conda package."
+    anaconda -t ${ANACONDA_TOKEN} upload -u intake --force `conda build --output ./conda`
+fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,26 @@
 language: generic
+
 git:
   depth: false
+
 os:
   - linux
+
 branches:
   only:
   - master
+
 env:
   matrix:
   - CONDA_PY=36
   global:
+    # ANACONDA_TOKEN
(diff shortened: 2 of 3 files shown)


scripts/ci/build.sh
```diff
@@ -0,0 +1,22 @@
...
+    EXTRA_OPTS="--no-test"
+else
+
+echo "Building conda package."
+conda build -c defaults -c conda-forge ${EXTRA_OPTS:-} ./conda
+
+# If tagged, upload package to main channel, otherwise if not a PR, upload to dev
+if [ -n "$TRAVIS_TAG" ]; then
+    LABEL_OPT=""
+elif ! [ -n "$TRAVIS_PULL_REQUEST" ]; then
```
martindurant: We should take this opportunity to stop pushing dev packages at all, i.e., the upload only happens for the tag branch

bryevdv:
As an aside, you might want to consider specifying versions for `CONDA_REQS` precisely. It's no fun having a release deploy fail at the last minute because one of those changed somehow. Hopefully less of a risk that it used to be, but still good to be safe.

martindurant:
> consider specifying versions for CONDA_REQS precisely

I go back and forth on this. We *do* want to know that things broke because a dependency was updated, but also want to know whether it was something in our PR code which caused the failure or not. I'm not sure there's a good solution to this - perhaps two travis runs, one with pinned versions and one without? How, then, do you decide when to update the pinning?

scripts/ci/build.sh
bryevdv: done in https://github.com/ContinuumIO/intake/pull/212/commits/9b151d3da9b7f88b3a7d140b7041ff029a55a109 (also fixed logic and syntax errors)

bryevdv:
> I go back and forth on this. We do want to know that things broke because a dependency was updated

Just to be clear, I am only talking about `conda-build`, etc. I agree about user dependencies. But `conda-build` I just want to work

martindurant:
> conda-build I just want to work

OK, that's fair

martindurant:
Do we need to make a new token? We could remove build from Travis altogether and have it as an offline process.

bryevdv:
> Do we need to make a new token? 

I removed the `set -x` before it ever ran, so I don't think so. I prefer automating the upload but it's definitely your preference for Intake.

martindurant:
OK, if the token is safe, leave at as is for now.

bryevdv:
In case you are curious for the future how I obscure the token for Bokeh, I run upload by running a subprocess from a real python script and control what is echo'ed:

https://github.com/bokeh/bokeh/blob/master/scripts/build_upload.py#L358-L371

bryevdv:
@martindurant The commit history is a bit messy now, do you have policy/preference re: rebases and force pushes?

martindurant:
> do you have policy/preference re: rebases and force pushes

I don't mind

bryevdv:
@martindurant this is ready for final review. 

Obviously as with anything that touches final release codepaths, there won't be a 100% test until a release is issued. But I think the code here is straightforward enough that things are likely to work as expected. (Famous last words!)

Also, I forgot earlier, but Travis should automatically try to censor any secret env var values that are leaked. We had to ask them to turn it off for Bokeh due to some weird interaction with our complicated logs, but AFAIK it's always on by default nowadays.

martindurant:
OK, let's put it in and deal with any release issues as they arise.

PR #215, opened 2018-12-07
Multi cat behaviour

### LINKED ISSUE #214: Improve opening multiple catalogs
- I think it's worth clearing up that when you open multiple catalog yaml files into one catalog, the catalog entries are combined into a new catalog, which is a union of those individual catalogs.

- When multiple catalog yamls are opened with `open_catalog`, the `__repr__` of the returned catalog object is no different from an empty catalog
```
<Intake catalog: None>
```
whereas if you open just one catalog yaml file `mycatalog.yaml`, it shows
```
<Intake catalog: mycatalog>
```
I understand this is because the new combined catalog now has no `name` and if I give explicitly `name='mycat'` keyword, it will show that. Still, I think it's an issue that empty catalog is not distinguished from non-empty catalog. I think one possible way is to indicate the number of catalog entries available: for non-empty it would be non-zero.

- If there are multiple sources with the same name, this raises no warning/exception and whatever is read later silently overwrites.

In fact, it seems much more natural to me to have a collection of catalogs objects (that's actually what I expected originally from opening multiple catalog yaml files). To be able to do something like
```
d = open_catalog(<multiple catalogs>)
a = d.catalog1.source1.read()
b = d.catalog2.source1.read()
```
If I open multiple catalog yamls with the same metadata field (e.g., description), what description would I get in the resulting combined catalog?

martindurant:
Fixes #214

### DIFF
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -105,7 +105,16 @@ def open_catalog(uri=None, **kwargs):
         Designator for the location of the catalog.
     kwargs:
         passed to subclass instance, see documentation of the individual
-        catalog classes.
+        catalog classes. For example, ``yaml_files_cat`` (when specifying
+        multiple uris or a glob string) takes the additional
+        parameter ``flatten=True|False``, specifying whether all data sources
+        are merged in a single namespace, or each file becomes
+        a sub-catalog.
+
+    See also
+    --------
+    intake.open_yaml_files_cat, intake.open_yaml_file_cat,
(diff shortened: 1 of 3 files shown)


martindurant:
@smoh for your review

martindurant:
(will merge anyway if I don't hear from you, as I will be on holiday for a while)

smoh:
Looks good to me and please do.

PR #226, opened 2019-01-10
Misc CLI work

bryevdv:
@martindurant In thinking about an eventual `intake browser` command, I had a peek at the CLI code. It was missing some small things I think are useful/common so I made a quick PR to add them:

* enable `-h` for default help:

  screen shot 2019-01-10 at 10 40 10

* add `-v` for reporting version:

  screen shot 2019-01-10 at 10 40 56

* add `intake info` command to report runtime env information, similar to `bokeh info`. This is very useful as a dead-simple ask to users to run to report intake-specific dependencies and configuration. 

  screen shot 2019-01-10 at 10 42 37

  I realize there its the `intake config list-defaults` already but that seems to serve a different purpose. Or perhaps they could be merged. 

---

I wanted to ask about more work before proceeding. I do think from a software design/maintenance  perspective it is nice to modularize subcommand handling more than they are currently in this project. I did that with conda (and it is still in use today) and also for bokeh. I'd be happy to make a pass at `intake` and `intake-server` (I rather enjoy this sort of cleanup). Let me know what you think.

### DIFF
--- a/intake/cli/client/__main__.py
+++ b/intake/cli/client/__main__.py
@@ -165,7 +165,7 @@ def main(argv=None):
     if argv is None:
         argv = sys.argv
 
-    parser = argparse.ArgumentParser(description='Intake Catalog CLI', add_help=False)
+    parser = argparse.ArgumentParser(description='Intake Catalog CLI')
     subparsers = parser.add_subparsers(help='sub-command help')
 
     parser.add_argument('-v', '--version', action='version', version=__version__)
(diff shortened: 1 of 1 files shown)


martindurant:
Yes, certainly that kind of refactoring is much appreciated, and this should also be reflected in the doc pages. 
Do you intend to ass more here, or were you just thinking of future extensions?

martindurant:
I suspect the number of things returned there will grow with time.

PR #229, opened 2019-01-14
Allow cat['item1', 'subitem2']

martindurant:
(also cleaned up some test warnings)

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -19,6 +19,7 @@ from .entry import CatalogEntry
 from .remote import RemoteCatalogEntry
 from .utils import flatten, reload_on_change, RemoteCatalogError
 from ..source.base import DataSource
+from ..compat import unpack_kwargs
 logger = logging.getLogger('intake')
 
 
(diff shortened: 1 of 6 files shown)


danielballan:
I'll play Devil's advocate on this one. This seems less clear to me than simply doing:

```py
cat['item1']['subitem2']
```
and it's no less verbose (trading `']['` for `', '`).

Reading this, I think it might give the impression that it returns a Catalog of multiple items, akin to the DataFrame multi-column syntax `df[['A', 'B']]`. The difference between the two is only subtle, and in numpy-world comma-separated series inside square bracket usually indicated "and also" not "nested inside", right?

martindurant:
`cat['item1']['subitem2']`

is also supported by this change. Before if would have needed to be `cat['item1]()[`subitem2']`.

To me, x[a, b] does mean "subselect on a, then subselect the results with b" and this is what happens with arrays. Using a *list* in the selector, as Pandas does, is something else again. 

I should have clarified the reason for this PR: at the moment one can access nested catalogues down to an individual data source as

    cat.subcat.subsubcat.data

(returns an Entry, which can be called to make a source, or any access to it's attributes like `to_dask()` automatically makes a source).

However, catalogues not made with python in mind will have name that aren't allowed, so quoting them with the [..] syntax seems like a good escape. We could, alternatively, coerce names to something valid, or provide some other shortcut to descend cat hierarchies.

danielballan:
Sure, it's clear that this doesn't break the existing usage, I was just leaning back a bit against on "too many ways to do the same thing". The fact that this avoids the `()` had not occurred to me.

> To me, x[a, b] does mean "subselect on a, then subselect the results with b"

I see. The analogy of subcatalog-as-axis wasn't immediately clear to me. I now appreciate how it is reasonable to read this similarly to "Column X in Row Y".

PR #234, opened 2019-01-17
Windows support

jsignell:
Adds window support and appveyor to make sure it doesn't break again. 

Some things I'm worried about not being tested properly:
 - does serialization work properly on windows?
 - should CATALOG_DIR have a trailing /?
 - does catalog flattening work properly on windows?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,5 +20,5 @@ install:
   - python setup.py install
   - conda list
 
-script:
+test_script:
   - pytest intake --verbose
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,8 @@ install:
   # and the user has the necessary permissions to make environment changes. In a CI
   # environment these are not necessary and slow things down noticeably on Windows.
   # also adding conda-forge as highest-priority default channel
-  - conda config --set always_yes yes --set changeps1 no --set auto_update_conda no --set safety_checks disabled --add channels conda-forge
+  - conda config --set always_yes yes --set changeps1 no --set auto_update_conda no --set safety_checks disabled
+  - conda config --add channels conda-forge
   - conda update -q conda
   - conda env create -n test_env python=3.6 pytest
   - conda activate test_env
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,23 @@
+# Inspired by appveyor in xarray and anaconda-project
+
+build: false
+
+environment:
+  global:
+    CONDA: "C:\\Miniconda36-x64"
+
+install:
+  - "SET PATH=%CONDA%;%CONDA%\\Scripts;%PATH%"
+  # The safety checks are simply intended to ensure that there is enough disk space
+  # and the user has the necessary permissions to make environment changes. In a CI
+  # environment these are not necessary and slow things down noticeably on Windows.
+  - conda config --set always_yes yes --set changeps1 no --set auto_update_conda no --set safety_checks disabled --add channels conda-forge
+  - conda update -q conda
+  - conda env create -n test_env python=3.6 pytest
+  - conda activate test_env
+  - conda install --file requirements.txt
+  - python setup.py install
+  - conda list
+
+script:
+  - pytest intake --verbose
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,25 +1,26 @@
-# Inspired by appveyor in xarray and anaconda-project
-
+# Based on bokeh appveyor set up
 build: false
 
+platform:
+  - x64
+
 environment:
-  global:
-    CONDA: "C:\\Miniconda36-x64"
+  matrix:
+    - MINICONDA: C:\Miniconda36-x64
+
+skip_branch_with_pr: true
+clone_depth: 5
+skip_tags: true
+
+init:
+  - cmd: set PATH=%MINICONDA%;%MINICONDA%\\Scripts;%MINICONDA%\\Library\\bin;%PATH%
+  - cmd: echo %path%
 
 install:
-  - "SET PATH=%CONDA%;%CONDA%\\Scripts;%PATH%"
-  # The safety checks are simply intended to ensure that there is enough disk space
-  # and the user has the necessary permissions to make environment changes. In a CI
-  # environment these are not necessary and slow things down noticeably on Windows.
-  # also adding conda-forge as highest-priority default channel
-  - conda config --set always_yes yes --set changeps1 no --set auto_update_conda no --set safety_checks disabled
-  - conda config --add channels conda-forge
-  - conda update -q conda
-  - conda env create -n test_env python=3.6 pytest
-  - conda activate test_env
-  - conda install --file requirements.txt
-  - python setup.py install
-  - conda list
+  - powershell .\\scripts\\ci\\appveyor\\install.ps1
+
+build_script:
+  - powershell .\\scripts\\ci\\appveyor\\build.ps1
 
 test_script:
-  - pytest intake --verbose
+  - py.test
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,7 @@ install:
   # The safety checks are simply intended to ensure that there is enough disk space
   # and the user has the necessary permissions to make environment changes. In a CI
   # environment these are not necessary and slow things down noticeably on Windows.
+  # also adding conda-forge as highest-priority default channel
   - conda config --set always_yes yes --set changeps1 no --set auto_update_conda no --set safety_checks disabled --add channels conda-forge
   - conda update -q conda
   - conda env create -n test_env python=3.6 pytest
--- a/intake/catalog/tests/test_local.py
+++ b/intake/catalog/tests/test_local.py
@@ -480,7 +480,6 @@ def test_no_plugins():
 
 
 def test_getitem_and_getattr():
-    from intake.source.csv import CSVSource
     fn = abspath('multi_plugins.yaml')
     catalog = open_catalog(fn)
     catalog['tables0']
--- a/intake/catalog/tests/test_local.py
+++ b/intake/catalog/tests/test_local.py
@@ -7,7 +7,6 @@
 
 import datetime
 import os.path
-import posixpath
 import shutil
 import tempfile
 import time
@@ -104,16 +103,16 @@ def test_get_dir():
     path = 'example/catalog.yml'
     out = get_dir(path)
     assert os.path.isabs(out)
-    assert out.endswith('/example')
+    assert out.endswith('/example/')
     path = '/example/catalog.yml'
     out = get_dir(path)
     # it's ok if the first two chars indicate drive for win (C:)
-    assert posixpath.dirname(path) in [out, out[2:]]
+    assert '/example/' in [out, out[2:]]
     path = 'example'
     out = get_dir(path)
     assert os.path.isabs(out)
     assert not out.endswith('/example')
-    assert not out.endswith('/')
+    assert out.endswith('/')
 
 
 @pytest.mark.parametrize("dtype,expected", [
--- a/intake/catalog/tests/test_local.py
+++ b/intake/catalog/tests/test_local.py
@@ -7,6 +7,7 @@
 
 import datetime
 import os.path
+import posixpath
 import shutil
 import tempfile
 import time
@@ -19,10 +20,10 @@ from .util import assert_items_equal
 from intake import Catalog, open_catalog
 from intake.catalog import exceptions, local
 from intake.catalog.local import get_dir, UserParameter, LocalCatalogEntry
-
+from intake.utils import make_path_posix
 
 def abspath(filename):
-    return os.path.join(os.path.dirname(__file__), filename)
+    return make_path_posix(os.path.abspath(filename))
 
 
 @pytest.fixture
@@ -99,11 +100,18 @@ def test_get_dir():
     assert get_dir('s3://path/catalog.yml') == 's3://path'
     assert get_dir('https://example.com/catalog.yml') == 'https://example.com'
     path = 'example/catalog.yml'
-    assert get_dir(path) == os.path.join(os.getcwd(), os.path.dirname(path))
+    out = get_dir(path)
+    assert os.path.isabs(out)
+    assert out.endswith('/example')
     path = '/example/catalog.yml'
-    assert get_dir(path) == os.path.dirname(path)
+    out = get_dir(path)
+    # it's ok if the first two chars indicate drive for win (C:)
+    assert posixpath.dirname(path) in [out, out[2:]]
     path = 'example'
-    assert get_dir(path) == os.path.join(os.getcwd(), '')
+    out = get_dir(path)
+    assert os.path.isabs(out)
+    assert not out.endswith('/example')
+    assert not out.endswith('/')
 
 
 @pytest.mark.parametrize("dtype,expected", [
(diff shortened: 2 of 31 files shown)


martindurant:
Do you need to activate appveyor first, or does this need to go in first?
I am unqualified to know if this sets things up OK, happy to merge as-is.

appveyor.yml
```diff
@@ -0,0 +1,23 @@
...
+install:
+  - "SET PATH=%CONDA%;%CONDA%\\Scripts;%PATH%"
+  # The safety checks are simply intended to ensure that there is enough disk space
+  # and the user has the necessary permissions to make environment changes. In a CI
+  # environment these are not necessary and slow things down noticeably on Windows.
+  - conda config --set always_yes yes --set changeps1 no --set auto_update_conda no --set safety_checks disabled --add channels conda-forge
+  - conda update -q conda
+  - conda env create -n test_env python=3.6 pytest
+  - conda activate test_env
+  - conda install --file requirements.txt
```
martindurant: normally requirements.txt should be the pip names of the packages, and conda names may differ. There may also be additional, optional packages that we want for the purpose of testing.
cc @bryevdv

jsignell: Think I can just take all that directly? I wasn't sure that we need to build intake to just run the tests

bryevdv: Probably so? It's the window analogue of this test install step https://github.com/ContinuumIO/intake/blob/master/scripts/ci/install.sh#L26-L27

jsignell: ok I switched it

jsignell:
Woohoo! We now have the same output as reported in https://github.com/ContinuumIO/intake/issues/233

martindurant:
You want to use this PR to do those fixes too?

jsignell:
I was planning on using this PR yeah. Unless you'd prefer to keep things separate.

martindurant:
I agree with you - then we can merge only when the big button is green.

danielballan:
Oh, I’m happy to see this as it was next on my list. Thanks, @jsignell!

jsignell:
Still a ways to go. I think I've broken some tests on linux along the way. I'll pick it back up on Monday. Please anybody chime in if you have opinions on the approach I'm taking.

jsignell:
I might add some test to check parsing of windows style urlpaths passed in the catalog. Does that seem like a thing that anyone would do?

jsignell:
Down to the last two windows failures. I need some help on these ones @martindurant 

```python-traceback

====================================================== FAILURES =======================================================
_________________________________________________ test_ndarray[ser0] __________________________________________________

ser = <intake.container.serializer.MsgPackSerializer object at 0x000001C79CA5B240>

    @all_serializers
    def test_ndarray(ser):
        expected_array = np.arange(35).reshape((5, 7))

        # Check roundtrip
>       array = ser.decode(ser.encode(expected_array, 'ndarray'), 'ndarray')

intake\cli\server\tests\test_serializer.py:39:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
intake\container\serializer.py:69: in decode
    return msgpack.unpackb(bytestr, object_hook=msgpack_numpy.decode)
..\Miniconda\envs\intake\lib\site-packages\msgpack\fallback.py:121: in unpackb
    ret = unpacker._unpack()
..\Miniconda\envs\intake\lib\site-packages\msgpack\fallback.py:602: in _unpack
    ret = self._object_hook(ret)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = {b'data': b'\x00\x00\x00\x00\x01\x00\x00\x00\x02\x00\x00\x00\x03\x00\x00\x00\x04\x00\x00\x00', b'kind': b'', b'nd': True, b'shape': [5, 7], ...}
chain = None

    def decode(obj, chain=None):
        """
        Decoder for deserializing numpy data types.
        """

        try:
            if b'nd' in obj:
                if obj[b'nd'] is True:

                    # Check if b'kind' is in obj to enable decoding of data
                    # serialized with older versions (#20):
                    if b'kind' in obj and obj[b'kind'] == b'V':
                        descr = [tuple(tostr(t) if type(t) is bytes else t for t in d) \
                                 for d in obj[b'type']]
                    else:
                        descr = obj[b'type']
                    return np.frombuffer(obj[b'data'],
>                               dtype=np.dtype(descr)).reshape(obj[b'shape'])
E                               ValueError: cannot reshape array of size 5 into shape (5,7)

..\Miniconda\envs\intake\lib\site-packages\msgpack_numpy.py:106: ValueError

___________________________________________________ test_conf_auth ____________________________________________________

    def test_conf_auth():
        with temp_conf({'auth': {'class': 'intake.auth.secret.SecretAuth',
                                 'kwargs': {'secret': 'test'}}}) as fn:
            env = os.environ.copy()
            env['INTAKE_CONF_FILE'] = fn
            with server(env=env, wait=5000):
                # raw request
                r = requests.get('http://localhost:5000/v1/info')
>               assert r.status_code == 403
E               assert 200 == 403
E                +  where 200 = <Response [200]>.status_code

intake\tests\test_config.py:72: AssertionError
================================== 3 failed, 258 passed, 2 skipped in 104.17 seconds ==================================

```

jsignell:
The second one was due to some local state. So I'm down to the last one test_ndarray[ser0]

jsignell:
GREEN!!!!

jsignell:
I am still seeing one failure locally with the serializer. I am going to do some mild testing on windows just to make sure that our tests actually ensure that intake works on windows.

jsignell:
These are the failures that I have locally: 

```python-traceback

====================================================== FAILURES =======================================================
_________________________________________________ test_ndarray[ser0] __________________________________________________

ser = <intake.container.serializer.MsgPackSerializer object at 0x0000020FAB708630>

    @all_serializers
    def test_ndarray(ser):
        expected_array = np.arange(35).reshape((5, 7))

        # Check roundtrip
>       array = ser.decode(ser.encode(expected_array, 'ndarray'), 'ndarray')

intake\cli\server\tests\test_serializer.py:39:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
intake\container\serializer.py:69: in decode
    return msgpack.unpackb(bytestr, object_hook=msgpack_numpy.decode)
..\Miniconda\envs\intake\lib\site-packages\msgpack\fallback.py:121: in unpackb
    ret = unpacker._unpack()
..\Miniconda\envs\intake\lib\site-packages\msgpack\fallback.py:602: in _unpack
    ret = self._object_hook(ret)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = {b'data': b'\x00\x00\x00\x00\x01\x00\x00\x00\x02\x00\x00\x00\x03\x00\x00\x00\x04\x00\x00\x00', b'kind': b'', b'nd': True, b'shape': [5, 7], ...}
chain = None

    def decode(obj, chain=None):
        """
        Decoder for deserializing numpy data types.
        """

        try:
            if b'nd' in obj:
                if obj[b'nd'] is True:

                    # Check if b'kind' is in obj to enable decoding of data
                    # serialized with older versions (#20):
                    if b'kind' in obj and obj[b'kind'] == b'V':
                        descr = [tuple(tostr(t) if type(t) is bytes else t for t in d) \
                                 for d in obj[b'type']]
                    else:
                        descr = obj[b'type']
                    return np.frombuffer(obj[b'data'],
>                               dtype=np.dtype(descr)).reshape(obj[b'shape'])
E                               ValueError: cannot reshape array of size 5 into shape (5,7)

..\Miniconda\envs\intake\lib\site-packages\msgpack_numpy.py:106: ValueError
__________________________________________________ test_flatten_flag __________________________________________________

multi_server = 'intake://localhost:5000'

    def test_flatten_flag(multi_server):
        cat = open_catalog(multi_server)
        assert list(cat) == ['cat1', 'cat2']
>       assert 'use_example1' in cat.cat1()

intake\cli\server\tests\test_server.py:255:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
intake\catalog\entry.py:79: in __call__
    return self.get(**kwargs)
intake\catalog\remote.py:79: in get
    page_size=self._page_size
intake\catalog\remote.py:100: in open_remote
    source = plugin_registry[response['plugin']](**response['args'])
intake\catalog\local.py:539: in __init__
    super(YAMLFileCatalog, self).__init__(**kwargs)
intake\catalog\base.py:94: in __init__
    self.force_reload()
intake\catalog\base.py:109: in force_reload
    self._load()
intake\catalog\local.py:557: in _load
    with file_open as f:
..\Miniconda\envs\intake\lib\site-packages\dask\bytes\core.py:177: in __enter__
    f = SeekableFile(self.fs.open(self.path, mode=mode))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <dask.bytes.local.LocalFileSystem object at 0x0000020FABB02DA0>
path = 'C:/Users/IEUser/AppData/Local/Temp/pytest-of-IEUser/pytest-22/test_flatten_flag0/cat1.yaml', mode = 'rb'
kwargs = {}

    def open(self, path, mode='rb', **kwargs):
        """Make a file-like object

        Parameters
        ----------
        mode: string
            normally "rb", "wb" or "ab" or other.
        kwargs: key-value
            Any other parameters, such as buffer size. May be better to set
            these on the filesystem instance, to apply to all files created by
            it. Not used for local.
        """
>       return open(self._normalize_path(path), mode=mode)
E       FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\IEUser\\AppData\\Local\\Temp\\pytest-of-IEUser\\pytest-22\\test_flatten_flag0\\cat1.yaml'

..\Miniconda\envs\intake\lib\site-packages\dask\bytes\local.py:58: FileNotFoundError
================================== 2 failed, 259 passed, 2 skipped in 125.42 seconds ==================================
```

intake/catalog/tests/test_local.py
```diff
@@ -362,13 +372,13 @@ def test_flatten_duplicate_error():
 
 
 def test_multi_cat_names():
-    fn = abspath("catalog_union*.yaml")
+    fn = abspath("catalog_union*.yml")
```
martindurant: Was this test wrong in the first place?

intake/cli/client/tests/test_local_integration.py
```diff
@@ -88,20 +91,22 @@ def test_discover():
...
 def test_get_pass():
     cmd = [ex, '-m', 'intake.cli.client', 'get', TEST_CATALOG_YAML, 'entry1']
-    process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+    process = subprocess.Popen(cmd, stdout=subprocess.PIPE,
+                               universal_newlines=True)
     out, _ = process.communicate()
 
-    assert 'Charlie1   25.0     3' in out.decode('utf-8')
-    assert 'Eve2   25.0     3' in out.decode('utf-8')
+    assert 'Charlie1   25.0     3' in out
```
martindurant: Potential py2/3 problem here?

intake/conftest.py
```diff
@@ -123,24 +125,24 @@ def http_server():
         p.communicate()
 
 
-@pytest.fixture()
+@pytest.fixture(scope='function')
```
martindurant: I thought that was the default...

intake/tests/test_config.py
```diff
@@ -61,23 +62,24 @@ def test_conf():
 
 
 def test_conf_auth():
-    with temp_conf({'auth': {'class': 'intake.auth.secret.SecretAuth',
+    with temp_conf({'port': 5556,
```
martindurant: Changed not to be the default port?

intake/utils.py
```diff
@@ -0,0 +1,10 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2012 - 2018, Anaconda, Inc. and Intake contributors
+# All rights reserved.
+#
+# The full license is in the LICENSE file, distributed with this software.
+#-----------------------------------------------------------------------------
+
+def make_path_posix(path):
+    return path.replace('\\', '/')
```
martindurant: This is a pretty important function, should have tests of its own. I wonder if we should only do this when `os.sep == '\\'`?

intake/catalog/tests/test_local.py
jsignell: I believe so unless intake is meant to see .yml as the same as .yaml.

martindurant: They are definitely not meant to be seen as the same, so it's a wonder that this worked before.

intake/cli/client/tests/test_local_integration.py
jsignell: Not sure. We should add py2 to the test matrix.

intake/conftest.py
jsignell: Indeed. I suspect I was trying to track something down and couldn't remember so was being explicit. I'll undo that

intake/cli/client/tests/test_local_integration.py
martindurant: py2 is a separate issue, probably needing a similar amount of work, if we choose to pursue it. I am just noting it for now.

intake/tests/test_config.py
jsignell: I was having issues where the port was already open in an unsecret way from other tests. Couldn't figure out how to make sure it was closed, so I just used a different port

intake/conftest.py
martindurant: It is fine to leave it, the Zen says that explicit is always better.

intake/catalog/tests/test_local.py
jsignell: It didn't fail, just printed an error.

intake/utils.py
jsignell: Definitely should have its own tests. I am not quite sure whether or not it should only be applied base on the os.sep. I imagine that depends on whether or not the user is allowed to use windows style paths in the catalog.

martindurant: Ah, you something like the case where a catalog is written on a windows machine and transferred to a linux one? I guess there's no way that the paths can make sense in such a situation anyway, so I would only apply in the windows-specific case.

jsignell: They might make sense if CATALOG_DIR is used.

martindurant: You have a point. OK, we can keep it as it is and see if further problems arise down the road.

martindurant:
Addressing the "worries":

> does serialization work properly on windows?

I don't see how this would be any different, and it is tested implicitly within the current tests.

> should CATALOG_DIR have a trailing /?

This one is trickier. Probably things should work whether or not a user remembers to put `{{CATALOG_DIR}}/...` with the slash, which would mean "yes" to the question. Where a catalog's dir is set (in yaml_file(s)_cat), should ensure that path is posix-like and has the trailing slash.

> does catalog flattening work properly on windows?

I don't see why this would be a problem. We do test catalog flattening.

jsignell:
This can go in once green

PR #236, opened 2019-01-18
Use entrypoints to manage drivers. Add subcommand.

danielballan:
I would like it to be possible to:

* Provide drivers that are discoverable by intake without necessarily packaging them in a package named `intake*`
* Have the option to _disable_ a specific intake driver from getting autodiscovered without uninstalling the package that provides it or disabling other drivers in that package

I think the [Jupyter notebook serverextension system](https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Distributing%20Jupyter%20Extensions%20as%20Python%20Packages.html) has settled on a nice way to manage this kind of configuration (after many iterations and pivots over the years). This PR imitates that system. It's just a first pass to evaluate interest and would need more careful thought before being merged.

Demo:

An `intake drivers` subcommand can list the drivers that are added to `intake.registry` at import time.

```
$ intake drivers list
netcdf                        intake_xarray.netcdf.NetCDFSource
opendap                       intake_xarray.opendap.OpenDapSource
rasterio                      intake_xarray.raster.RasterIOSource
remote-xarray                 intake_xarray.xarray_container.RemoteXarray
zarr                          intake_xarray.xzarr.ZarrSource
```

A verbose option includes `__file__` locations, potentially useful for untangling issues with environments.
```
$ intake drivers list -v
netcdf                        intake_xarray.netcdf.NetCDFSource @ /home/dallan/Repos/bnl/intake-xarray/intake_xarray/netcdf.py
opendap                       intake_xarray.opendap.OpenDapSource @ /home/dallan/Repos/bnl/intake-xarray/intake_xarray/opendap.py
rasterio                      intake_xarray.raster.RasterIOSource @ /home/dallan/Repos/bnl/intake-xarray/intake_xarray/raster.py
remote-xarray                 intake_xarray.xarray_container.RemoteXarray @ /home/dallan/Repos/bnl/intake-xarray/intake_xarray/xarray_container.py
zarr                          intake_xarray.xzarr.ZarrSource @ /home/dallan/Repos/bnl/intake-xarray/intake_xarray/xzarr.py
```

Now suppose I want to disable the 'zarr'` driver provided by intake_xarray. Perhaps I have a different implementation that I want to use with 'zarr' and I need to avoid the name collision.

```
$ intake drivers disable intake_xarray.xzarr.ZarrSource
$ intake drivers list
netcdf                        intake_xarray.netcdf.NetCDFSource
opendap                       intake_xarray.opendap.OpenDapSource
rasterio                      intake_xarray.raster.RasterIOSource
remote-xarray                 intake_xarray.xarray_container.RemoteXarray
$ python -c "import intake; print('zarr' in intake.registry)"
False
```

I can later re-enable it:

```
$ intake drivers enable intake_xarray.xzarr.ZarrSource
$ python -c "import intake; print('zarr' in intake.registry)"
True
```

The enable/disable state is stored in a separate YAML file for each driver in `~/.intake/drivers.d`, imitating the system used by Jupyter. For backward compatibility, drivers in packages that begin with `intake*` are included in the registry unless they are explicitly disabled. (That is, they need not have _any_ configuration in `~/.intake/drivers.d`.) Drivers in packages with other names can be explicitly enabled:

```
$ intake drivers enable offbrand_catalog.MongoMetadataStoreCatalog
$ intake drivers list
netcdf                        intake_xarray.netcdf.NetCDFSource
opendap                       intake_xarray.opendap.OpenDapSource
rasterio                      intake_xarray.raster.RasterIOSource
remote-xarray                 intake_xarray.xarray_container.RemoteXarray
zarr                          intake_xarray.xzarr.ZarrSource
mongo_metadatastore           offbrand_catalog.MongoMetadataStoreCatalog
```

The `enable` command created the following file at `~/.intake/drivers.d/offbrand_catalog.MongoMetadataStoreCatalog.yml`:

```yaml
offbrand_catalog.MongoMetadataStoreCatalog:                                                                                                                                                   
  enabled: true
```

As [documented by Jupyter](https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Distributing%20Jupyter%20Extensions%20as%20Python%20Packages.html#Automatically-enabling-a-server-extension-and-nbextension), packages can automatically enable their drivers at install time by using `data_files` in `setup.py` to place the corresponding files in `~/.intake/drivers.d/`.

### DIFF
--- a/docs/source/making-plugins.rst
+++ b/docs/source/making-plugins.rst
@@ -202,18 +202,94 @@ It is also important to note that source attributes should be set after ``read()
 Driver Discovery
 ----------------
 
+Intake discovers available drivers in three different ways, described below.
+After the discovery phase, Intake will automatically create
+``open_[driver_name]`` convenience functions under the ``intake`` module
+namespace.  Calling a function like ``open_csv()`` is equivalent to
+instantiating the corresponding data-source class.
+
+Entrypoints
+'''''''''''
+
+If you are packaging your driver into an installable package to be shared, you
+should add the following to the package's ``setup.py``:
+
+.. code-block:: python
+
+   setup(
+       ...
+       entry_points={
+           'intake.drivers': [
+               'some_format_name = some_package.and_maybe_a_submodule.YourDriverClass',
+               ...
+           ]
+       },
+   )
+
+Entry points are a way for Python packages to advertise objects with some
+common interface. When Intake is imported, it discovers all packages installed
+in the current environment that advertise ``'intake.drivers'`` in this way.
+
+Configuration
+'''''''''''''
+
+The intake configuration file can be used to:
+
+* Specify precedence in the event of name collisions---for example, if two different
+  ``csv`` drivers are installed.
+* Disable a troublesome driver.
+* Manually make intake aware of a driver, which can be useful for
+  experimentation and early development until a ``setup.py`` with an
+  entrypoint is prepared.
+* Assign a driver to a name other than the one assigned by the driver's
+  author.
+
+The commandline invocation
+
+.. code-block:: bash
+
+   intake drivers enable some_format_name some_package.and_maybe_a_submodule.YourDriverClass
+
+is equivalent to adding this to your intake configuration file:
+
+.. code-block:: yaml
+
+   drivers:
+     some_format_name: some_package.and_maybe_a_submodule.YourDriverClass
+
+You can also disable a troublesome driver
+
+.. code-block:: bash
+
+   intake drivers disable some_format_name
+
+which is equivalent to
+
+.. code-block:: yaml
+
+   drivers:
+     your_format_name: false
+
+Deprecated: Package Scan
+''''''''''''''''''''''''
+
 When Intake is imported, it will search the Python module path (by default includes ``site-packages`` and other
 directories in your ``$PYTHONPATH``) for packages starting with ``intake_`` and discover DataSource subclasses inside
 those packages to register.  drivers will be registered based on the``name`` attribute of the object.
 By convention, drivers should have names that are lowercase, valid Python identifiers that do not contain the word
 ``intake``.
 
-After the discovery phase, Intake will automatically create ``open_[driver_name]`` convenience functions under the
-``intake`` module namespace.  Calling a function like ``open_csv()`` is equivalent to instantiating the
-corresponding data-source class.
+This approach is deprecated because it is limiting (requires the package to
+begin with "intake_") and because the package scan can be slow. Using
+entrypoints is strongly encouraged. The package scan *may* be disabled by
+default in some future release of intake.
 
-To take advantage of driver discovery, give your installed package a name that starts with ``intake_`` and define
-your driver class(es) in the ``__init__.py`` of the package.
+Python API to Driver Discovery
+''''''''''''''''''''''''''''''
+
+.. autofunction:: intake.source.discovery.autodiscover
+.. autofunction:: intake.source.discovery.enable
+.. autofunction:: intake.source.discovery.disable
 
 .. _remote_data:
 
--- a/docs/source/making-plugins.rst
+++ b/docs/source/making-plugins.rst
@@ -282,7 +282,11 @@ By convention, drivers should have names that are lowercase, valid Python identi
 This approach is deprecated because it is limiting (requires the package to
 begin with "intake_") and because the package scan can be slow. Using
 entrypoints is strongly encouraged. The package scan *may* be disabled by
-default in some future release of intake.
+default in some future release of intake. During the transition period, if a
+package named ``intake_*`` provides an entrypoint for a given name, that will
+take precedence over any drivers gleaned from the package scan having that
+name. If intake discovers any names from the package scan for which there are
+no entrypoints, it will issue a ``FutureWarning``.
 
 Python API to Driver Discovery
 ''''''''''''''''''''''''''''''
--- a/intake/cli/client/subcommands/drivers.py
+++ b/intake/cli/client/subcommands/drivers.py
@@ -5,7 +5,7 @@
 # The full license is in the LICENSE file, distributed with this software.
 #-----------------------------------------------------------------------------
 '''
-
+CLI for listing, enabling, disabling intake drivers
 '''
 
 from __future__ import print_function
--- a/intake/cli/client/subcommands/drivers.py
+++ b/intake/cli/client/subcommands/drivers.py
@@ -8,8 +8,6 @@
 CLI for listing, enabling, disabling intake drivers
 '''
 
-from __future__ import print_function
-
 import logging
 log = logging.getLogger(__name__)
 
--- a/intake/cli/client/subcommands/drivers.py
+++ b/intake/cli/client/subcommands/drivers.py
@@ -73,12 +73,7 @@ class Drivers(Subcommand):
 
     def _enable(self, args):
         drivers_d = os.path.join(confdir, 'drivers.d')
-        # In Python 3 you can use exist_ok=True for this.
-        try:
-            os.makedirs(drivers_d)
-        except OSError as err:
-            if err.errno != errno.EEXIST:
-                raise
+        os.makedirs(drivers_d, exist_ok=True)
         for driver in args.drivers:
             filepath = '{}.yml'.format(os.path.join(drivers_d, driver))
             if os.path.isfile(filepath):
@@ -92,12 +87,7 @@ class Drivers(Subcommand):
 
     def _disable(self, args):
         drivers_d = os.path.join(confdir, 'drivers.d')
-        # In Python 3 you can use exist_ok=True for this.
-        try:
-            os.makedirs(drivers_d)
-        except OSError as err:
-            if err.errno != errno.EEXIST:
-                raise
+        os.makedirs(drivers_d, exist_ok=True)
         for driver in args.drivers:
             filepath = '{}.yml'.format(os.path.join(drivers_d, driver))
             if os.path.isfile(filepath):
--- a/intake/cli/client/subcommands/drivers.py
+++ b/intake/cli/client/subcommands/drivers.py
@@ -50,12 +50,13 @@ class Drivers(Subcommand):
         list.add_argument('-v', '--verbose', action='store_true', help='Show module path.')
         list.set_defaults(invoke=self._list)
 
-        enable = sub_parser.add_parser('enable', help='Enable one or more intake drivers.')
-        enable.add_argument('drivers', type=str, help='Module path and class name, as in package.submodule.ClassName', nargs='+')
+        enable = sub_parser.add_parser('enable', help='Enable an intake driver.')
+        enable.add_argument('name', type=str, help='Driver name')
+        enable.add_argument('driver', type=str, help='Module path and class name, as in package.submodule.ClassName')
         enable.set_defaults(invoke=self._enable)
 
         disable = sub_parser.add_parser('disable', help='Disable one or more intake drivers.')
-        disable.add_argument('drivers', type=str, help='Module path and class name, as in package.submodule.ClassName', nargs='+')
+        disable.add_argument('names', type=str, help='Driver names', nargs='+')
         disable.set_defaults(invoke=self._disable)
 
     def invoke(self, args):
@@ -71,9 +72,8 @@ class Drivers(Subcommand):
                   file=sys.stderr)
 
     def _enable(self, args):
-        for driver in args.drivers:
-            enable(driver)
+        enable(args.name, args.driver)
 
     def _disable(self, args):
-        for driver in args.drivers:
-            disable(driver)
+        for name in args.names:
+            disable(name)
--- a/intake/cli/client/subcommands/drivers.py
+++ b/intake/cli/client/subcommands/drivers.py
@@ -1,0 +1,109 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2012 - 2018, Anaconda, Inc. and Intake contributors
+# All rights reserved.
+#
+# The full license is in the LICENSE file, distributed with this software.
+#-----------------------------------------------------------------------------
+'''
+
+'''
+
+from __future__ import print_function
+
+import logging
+log = logging.getLogger(__name__)
+
+#-----------------------------------------------------------------------------
+# Imports
+#-----------------------------------------------------------------------------
+
+# Standard library imports
+import errno
+from importlib import import_module
+import inspect
+import os
+import sys
+
+# External imports
+import yaml
+
+# Intake imports
+from intake import __version__
+from intake.cli.util import Subcommand
+from intake.source.discovery import all_enabled_drivers
+from intake.config import confdir
+
+#-----------------------------------------------------------------------------
+# API
+#-----------------------------------------------------------------------------
+
+class Drivers(Subcommand):
+    '''
+    List, enable, and disable intake drivers.
+    '''
+
+    name = "drivers"
+
+    def initialize(self):
+        sub_parser = self.parser.add_subparsers()
+
+        list = sub_parser.add_parser('list', help='Show all intake drivers.')
+        list.add_argument('-v', '--verbose', action='store_true', help='Show module path.')
+        list.set_defaults(invoke=self._list)
+
+        enable = sub_parser.add_parser('enable', help='Enable one or more intake drivers.')
+        enable.add_argument('drivers', type=str, help='Python package name, such as intake_xarray', nargs='+')
+        enable.set_defaults(invoke=self._enable)
+
+        disable = sub_parser.add_parser('disable', help='Disable one or more intake drivers.')
+        disable.add_argument('drivers', type=str, help='Python package name, such as intake_xarray', nargs='+')
+        disable.set_defaults(invoke=self._disable)
+
+    def invoke(self, args):
+        self.parser.print_help()
+
+    def _list(self, args):
+        if args.verbose:
+            fmt = '{name:<30}{cls.__module__}.{cls.__name__} @ {file}'
+        else:
+            fmt = '{name:<30}{cls.__module__}.{cls.__name__}'
+        for name, cls in all_enabled_drivers().items():
+            print(fmt.format(name=name, cls=cls, file=inspect.getfile(cls)))
+
+    def _enable(self, args):
+        drivers_d = os.path.join(confdir, 'drivers.d')
+        # In Python 3 you can use exist_ok=True for this.
+        try:
+            os.makedirs(drivers_d)
+        except OSError as err:
+            if err.errno != errno.EEXIST:
+                raise
+        for driver in args.drivers:
+            filepath = '{}.yml'.format(os.path.join(drivers_d, driver))
+            if os.path.isfile(filepath):
+                with open(filepath, 'r') as f:
+                    conf = yaml.load(f.read())
+            else:
+                conf = {}
+            conf.update({driver: {'enabled': True}})
+            with open(filepath, 'w') as f:
+                f.write(yaml.dump(conf, default_flow_style=False))
+
+    def _disable(self, args):
+        drivers_d = os.path.join(confdir, 'drivers.d')
+        # In Python 3 you can use exist_ok=True for this.
+        try:
+            os.makedirs(drivers_d)
+        except OSError as err:
+            if err.errno != errno.EEXIST:
+                raise
+        for driver in args.drivers:
+            filepath = '{}.yml'.format(os.path.join(drivers_d, driver))
+            if os.path.isfile(filepath):
+                with open(filepath, 'r') as f:
+                    conf = yaml.load(f.read())
+            else:
+                conf = {}
+            conf.update({driver: {'enabled': False}})
+            with open(filepath, 'w') as f:
+                f.write(yaml.dump(conf, default_flow_style=False))
--- a/intake/cli/client/subcommands/drivers.py
+++ b/intake/cli/client/subcommands/drivers.py
@@ -32,6 +32,7 @@ from intake import __version__
 from intake.cli.util import Subcommand
 from intake.source.discovery import all_enabled_drivers
 from intake.config import confdir
+from ....utils import yaml_load
 
 #-----------------------------------------------------------------------------
 # API
@@ -82,7 +83,7 @@ class Drivers(Subcommand):
             filepath = '{}.yml'.format(os.path.join(drivers_d, driver))
             if os.path.isfile(filepath):
                 with open(filepath, 'r') as f:
-                    conf = yaml.load(f.read())
+                    conf = yaml_load(f.read())
             else:
                 conf = {}
             conf.update({driver: {'enabled': True}})
@@ -101,7 +102,7 @@ class Drivers(Subcommand):
             filepath = '{}.yml'.format(os.path.join(drivers_d, driver))
             if os.path.isfile(filepath):
                 with open(filepath, 'r') as f:
-                    conf = yaml.load(f.read())
+                    conf = yaml_load(f.read())
             else:
                 conf = {}
             conf.update({driver: {'enabled': False}})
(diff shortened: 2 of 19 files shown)


martindurant:
Before looking at this, I wanted to comment that a data source can specify its driver(s) by giving a fully qualified class name, `package.module.DriverClass`, and that a catalog yaml can give a set of modules to search for drivers at the time the file is parsed. Also, it is conceivable that a data source could be loaded by multiple drivers, and that multiple drivers happen to share the same simple name ( https://github.com/ContinuumIO/intake/pull/163 ).

martindurant:
@bryevdv , you may have opinions on managing a plugin system via configuration. Note that the data drivers are just one thing that is pluggable in Intake.

danielballan:
For my use case, I want data sources to stick with just a name, not a fully-specified DriverClass, to maintain the degree of freedom whereby the consumer of the data source can have their own opinions about which of the potentially-multiple drivers they want to use.

seibert:
I am also interested to see more flexible ways to deal with plugins.  The current system was basically copied from flask after I did some looking around for better approaches and gave up.

The main requirement (in my mind) is that installing a driver (with pip or conda) automatically makes it available in the registry without requiring an explicit registration step from the user.  Although most package managers support automated post-install scripts, it is much easier if they are not required.

Being able to manage plugins that offer alternative implementations of the same format is interesting.  That certainly makes it easier for people to experiment with better ways to read already-supported file formats.

martindurant:
Yeah, I think it's a good idea to be able to specify which class provides a particular driver, and which are not to be loaded at all, but I don't see an easy way to switch to a model where we don't do the package scan at all, since that's how thins are designed now.

On the CLI commands, I think they are basically correct (and exactly the same should be available in API also). The config like
```yaml
drivers:
  zarr: intake_xarray.xzarr.ZarrSource
  csv: false
  netcdf: intake_xarray.netcdf.NetCDFSource
```

where "csv" and `open_csv` will be missing, and "netcdf" represents a choice. `intake drivers -v` would then give details of a) entries in the config, whether they imported successfully, and b) the discover process, as each name is registered and any that are skipped because of configuration or because of name collision.

martindurant:
That way, any name that is not explicitly given in the config (which is an action done by a user), it will be automatically imported without placing special files in any place or calling import-time scripts.

danielballan:
> The main requirement (in my mind) is that installing a driver (with pip or conda) automatically makes it available in the registry without requiring an explicit registration step from the user. Although most package managers support automated post-install scripts, it is much easier if they are not required.

I agree completely. The breakthrough moment for me was when I learned that this can be done in a plain setup.py via the data_files argument. Other package managers can decide to use post-install hooks instead if they so choose, of course, as Jupyter’s documentation notes, but it’s not necessary.

> I don't see an easy way to switch to a model where we don't do the package scan at all, since that's how thins are designed now.

I agree. We could nonetheless undertake adding this mechanism to all known existing drivers and then add a config option to opt out of the package scan (and thus make import faster and arguably “cleaner”). That option could even potentially become the default in the *far* future after a long warning cycle.

martindurant:
> this can be done in a plain setup.py via the data_files argument

It it ends up in `~/.intake`, then I don't think conda will find it, so then we'd be wanting to search in multiple places. I can imagine it getting somewhat complex, and I'm scared of cases of "I installed X but I can't find it" down the line.

danielballan:
I have used this pattern in the past:

```py
if os.name == 'nt':
    _user_conf = os.path.join(os.environ['APPDATA'], 'databroker')
    CONFIG_SEARCH_PATH = (_user_conf,)
else:
    _user_conf = os.path.join(os.path.expanduser('~'), '.intake')
    _sys_prefix_etc = os.path.join(os.path.dirname(os.path.dirname(sys.executable)),
                               'etc', 'intake')
    _system_etc = os.path.join('/', 'etc', 'intake')
    CONFIG_SEARCH_PATH = (_user_conf, _sys_prefix_etc, _system_etc)
```

Of course one has to decide how to compose multiple potentially-comflicting configurations, but I think giving each driver a separate file makes this fairly clean. (I believe that was a factor in driving Jupyter to this approach.) The CLI can be explicit about where it found the config that is stipulating the state of each driver.

I don't want to make intake's configuration an order of magnitude more complicated all at once, but having a "sys prefix" solution for conda (and environments generally) may well be worth including in the first iteration.

martindurant:
Exactly where to put stuff is complex enough that there is [a module](https://github.com/ActiveState/appdirs) for it, and in intake we want the config directory to be itself configurable via an environment variable. In addition, we need a firm way to prevent different packages from having conflicting file-names for the nugget that they would want to insert into that location. Finally, it's not clear whether the driver preferences should be per install prefix, user, or global.

Sorry if I only seem to be raising problems!

danielballan:
Haha, no problem. Forethought are wariness are called for her, especially as we'll be stuck with maintaining whatever we decide approximately forever.

> we need a firm way to prevent different packages from having conflicting file-names for the nugget that they would want to insert into that location.

In this draft, intake expects to find config in a directory named exactly like the package. We can't prevent packages from stepping on another package's name in error, but I think we are safe from collisions between two _correct_ packages, no?

>  Finally, it's not clear whether the driver preferences should be per install prefix, user, or global.

I believe Jupyter looks in all three of those in that order, shows in the startup logs and CLIs' output which files it has found and used, and _merges_ them. Tricky but good for situations where the sysadmin provides some basics by the user wants to customize "just this one thing".

Nice, I hadn't encountered appdirs yet. It seems quite lightweight, may be worth using here.

martindurant:
Sorry to be slow to get to this... Some thoughts, much of which is already covered by this PR.

- the idea to be able to dis/able and list the set of drivers via the CLI is certainly very useful and a definite good addition. The same functionality should be available via the library API too, though
- with debug logging (e.g., INTAKE_LOG_LEVEL=DEBUG), the decisions of which modules are inspected and which drivers imported from where should be shown, i.e., the stages of `all_enabled_drivers` need logger output
- the current autodiscover behaviour must be maintained for backward compatibility for the foreseeable future; they do *not* need to added to an enabled list for this to happen, but they must not appear on a disabled list; collisions here should cause warnings
- I agree with the idea of a place where installed packages can put files which then mark these packages for import by intake without being names intake_*. That could be done via the CLI with a script, API call, or special install file hook. I regard the last as very brittle, because it brings the problems of exactly *where* the files should go: global, user, environment, config-dir, appdirs default ... which are not necessarily fixed at install time; and I don't think there's a way for setuptools and conda to use the *same* method for placing a file. This needs some thought
- the config directory can gain a new file detailing data drivers (could in theory be a section of the main config too): 
    - a set of specific modules/classes to import; The appearance of a "config.d" file is exactly equivalent to a module entered here
    - a set of specific modules/classes *not* to import, and 
    - a mapping `name: [<class path> | 'disabled']` for either expressing a preference for which driver should fulfil a given driver name, or prevent registry of all drivers of the given name. 
This does overlap latter point overlaps with the first, which may be rarely used, therefore. It is handy to prevent the import of a module before finding the name of the contained class and rejecting it, but we would need to decide precedence when something is in both the avoid list and the mapping.

Previously, individual catalogues could provide a more specific drivers config too (they can still provide what is effectively the first item), this could be considered for reimplementation.

martindurant:
Agree totally with that summary, @danielballan .

Shall we keep this PR, then, to the direct actionable items, and I'm OK to allow for intake-imported packages to announce themselves via a file-per-package, as you intended. That approach would still require a separate config file too, for driver name clashes. I would expect, for now, packages to directly call the intake API to announce themselves via a conda post-install hook, so we can keep the implementation flexible for the future.

minrk:
>  I recall that IPython/Jupyter moved from one central file to one-file-per-extension and trying the former for a release or two and finding it problematic. I am not sure I correctly recall their reasoning.

I don't have a link off the top of my head, but the principle is that most package installers (pip wheels, conda, apt) either can't (wheels) or greatly prefer not to (conda, apt) take actions to modify files at installation time. They like as much as possible for installation to be placing unmodified files in stable locations. So whatever action it is that you want to take place at install time (in our case, enabling an extension), it is much easier for packagers to handle that by placing a file, rather than by editing a file. There have been all kinds of issues with conda's post-link and pre-unlink hooks, which are where we can take these modification actions, and caused headaches (e.g. dependencies being unavailable because they are not ready, maybe the enable/disable step occurred manually already, etc.).

So as much as possible, I would encourage actions to be taken at install time to be:

1. plopping a file in a location
2. user-overrides to this *not* involving editing the same file, but editing a different file that takes precedence

to avoid conflicts between user actions and packager install/uninstall.

I'm not sure I can speak to whether conf.d or setuptools entrypoints or whatever is best for your case in terms of discoverability. We've had good success in nbconvert and more recently jupyterhub with using entrypoints purely for the discoverability step, as long as what is to be discovered is e.g. a Python function or class (not the case for nbextensions, but it is the case for nbconvert Exporters). @takluyver's [entrypoints](https://entrypoints.readthedocs.io) package makes using the entrypoints standard established by setuptools super nice without having to invoke pkg_resources and all the baggage that comes with. In our case for extensions, discoverability isn't enough because we don't want the set of enabled extensions to always be the set of *installed* extensions, which is part of what @danielballan is asking for, I think. But if that is okay for you, then entrypoints is probably a good solution and you can skip defining your own conf.d standard. Even if you want that to be the *default* behavior, you could probably get away without conf.d and have a user-config extension whitelist (and/or blacklist) for cases where "less than all installed" is desired. If either of these suits your use case well enough, that's probably what I would recommend before conf.d.

martindurant:
@danielballan , sounds like the many-file option is the clear winner, as you predicted. A general set of preferences is still needed, though (e.g., which driver to use for given data type name), which could be a file unto itself or in the main config. I think the many-files should be in the install directory so that pip and conda are ok with them (and they reflect installed packages local to that env anyway), but the config can stay in the confdir,  ~/.intake/ by default, for now.

danielballan:
Thanks for taking the time to provide that useful input, @Carreau and @minrk.

Agreed with all of the above, @martindurant. I think we have enough clarity at this point that I should move the code forward, and then we can kick the tires and discuss some more.

martindurant:
Thanks to all the people here for developing and for joining the conversation!

martindurant:
@danielballan , do you still have plans here?

danielballan:
Yep! Thanks for your patience -- I just got back from a nice long vacation.

danielballan:
@martindurant Thanks again for waiting for me to circle back on this. It is taking shape now, though there are still some details to work out, and it needs tests. The documentation update provides a decent summary of the design. Do you think this is converging in the right way?

P.S. I'll be on my way to Austin shortly---see you there!

martindurant:
I'll have a read, glad to see this going forward.

> I'll be on my way to Austin shortly

Unfortunately I am not there :|  Since there was no Intake tutorial granted, I decided to stay home this year with my pregnant wife.

danielballan:
!!! Congratulations! A solid choice! :-D

martindurant:
The current state of this PR looks pretty solid to me. I am trying to get a couple of opinions from colleagues, since this is a little outside of my normal experience. Anyone you would like to tag, @danielballan ?

docs/source/making-plugins.rst
```diff
@@ -202,18 +202,94 @@ It is also important to note that source attributes should be set after ``read()
...
+If you are packaging your driver into an installable package to be shared, you
+should add the following to the package's ``setup.py``:
+
+.. code-block:: python
+
+   setup(
+       ...
+       entry_points={
+           'intake.drivers': [
+               'some_format_name = some_package.and_maybe_a_submodule.YourDriverClass',
```
Carreau: I don't know enough about intake, but it seem that now multiple packages could register the same `some_format_name`, is that of any concern ? 

I also don't remember how entrypoints behave in that case and whether it would only list one entrypoint for that name, of multiples.

docs/source/making-plugins.rst
```diff
@@ -202,18 +202,94 @@ It is also important to note that source attributes should be set after ``read()
...
 By convention, drivers should have names that are lowercase, valid Python identifiers that do not contain the word
 ``intake``.
 
-After the discovery phase, Intake will automatically create ``open_[driver_name]`` convenience functions under the
-``intake`` module namespace.  Calling a function like ``open_csv()`` is equivalent to instantiating the
-corresponding data-source class.
+This approach is deprecated because it is limiting (requires the package to
+begin with "intake_") and because the package scan can be slow. Using
+entrypoints is strongly encouraged. The package scan *may* be disabled by
+default in some future release of intake.
```
Carreau: what happens for `intake_*` package that would _also_ provide entrypoints during migrations period ?

intake/cli/client/subcommands/drivers.py
```diff
@@ -0,0 +1,79 @@
...
+# Copyright (c) 2012 - 2018, Anaconda, Inc. and Intake contributors
+# All rights reserved.
+#
+# The full license is in the LICENSE file, distributed with this software.
+#-----------------------------------------------------------------------------
+'''
+
+'''
+
+from __future__ import print_function
```
Carreau: is intake Py2 compatible ?

intake/cli/client/subcommands/drivers.py
```diff
@@ -0,0 +1,79 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2012 - 2018, Anaconda, Inc. and Intake contributors
+# All rights reserved.
+#
+# The full license is in the LICENSE file, distributed with this software.
+#-----------------------------------------------------------------------------
+'''
+
+'''
```
Carreau: Empty module documentation, is that on purpose ?

intake/cli/client/subcommands/drivers.py
martindurant: It was initially, but we found it not worthwhile to maintain as other pydata projects were dropping support

intake/source/discovery.py
```diff
@@ -12,19 +12,158 @@
...
+    do_package_scan : boolean
+        Default is True.
+
+    Returns
+    -------
+    drivers : dict
+        Name mapped to driver class.
+    """
+    # Discover drivers via package scan.
+    if do_package_scan:
```
Carreau: Likely emit a `DeprecationWarning('deprecated since vX.Y.Z', stacklevel=2)` if plugin_prefix is passed. 
You can also set the default values to a sentinel (like None), to detect if people are passing the option or not.

martindurant: I would prefer to introduce the new capability without deprecating the previous behaviour, and have both live side-by-side for a while. Agree, though, that a specific warning is better than the general one.

intake/source/discovery.py
```diff
@@ -12,19 +12,158 @@
...
+                     entrypoint.module_name,
+                     entrypoint.object_name)
+
+    # TODO Un-comment this code to unleash warnings after most known intake_*
+    # packages have been updated to provide 'intake.drivers' entrypoints.
+    # if package_scan_results:
+    #     warnings.warn(
+    #         f"The drivers {package_scan_results} do not specify entry_points "
+    #         f"and were only discovered via a package scan. This may break in a "
+    #         f"future release of intake. The packages should be updated.")
```
Carreau: You could make this class a custom warning class that is filtered out by default, allowing people to turn it one via filterwarnings. Maybe a "FutureWarning" ?

martindurant: Great idea

intake/source/discovery.py
```diff
@@ -12,19 +12,158 @@
...
+        return entrypoint.load()
+    except ModuleNotFoundError:
+        raise ConfigurationError(
+            f"Failed to load {entrypoint.name} driver because module "
+            f"{entrypoint.module_name} could not be found.")
+    except AttributeError:
+        raise ConfigurationError(
+            f"Failed to load {entrypoint.name} driver because no object "
+            f"named {entrypoint.object_name} could be found in the module "
+            f"{entrypoint.module_name}.")
```
Carreau: If py3+ only, you may want to `raise ... from e`, the ModuleNotFound may not be the entrypoint itself, but one of its dependencies.  Same for below. AttributeError, it may be loading the entrypoint trigger some unrelated AttributeErrors, in which case the error message is might be wrong. It is the most likely cause – I agree –, but not always and, not raising from may make the issue quite hard to track.

intake/source/discovery.py
```diff
@@ -74,3 +212,38 @@ def load_plugins_from_module(module_name):
...
     return plugins
+
+
+class ConfigurationError(Exception):
+    pass
+
+
+def enable(name, driver):
+    """
+    Update config file drivers seciton to explicitly assign a driver to a name.
```
Carreau:
```suggestion
    Update config file drivers section to explicitly assign a driver to a name.
```

intake/source/discovery.py
```diff
@@ -74,3 +212,38 @@ def load_plugins_from_module(module_name):
...
+        Dotted object name, as in ``'intake_xarray.xzarr.ZarrSource'``
+    """
+    if 'drivers' not in conf:
+        conf['drivers'] = {}
+    conf['drivers'][name] = driver
+    save_conf()
+
+
+def disable(name):
+    """Update config file drivers seciton to disable a name.
```
Carreau:
```suggestion
    """Update config file drivers section to disable a name.
```

intake/source/discovery.py
martindurant: +1 (or at least logging)

martindurant:
Thanks for the review, @Carreau

docs/source/making-plugins.rst
danielballan: Good question. Entrypoints get precedence. More detail added in d79070e5ca8f20b03450bc998bcec5b9702e208d.

docs/source/making-plugins.rst
danielballan: The `entrypoints` package gives us `entrypoints.get_group_named` (return dict, drop duplicates) and `entrypoints.get_group_all` (return list). In 71c8eb12b4e65c10131046f70133e15c17c3c953 I added code to check for duplicates and debug-log that they are being ignored. 

In 596e96b80cb818927ee52df3331b43089a4b812d I added `autodiscover_all`, which is to `autodiscover` as `get_group_all` is to `get_group_named`. I used it in `intake drivers list` to show all plugins that are discovered but not enabled. They may be not enabled because of a name collision between two or more entrypoints, because of a name collision between entrypoints and package scan results, or because they are explicitly disabled by the configuration file.

intake/cli/client/subcommands/drivers.py
danielballan: Result of copy/paste from another CLI module. Updated, thanks.

intake/source/discovery.py
danielballan: Roger @martindurant. We can add this later.

danielballan:
Maybe @stuartcampbell, @tacaswell, and/or @jklynch would take a look.

intake/cli/client/subcommands/drivers.py
stuartcampbell: Should this line be removed then ?

intake/source/discovery.py
Carreau: it's just that the docstring says "Deprecated", and in my experience users don't read docstring. I'm just in favor of having both the docstring and the code be coherent, and give a release number so that it has a change of being cleaned up after. The `DeprecationWarning` or even `PendingDeprecationWarning` (which is foo by default), is usefull for integrator that run with `-We` to know about potential future deprecations.

intake/source/discovery.py
```diff
@@ -9,22 +9,253 @@
...
+        Name mapped to driver class.
+    """
+    # Discover drivers via package scan.
+    if do_package_scan:
+        package_scan_results = _package_scan(path, plugin_prefix)
+
+    # Discover drivers via entrypoints.
+    group = entrypoints.get_group_named('intake.drivers', path=path)
+    group_all = entrypoints.get_group_all('intake.drivers', path=path)
+    if group_all != group:
```
Carreau: group_all:list  , group:dict, so this will likely always be false ? – would need to check.
Force set-coercion, or compare length ?

intake/source/discovery.py
```diff
@@ -9,22 +9,253 @@
...
+        # There are some name collisions. Let's go digging for them.
+        for name, matches in itertools.groupby(group_all, lambda ep: ep.name):
+            matches = list(matches)
+            if len(matches) != 1:
+                winner = group[name]
+                logger.debug(
+                    "There are %d 'intake.driver' entrypoints for the name "
```
Carreau: f- all the strings ?

Carreau: Scratch that,  is a log.debug.

Carreau:
Thanks, it's a good start. 🤣

intake/source/discovery.py
danielballan: Ah, I meant this to be a len comparison.

martindurant:
In as far as this doesn't break existing functionality, I'm minded to merge. First we should have some tests, though, which in practice would mean working in a temporary config directory and calling both API and CLI functions to register/remove/forbid drivers. Also some tests which edit the config files directly? That tests the specific implementation of the config, which may change later.

danielballan:
OK, great, I'll move ahead with adding tests. FYI, I'm on vacation the rest of this week, so don't expect commits to be rolling in until next week.

intake/source/discovery.py
danielballan: @martindurant How would you feel about a `PendingDeprecationWarning` here?

martindurant: Is that like the "future" one, that doesn't show up unless specifically requested? In that case, yes.

intake/source/discovery.py
```diff
@@ -9,22 +9,253 @@
...
+    """Discover intake drivers including those registered for the same name.
+
+    Parameters
+    ----------
+    path : str or None
+        Default is ``sys.path``.
+    plugin_prefix : str
+        DEPRECATED. Default is 'intake_'.
+    do_package_scan : boolean
+        Default is True.
```
jsignell: Will this be deprecated as well? It seems like the default will at least be changed at some point. Not sure how best to expose that though.

docs/source/making-plugins.rst
```diff
@@ -202,18 +202,98 @@ It is also important to note that source attributes should be set after ``read()
...
+.. code-block:: yaml
+
+   drivers:
+     some_format_name: some_package.and_maybe_a_submodule.YourDriverClass
+
+You can also disable a troublesome driver
+
+.. code-block:: bash
+
+   intake drivers disable some_format_name
```
jsignell: How does this work for name collisions? Would they all be disabled?

intake/source/discovery.py
danielballan: Yes, I think the only warning type that is shown by default is UserWarning.

docs/source/making-plugins.rst
danielballan: Yes. The user could then enable one driver specifically or leave no drivers enabled for that name.

intake/source/discovery.py
danielballan: Added and confirmed that that the `PendingDeprecationWarning` is _not_ displayed when this codepath is hit by the CLI or in interactive use in IPython.

intake/source/discovery.py
danielballan: Good catch. Added in dcc84a0. Thanks!

martindurant:
Sounds like everything is ready here.
Does anything remain to be done?

Carreau:
> Does anything remain to be done?

Merge ?

martindurant:
OK then!
This may be a very nice little thing to write a blog about.

danielballan:
Thanks, all! I had planned to add tests before the merge, but I'll just open a separate PR. In parallel, I will start adding entrypoints to `intake-*` packages. (If they are used with an old version of intake, pre-dating this PR, they will do no harm.)

martindurant:
Thanks, @danielballan

PR #238, opened 2019-01-22
Exclude invalid identifiers from __dir__. Include normal attrs.

### LINKED ISSUE #235: dir gives all entries, even invalid ones
The names of entries in a catalog are allowed to be arbitrary strings. Since attribute access in python requires legal identifier names, only those entries with such names should be given by `dir()` (i.e., only those should show up for tab-completion).

danielballan:
This also adds a separate `__dir__` implementation for `RemoteCatalog`, as described in #235 and in tests and comments here.

Closes #235.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -191,10 +191,13 @@ class Catalog(DataSource):
     def __dir__(self):
         # Include tab-completable entries and normal attributes.
         return (
-            [entry for entry in self._get_entries() if
+            [entry for entry in self if
              re.match("[_A-Za-z][_a-zA-Z0-9]*$", entry)  # valid Python identifer
              and not keyword.iskeyword(entry)]  # not a Python keyword
             + list(self.__dict__.keys()))
+
+    def _ipython_key_completions_(self):
+        return list(self)
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


martindurant:
Any thoughts on defining `_ipython_key_completions_()` (see https://github.com/NCAR/intake-siphon/pull/2 ) ?

intake/catalog/base.py
```diff
@@ -364,6 +381,24 @@ def __init__(self, url, http_args={}, page_size=None, **kwargs):
     def _make_entries_container(self):
         return Entries(self)
 
+    def __dir__(self):
+        # Include (cached) tab-completable entries and normal attributes.
+        warnings.warn(
```
martindurant: I think this should only appear in the case that we know there are more entries than we are going to show, otherwise could be quite annoying

danielballan: Good point! Fixed and thoroughly tested.

danielballan:
Rebased on master, made the warnings appear only when applicable (needed one more piece of public state on `Entries` to make it possible cleanly), and implemented `_ipython_key_completions_`. Thanks for the suggestion!

martindurant:
Looks good, thank you!

PR #239, opened 2019-01-22
Compute len on server side.

danielballan:
Current behavior in `master` is that `len(catalog)` will cause all the entries to be fetched. The new behavior in this PR computes the length on the server side and sends it in response to `GET /info`. Catalogs may optionally implement `__len__` if they have an efficient means of computing their own length short of iterating over all their entries and counting them up.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -360,6 +360,7 @@ class RemoteCatalog(Catalog):
         self.http_args['headers'] = self.http_args.get('headers', {})
         self._page_size = page_size
         self._source_id = kwargs.get('source_id', None)
+        self._len = None
         if self._source_id is None:
             self.name = urlparse(url).netloc.replace(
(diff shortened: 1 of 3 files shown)


danielballan:
Mentioned in #235

intake/catalog/base.py
```diff
@@ -480,3 +484,12 @@ def _load(self):
                     auth=self.auth,
                     http_args=self.http_args, **source)
                  for source in info['sources']})
+
+    def __len__(self):
+        if self._len is None:
+            # The server is running an old version of intake and did not
+            # provide a length, so we have no choice but to do this the
+            # expensive way.
+            return sum(1 for entry in self)
```
martindurant: `len(self)` ?

martindurant:
OK, this one is straight-forward. Will merge after #234 , if there are no conflicts.

PR #240, opened 2019-01-24
Fix for trailing forward slash in CATALOG_DIR breaking pattern

jsignell:
Found this while working on intake-xarray but added tests that would fail here now if it breaks again.


intake/utils.py
```diff
@@ -6,5 +6,7 @@
 #-----------------------------------------------------------------------------
 
 def make_path_posix(path):
-    return path.replace('\\', '/')
-    
\ No newline at end of file
+    """ Make path generic """
+    if '://' in path:
+        return path
+    return path.replace('\\', '/').replace('//', '/')
```
jsignell: If you'd rather this logic just be in utils that would also be fine with me.

Issue #228, opened 2019-01-14
Is Python 2 support currently intended?


danielballan:
I couldn't find a general issue tracking the status of (or roadmap for) Python 2.7 support in intake. I notice some 2+3-isms in the source (usage of `six`, etc.) but I also notice that CI tests are [only run against 3.6](https://github.com/ContinuumIO/intake/blob/master/.travis.yml#L15). Is full support for 2.7 still intended? If so, shall I add a 2.7 build to the CI to verify it is not accidentally broken?

As a separate question, I would personally advocate for building consensus to drop Python 2.7 as soon as practical.

seibert:
Given how new intake is, and the abandonment of Python 2 by the SciPy ecosystem happening in 2019, I don't think it is unreasonable to drop Python 2 by the end of the year.

That said, since the current code base should work for Python 2, I think it makes sense to add it to the CI configuration to make sure it does until dropping Python 2 is agreed upon.

martindurant:
I completely agree with your concerns, and encourage other people to speak up on this issue.

I believe that at the moment, tests will fail for py2, and the pypi/conda selectors are for py3. Originally, Intake was developed with py2 in mind, and most of the code will be fine, but this has slipped as some of the plugins required py3-only dependencies, and external packages like numpy and dask are actively pursuing py3-only right now (or, in some cases, are having conversations about the same). 

From our point of view, it would be reasonable to investigate how much work py2 support would be for the base, thinking that there may well be legacy data loading packages that Intake wants to make use of.

seibert:
I agree that plugins may have to drop Python 2 on a different and quicker schedule.  If the Intake core can preserve/restore Python 2 compatibility for the short term, that would at least give plugin authors the option to decide what they want to do.

danielballan:
One possibility would be to re-instate Python 2 support, cut a release, and then declare intake Python 3 only from there on out, as happened [in this case](https://github.com/dask/dask-kubernetes/pull/114#issuecomment-453314929).

martindurant:
As things stand, I think it would take some work to get a py2 version out. I'll look into how bad it is.

danielballan:
[...]

PR #243, opened 2019-01-28
Explicit __init__ signatures for Catalog, RemoteCatalog

danielballan:
I have changed the `__init__` of `Catalog` and `RemoteCatalog` to make their signatures explicit (as opposed to `Catalog(*args, **kwargs)`).

The only reason I can think of _not_ to do this is that `**kwargs` is a decent way to force keyword-only arguments in Python 2. In Python 3, of course, we can spell out the keyword arguments explicitly _and_ make them keyword-only. It may make sense to put this PR on the back burner until #228 is resolved.

The reasons _to_ do this are:
- `Catalog` presently silently ignores keyword parameters that are misspelled by the user. With this change, it would raise and alert the user rather than silently ignoring unrecognized input.
- There are couple places where intake seems to be relying on unused keyword arguments to Catalog subclasses falling through and eventually being ignored by Catalog in `**kwargs`. I think it would be better to catch these further up (as in 7733472 for example).

The tests currently do not pass, and I have question about how best to resolve it. `open_remote` passes `parameters=user_parameters` into the `container` it is constructing:

https://github.com/ContinuumIO/intake/blob/6b084f426caf070da141ec044f660fe1f921cee1/intake/catalog/remote.py#L108-L110

Some containers use this information, but `RemoteCatalog` was previously swallowing it in `**kwargs` and ignoring it. Should it:
(1) start accepting `parameters` and do something with them?
(2) explicitly ignore `parameters`
(3) not accept `parameters` in which we have to fix `open_remote` to check if the `container` is a catalog


martindurant:
I think 7733472 is exactly a reason to consider this PR carefully: it does not look good to me to explicitly have a parameter in the signature that is meant to be ignored. The way you phrase the docstring for that parameter makes clear that you don't much like this either ;)

As a point of reference, the DataSources (as well as filesystem backends and others) in general tent to take **kwargs, and pass them on, perhaps modifies, or just pluck off the ones they are interested in. Then either things get ignored, or you get the unexpected keyword error further into the call stack. It's a pretty common pattern and is indeed fragile to misspellings.

On balance, I think it would be best to be explicit about all of the reasonably expected keywords, and in cases where the kwargs are to be passed on, expect an explicit appropriately named dict.
e.g, 

```
def __init__(self, par1, par2, kw1=None, **kwargs):
    self.thing = kwargs.pop('thing', None)
    manipulate(kwargs)
    call_other(**kwargs)

to 

def __init__(self, par1, par2, kw1=None, thing=None, other_kwargs=None, **kwargs):
    self.thing = thing
    call_other(**(other_kwargs or {}))
```

This still allows for silent ignores, but at least we can describe what goes where, as well as allow for the signature to change in the future without backward incompatibility.

danielballan:
Haha, guilty as charged. My preferred solution would be to "fix" it in the caller, to not passed `flatten` to something to which it is not applicable. I guessed swallowing it "higher up" was a less invasive change.

I agree that passing kwargs down is a common pattern, but if a kwarg makes it all the way down to the _base class_,`Catalog`, and no subclass has picked it off yet, I think we should raise. If we need to add a kwarg in the future, we can still do that safely. If we need to _remove_ one, I would argue that it's better to raise ("We removed that option and may not be doing what you want us to do!") than to silently ignore. Am I missing your point about backward compatibility; can you illustrate with an example?

danielballan:
Oops -- power-cycled via stray mouse click. Sorry about that.

Thanks for being open-minded! :- D

Any thoughts on whether `RemoteCatalog` should be doing anything with `UserParameters` or accepting them at all?

martindurant:
I can image a RemoteCatalog making use of parameters, which is rather similar in fact to what your implementation of search does now. As with other sources, parameters could be passed to the server in order to instantiate the catalog object there.

intake/catalog/base.py
```diff
@@ -438,15 +448,19 @@ def fetch_page(self, page_offset):
...
-            getshell=self.getshell,
-            auth=self.auth,
-            http_args=self.http_args,
-            page_size=self._page_size,
-            **source)
-            for source in info['sources']}
+        page = {}
+        for source in info['sources']:
+            user_parameters = source.get('user_parameters', [])
+            # TODO Do something with self._parameters.
```
danielballan: I'm confused about how to merge the `source['user_parameters']` (from the server) with `self._parameters` that would have come in through `open_remote`. Are they expected to just be the same thing?

danielballan:
This now passes tests (locally anyway -- we'll see if Travis agrees) and is ready for another round of review. I still have a conceptual confusion and need some guidance. See comment in line. Thanks!

intake/catalog/base.py
danielballan: To elaborate: The code here replicates the pre-existing behavior. Previously, `parameters` were passed in by `open_remote` and then dropped on the floor by `Catalog`'s `**kwargs`. Now `parameters` is picked off by `RemoteCatalog` and...dropped on the floor. The question is how/whether to merge those parameters with those in `source`.

martindurant: I think it is OK to leave like this until we know we need it. As you say, there is more than one way to pass these things around, I expect some evolution.

PR #245, opened 2019-01-30
Include license in sdists

toddrme2178:
The terms of the BSD-2-Clause license require all copies of the software to include the license text


martindurant:
Thanks, that is very reasonable.

PR #246, opened 2019-01-31
Adding waffle badge

jsignell:
Should update as we update the columns.


martindurant:
I can only suppose that code chunk is correct. 
Any idea why we now get two appveyor builds?

PR #249, opened 2019-02-01
Fix missing dependencies when installing with pip

psarka:
intake reports missing dependencies (toolz, yaml and cloudpickle) when pip installed from the pypi, or when trying to run tests against current master. This PR fixes that.

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -8,7 +8,7 @@
 import logging
 import os
 import posixpath
-import yaml
+from ruamel import yaml
 
 try:
     from ruamel.yaml.constructor import DuplicateKeyError
--- a/intake/cli/client/subcommands/cache.py
+++ b/intake/cli/client/subcommands/cache.py
@@ -21,7 +21,7 @@ log = logging.getLogger(__name__)
 import os
 
 # External imports
-import yaml
+from ruamel import yaml
 
 # Intake imports
 from intake.cli.util import Subcommand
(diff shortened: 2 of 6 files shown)


martindurant:
Thanks, I'll try to look at this soon.
I am not too worried about test dependencies, though, since we don't yet have testing instructions. pip dependencies often go without.

psarka:
Do you remember what was the reason for Ruamel? I replaced it with pyyaml and all tests pass locally.

martindurant:
Where are toolz and cloudpickle used in the codebase?

martindurant:
If everything passes without it, then I daresay we no longer need ruamel. I think it was used to give a better indication of where in a file an error occurred, but we don't do that any more. @seibert , any comment?

psarka:
Both toolz and cloudpickle come from dask during the import. Missing cloudpickle gets triggered in line 100 in `intake/source/csv.py` and missing toolz in line 19 in `intake/source/cache.py`.

It is possible to get both by requiring `dask[bag]` or `dask[dataframe]`. Would that be preferrable to requiring toolz and cloudpicke explicitly? The  relevant section of dask setup.py is as follows:

```python
extras_require = {
  'array': ['numpy >= 1.11.0', 'toolz >= 0.7.3'],
  'bag': ['cloudpickle >= 0.2.1', 'toolz >= 0.7.3', 'partd >= 0.3.8'],
  'dataframe': ['numpy >= 1.11.0', 'pandas >= 0.19.0', 'toolz >= 0.7.3',
                'partd >= 0.3.8', 'cloudpickle >= 0.2.1'],
  'distributed': ['distributed >= 1.22'],
  'delayed': ['toolz >= 0.7.3'],
}
```

PR #253, opened 2019-02-04
Factor dependencies differently.

danielballan:
Following up on #249, I liked @psarka's suggestion of depending on
``dask[bag]`` to obtain ``cloudpickle`` and ``toolz``, since intake itself does
not import either directly.

Just submitted for consideration --- no strong opinion on this.


martindurant:
I suggest we clean up the remaining PRs and return the this before the next release.

PR #255, opened 2019-02-04
Give UserParameter an eval-able repr.

danielballan:
This was useful to me during debugging and may be useful to others.


martindurant:
Totally in favour, but could use a test

martindurant:
On second thoughts, do you think there's any merit to having this more verbose form only for `str`, so the user has to explicitly print in order to see it?

danielballan:
Fine by me. Changed to `__str__`; added test.

martindurant:
Thanks!

PR #256, opened 2019-02-05
Raise validation error when parameter has no default

psarka:
Documentation specifies that 'Every parameter must have a default'. I missed this at first, and had some head scratching errors when the file failed to be found.

If it is indeed the case that default is not optional, I think it is helpful to inform the user when it is missing.

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -390,7 +390,7 @@ class CatalogParser(object):
             'name': name,
             'description': self._getitem(data, 'description', str),
             'type': self._getitem(data, 'type', str, choices=valid_types),
-            'default': self._getitem(data, 'default', object, required=False),
+            'default': self._getitem(data, 'default', object),
             'min': self._getitem(data, 'min', object, required=False),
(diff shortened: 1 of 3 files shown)


martindurant:
The alternative, up for discussion, would be to have an exception when trying to instantiate from a spec with a user parameter, no default, and nothing specified by the user.

danielballan:
I gather the original motivation for the requirement of always providing a default was to ensure that eliding the get() call and going straight to read() always returns _something_.

I think the alternative proposed by @martindurant preserves a point of flexibility that could be useful and has little downside.

psarka:
ok, I'll take a look and will update this accordingly.

psarka:
Unfortunately, I ran out of spare time, and it is unlikely that I will manage to finish this one anytime soon. I will close it to reduce the clutter, if someone want's to pick it up, feel free to reopen.

martindurant:
Thanks for raising it and making a start. It will be done at some point, but not sure when.

PR #261, opened 2019-02-06
Updating travis to run coveralls

jsignell:
This PR should add a coveralls hook


martindurant:
Did this work? I can see that the runs passed, but I don't know what I should expect from coveralls to know that stats were successfully transferred.

jsignell:
I wasn't quite sure. The build shows up on coveralls, but I think it needs to get coverage on master to set a baseline.

martindurant:
So we should merge to see what happens?

jsignell:
:)

PR #263, opened 2019-02-08
Plugin missing message

### LINKED ISSUE #262: Uninformative errors when plugins are missing
Here is an example of the sort of error you might see if `intake-parquet` isn't installed when it is required:

image

To a newcomer, this message does not convey the root problem that `intake-parquet` is missing. Ideally the message would be something like:

```
'no plugins loaded for this entry: parquet. Please install the intake-parquet package.'
```

Unfortunately, this isn't possible as intake doesn't know which packages supply which drivers. Instead, the error message could link to the [plugin's page](https://intake.readthedocs.io/en/latest/plugin-directory.html) like this:

```
'no plugins loaded for this entry: parquet. A listing of installable plugins can be found 
 at https://intake.readthedocs.io/en/latest/plugin-directory.html.'
```

The plugin page could then be enhanced with a list the drivers offered by each of the plugin packages, making it easier for users unfamiliar with intake to find the missing plugin that they need to install.

martindurant:
Fixes #262

### DIFF
--- a/docs/source/plugin-directory.rst
+++ b/docs/source/plugin-directory.rst
@@ -3,29 +3,32 @@
 Plugin Directory
 ================
 
-This is a list of known projects which install driver plugins for Intake:
+This is a list of known projects which install driver plugins for Intake, and the named drivers each
+contains in parentheses:
 
-* `intake-astro <https://github.com/ContinuumIO/intake-astro>`_ Table and array loading of FITS astronomical data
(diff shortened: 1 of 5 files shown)


seibert:
I think you might have accidentally mixed changes in the persistence branch with this one.  :)

martindurant:
@seibert fixed

Issue #254, opened 2019-02-04
Cache files get created in the current directory


psarka:
Perhaps thanks to https://github.com/ContinuumIO/intake/issues/151, cache files are now always created in the current directory, even though the `cache_dir` is `~/.intake/` and documentation mentions that `~/.intake/` is the default location (and it ends up containing `cache_metadata.json`, just not the files themselves)

Looking here https://github.com/ContinuumIO/intake/blob/da2809d13045dc97dd3753f5ac4bcc9e19299d74/intake/source/cache.py#L322 it seems to me that the cache directory is not taken into account when creating `files_out`.

repro.py:
```
import intake
intake.open_catalog('123.yml').onetwothree.read()
```

123.yml:
```
metadata:
  version: 1

sources:
  onetwothree:
    description: bucket csv test
    driver: csv
    args:
      urlpath: 'gs://puodas/123.csv'
      storage_options:
        token: "anon"
    cache:
      - argkey: urlpath
        regex: 'gs://puodas/'
        type: file
```

martindurant:
I believe it will work if regex is 'puodas/123.csv' or just 'puodas/' or even `'puodas/.*'`.  The regex is supposed to be "the part of the path that makes the url unique" or something like that; apparently it gets treated differently from the real urlpath, which I don't yet understand. So that off behaviour should be fixed (it is not related to the "use directory path" PR I don't think), but the documentation of how to formulate the regex is more important. Your help appreciated, once you have it working.

martindurant:
@mmccarty , thoughts on correct usage here?

mmccarty:
You are correct Martin. Drop the protocol from the regex.

jsignell:
I have also found the regex confusing and don't think it is necessary. Was the goal to not duplicate identical files that are called using different protocols (https vs s3 for instance)?

martindurant:
The current very brief [documentation](https://intake.readthedocs.io/en/latest/catalog.html#caching-source-files-locally) is as follows

> regex: A regular expression to match against the URL path, where the matching portion will be replaced by a path in the local cache directory.

The idea is, that without being explicit with the matching string, it is hard to know how much of the urlpath is unique per file, for example s3://bucket/folder1/file.csv, s3://bucket/folder2/file.csv - we don't want the two file.csv to clash. It could perhaps be done by finding the common posixpath root (we already have code for this). The set of remote files may be different each time, but that probably doesn't worry us, since we either cache or don't, there is no incremental update.

[...]

PR #264, opened 2019-02-11
Make regex field optional

psarka:
This PR makes the regex field optional.

The issues of potentially [illegal path](https://github.com/ContinuumIO/intake/issues/265) or file name collisions are not affected by this PR.

Alleviates https://github.com/ContinuumIO/intake/issues/254. It does not fully resolve it, as if user does provide a regex which does not match, files are still dropped in the cwd instead of `cache_dir`.

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -259,12 +259,11 @@ the value when accessing the data.
 Caching Source Files Locally
 ''''''''''''''''''''''''''''
 
-To enable caching on the first read of remote data source files, ``cache`` specifications have the following attributes
-in the catalog.
+To enable caching on the first read of remote data source files, add the ``cache`` section with the
+following attributes:
 
-- ``argkey``: Of the keys in the args section in this same data source, which contains the URL(s) of the data to be cached.
-- ``regex``: A regular expression to match against the URL path, where the matching portion will be replaced by a path in the local cache directory.
-- ``type``: One of the keys in the cache registry [`intake.source.cache.registry`], referring to an implementation of caching behaviour. The default if "file" for the caching of one or more specific remote files.
+- ``argkey``: The args section key which contains the URL(s) of the data to be cached.
+- ``type``: One of the keys in the cache registry [`intake.source.cache.registry`], referring to an implementation of caching behaviour. The default is "file" for the caching of one or more files.
 
 Example:
 
@@ -275,7 +274,6 @@ Example:
     driver: csv
     cache:
       - argkey: urlpath
-        regex: '{{ CATALOG_DIR }}/cache_data'
         type: file
     args:
       urlpath: '{{ CATALOG_DIR }}/cache_data/states.csv'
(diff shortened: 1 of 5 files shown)


martindurant:
As it happens, this seems to make an illegal path on windows

psarka:
oh dear :man_facepalming:

docs/source/catalog.rst
```diff
@@ -287,6 +285,11 @@ The special value ``"catdir"`` implies that cached files will appear in the same
 catalog file in which the data source is defined, within a directory named "intake_cache". These will
 not appear in the cache usage reported by the CLI.
 
+Optionally, the cache section can have a ``regex`` attribute, that modifies the path of the cache on
+the disk. By default, the cache path is made by concatenating ``cache_dir``, dataset name, hash of
+the url, and the url itself (without the protocol). ``regex`` attribute allows to remove part of the
```
martindurant: Is there a better description of regex than this? It feels a little hard to understand - but indeed, the usage *is* hard to uderstand.

intake/source/cache.py
```diff
@@ -91,11 +91,15 @@ def _ensure_cache_dir(self):
     def _munge_path(self, cache_subdir, urlpath):
         import re
 
-        regex = sanitize_path(self._spec['regex'])
         path = sanitize_path(urlpath)
 
+        if 'regex' in self._spec:
+            regex = r'%s' % sanitize_path(self._spec['regex'])
+        else:
+            regex = r'^'  # if no regex provided, use full path
```
martindurant: Interesting minimum-code solution!

docs/source/catalog.rst
psarka: I was not able to come up with something clear and succinct, so decided to explicitly spell out what it does. 

What could be a good idea, is to warn user about cache files ending up in cwd if regex is not well specified. Where well specified regex to my understanding is one that matches part of the url that includes the beginning, but excludes the protocol. Should I add this?

intake/source/cache.py
psarka: :)

docs/source/catalog.rst
martindurant: > warn user about cache files ending up in cwd

Sorry, I thought that would no longer be the case, that they would instead end up in the cache directory, under the hash, but with a deeper path

martindurant:
Fixes #254

docs/source/catalog.rst
martindurant: > Should I add this?

but yes, I think exactly that wording would be good

psarka:
Instead of documenting accidental writing to cwd I realized I can fix it easily.

Note that to be sure that file always ends up in the cache dir I removed the `return url_path if cache_path == path logic`. I have to admit I don't understand why would `url_path` be useful as `cache_path`, and the only failing unit test (which I removed too), did not reveal the usefulness.

martindurant:
Fair enough. Could you please a test showing that files do not end up in the cwd for the failed regex case, but they do end up in the catalog directory path if so specified?

psarka:
I think this should test that. Can you please take a look?

martindurant:
Thanks for all your work and patience, @psarka

PR #271, opened 2019-02-19
Revisit kwargs in open_remote.

danielballan:
This passes more Catalog parameters through and I think a little more
clear as well.

### DIFF
--- a/intake/catalog/remote.py
+++ b/intake/catalog/remote.py
@@ -86,12 +86,14 @@ class RemoteCatalogEntry(CatalogEntry):
             self.url, self.name, container=self.container,
             user_parameters=user_parameters, description=self.description,
             http_args=http_args,
-            page_size=self._page_size
-            )
+            page_size=self._page_size,
+            auth=self.auth,
+            getenv=self.getenv,
+            getshell=self.getshell)
 
 
 def open_remote(url, entry, container, user_parameters, description, http_args,
-                page_size=None):
+                page_size=None, auth=None, getenv=None, getshell=None):
     """Create either local direct data source or remote streamed source"""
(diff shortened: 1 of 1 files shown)


intake/catalog/remote.py
```diff
@@ -109,16 +111,20 @@ def open_remote(url, entry, container, user_parameters, description, http_args,
...
+            response.pop('container')
+            response.update({'name': entry, 'parameters': user_parameters})
             if container == 'catalog':
-                # Propagate the page_size setting into nested Catalogs.
-                response['page_size'] = page_size
-                response.pop('container')
-            source = container_map[container](
-                url, http_args, parameters=user_parameters,
-                name=entry, **response)
-        source.description = response.get('description', description)
```
danielballan: I replaced this with `source.description = description` above because the response only _has_ a description if it is a direct-access response, so there was no need to do it in the proxied-access case as well.

martindurant:
@danielballan , perhaps we should catch up in the next couple of days, if you have the time.

martindurant:
To clarify: this PR affects the case that a remote catalog is opened from within another remote catalog - then the three parameters getenv, getshell and auth should be passed on. 

I think this is a good idea in general. Do you think it's possible that the auth parameter (or all three) should be overridable somehow? I'm thinking that the new remote cat may be on a different server - is this possible?

danielballan:
It does seem like it might be possible and useful to allow for. At this point I don't think I have a firm enough grasp of the different ways that information can flow to be sure how to do implement that.

martindurant:
OK, I will merge this now, and we can postpone working on the idea I surfaced above. Perhaps that should be written into a new issue to track it.

PR #277, opened 2019-02-21
Add DAT cache mechanism

### LINKED ISSUE #267: DAT as a caching mechanism
See https://docs.datproject.org/ecosystem
DAT is a secure peer-to-peer directory sharing service popular for large scientific datasets, whereby the whole directory contents is a dereferencable blob, and the network transport is automatically handled.

This is an ideal cache use case for Intake, you just have the reference, and DAT puts the files in the right place.

cc @pzwang

martindurant:
Fixes #267

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -4,7 +4,7 @@ Glossary
 .. glossary::
 
     Cache
-        Local copies of remote files. Intake allows for download-on-first-use for data-soruces,
+        Local copies of remote files. Intake allows for download-on-first-use for data-sources,
         so that subsequent access is much faster, see :ref:`caching`. The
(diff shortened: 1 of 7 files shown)


martindurant:
@pzwang , how does this look? The incantation in the test is:
```
  dat_data:
    driver: textfiles
    args:
      urlpath: 'dat://2c3bcbba714723fc95da82b17e3575627e98ed28da3f9815ec7e32245185c014/*.json'
      decoder: !!python/name:json.loads
    cache:
      - type: dat
```
i.e., the input to the reader is a set of paths in the downloaded DAT folder, in this case a single file.

martindurant:
(happens to be a JSON file, but could have read as text without the `decoder:` line)

pzwang:
Looks fine to me!

PR #279, opened 2019-02-25
Updating references to ContinuumIO -> intake

jsignell:
I think I got them all with the exception of gitter which doesn't seem to have moved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # Intake: A general interface for loading data
 
-![Logo](https://github.com/ContinuumIO/intake/raw/master/logo-small.png)
+![Logo](https://github.com/intake/intake/raw/master/logo-small.png)
 
-[![Build Status](https://travis-ci.org/ContinuumIO/intake.svg?branch=master)](https://travis-ci.org/ContinuumIO/intake)
-[![Coverage Status](https://coveralls.io/repos/github/ContinuumIO/intake/badge.svg?branch=master)](https://coveralls.io/github/ContinuumIO/intake?branch=master)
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1,12 +1,12 @@
 # Intake: A general interface for loading data
 
-![Logo](https://github.com/ContinuumIO/intake/raw/master/logo-small.png)
+![Logo](https://github.com/intake/intake/raw/master/logo-small.png)
 
-[![Build Status](https://travis-ci.org/ContinuumIO/intake.svg?branch=master)](https://travis-ci.org/ContinuumIO/intake)
-[![Coverage Status](https://coveralls.io/repos/github/ContinuumIO/intake/badge.svg?branch=master)](https://coveralls.io/github/ContinuumIO/intake?branch=master)
+[![Build Status](https://travis-ci.org/intake/intake.svg?branch=master)](https://travis-ci.org/intake/intake)
```
martindurant: This one breaks - Travis is still watching ContinuumIO, which is probably fine to leave as it is.

jsignell: I don't see it as broken

martindurant: Sorry, must have been some kind of lag

martindurant:
The failure is on the DAT cache, the test dataset seems to have disappeared

martindurant:
Can you just set intake/source/tests/test_cache::test_dat to xfail for now?

martindurant:
Seems like the way to get DAT to successfully find files, download and exit is to host them on the same machine... will look into changing the test to do that later.

jsignell:
Ok I'll let it fail

PR #280, opened 2019-02-26
More persist

martindurant:
Further persist updates

- persist catalogues
- persist remote sources
- persist sequence sources

### DIFF
--- a/.ci-coveragerc
+++ b/.ci-coveragerc
@@ -1,2 +1,2 @@
 [run]
-omit = intake/*, *tests/*, */_version.py
+omit = *tests/*, */_version.py
--- a/docs/source/api_user.rst
+++ b/docs/source/api_user.rst
@@ -30,22 +30,22 @@ These are reference class and function definitions likely to be useful to everyo
    the plugins in the registry at import time.
 
 .. autoclass:: intake.source.csv.CSVSource
-   :members: __init__, discover, read_partition, read, to_dask
+   :members: __init__, discover, read_partition, read, to_dask, persist, export
 
 .. autoclass:: intake.source.zarr.ZarrArraySource
-   :members: __init__, discover, read_partition, read, to_dask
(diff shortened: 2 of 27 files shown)


martindurant:
@danielballan , you might find the changes here interesting, specially around serialisation of user parameters and catalogues in general. You can now persist the result of one of your catalogues which was made from a mongo query.

martindurant:
@rabernat , you may also find this interesting. You can persist the contents of a THREDDS lookup
```
cat = intake.open_thredds_cat('http://dap.nci.org.au/thredds/catalog.xml')
cat2 = cat.persist()
```
(even better if the original entry is itself within a catalogue, rather than created with `open_thredds_cat`)
Here, `cat2` is a catalogue with the same entries, but stored in a local YAML file - so you can skip the lookup step, which can take a while in some cases. Soon, can also `.persist()` the data sources themselves, which will create local zarr output. *Eventually* the location of that output can be configurable to external storage like GCS.

martindurant:
@seibert , what do you think of `[499090e](/intake/intake/pull/280/commits/499090ef664b2ac9838e5d1f749373d46d3d83bd) ` ?

martindurant:
The only remaining thing I mean to add to this (perhaps overloaded) PR is to persist or export to remote. Currently there are various `os` calls.

martindurant:
Coming up against the `dask.bytes`/`fsspec` limit here: the filesystem shims in Dask don't have `rm()`, but of course any self-respecting file-system should. Replacing the file-system stuff in the persist module with fsspec works fine; but dask.bytes is called in many other places, and the two main remote stores (s3fs, gcsfs) have only experimental PRs for fsspec compatibility. Not completely sure how well they would work with the original master branch.

Thought? To make things pass here, could simply avoid exception when attempting to remove files; but the previous commit called `shutils.rmtree` just fine.

Issue #225, opened 2019-01-09
Next-generation GUI


martindurant:
Supersedes #183, #182, #156, #95

Following the simple but useful jupyter-widgets GUI:
![GUI](https://intake.readthedocs.io/en/latest/_images/gui2.png)

The previous GUI is useful, and only works in, the jupyter-notebook context. The selected item becomes available to the user, so that they can continue their analysis in normal code.

A new GUI should be based on panel or bokeh, so that it can be run within a notebook or stand-alone with the same experience. For the stand-alone version to be useful, it must be able to interact with the data to some degree, rather than just listing datasets and entry details.

Here we want to sketch out ideas for a data application on Intake, which can be stand-alone. 

## Phase 1

Initially, this can be the same as the previous version, but with the following addition:
- show "discover" information about selected dataset
- display head/sample of the data for given dataset
- pick between the pre-determined plots to show, or allow arbitrary selection of the methods and arguments available in hvplot

This extra information can be a box below the main area, which can be similar to what we have now.

## Phase 2

Currently, the file browser only navigates the local filesystem, but we have the ability to view any. The "load URL" box shows that we can actually load from any place. Would be nice to be able to import catalogues from wherever by browsing. 
Similarly, the search interface (API and GUI) is very simplistic. intake-bluesky is showing that something more powerful can be done. For the GUI, it should be possible to save a catalogue created via search.
In a similar vein, copy/paste of data-sets between catalogues, including starting a new cat and copying things into it, are the kinds of things that a full data app should be able to do.
At some point, "persist" should be an option in the interface, which simply calls that method on the selected data-set.

## Phase 3

Many online platforms in this space have a "create data source" wizard, in which you can select data type, select files/query, and for each choose from appropriate options to define a file source, insert it into the given catalogue and save the result. 
This wizard would take some thinking about... and obviously lends itself to creating derived data-sets in the manner that Mosaic once did.

martindurant:
@bryevdv

rsignell-usgs:
I was thinking about a GSOC project for [Pangeo](pangeo.io) to build a "[ncview](http://meteora.ucsd.edu/~pierce/ncview_home_page.html)" like interface, but using Panel, Intake and hvplot.   The user would:
* paste the URL for an intake catalog into a text widget
* select a dataset from drop-down widget 
* select a variable from a drop-down widget 

Using xarray.hvplot, render map of the data and if there are additional dimensions to the data, display widgets to select indices for those dimensions.

[...]

Issue #14, opened 2018-01-08
Pick location of "default" catalog files


seibert:
Intake should have a "default" catalog composed from catalog files found in a standard location.
What should the search path for the catalog files be?

Proposal is to autoload `intake/catalog.d/*.{yaml,yml}` under the following directories:
- `$CONDA_PREFIX` - for conda-installable catalogs
- *What is the equivalent prefix for virtualenv?* - for pip-installable catalogs
- `~/.local/` - Linux user data dir
- `~/Library/` - macOS user data dir
- `%APPDATA%\` - Windows user data dir

The default catalog should be the union of all catalog files from these locations.

martindurant:
Suggest to close

seibert:
Yeah, this is done and implemented.

PR #286, opened 2019-03-05
New Panel GUI

jsignell:
## Try it locally
At the moment this work depends on development version of panel and bokeh:
    
    conda install -c conda-forge panel==0.5.1 hvplot==0.4.0 bokeh==1.1.0
    jupyter labextension install @pyviz/jupyterlab_pyviz  # if using jupyterlab

**Then**: take a look at examples/GUI.ipynb or do `panel serve intake/gui/server.py`

Towards #225 

Feature comparability with ipywidgets GUI:
 - [x] local files
 - [x] remote files
 - [x] catalog selection
 - [x] data source selection
 - [x] data source description
 - [x] catalog search
 - [x] initialization as intake.gui

Additional functionality:
 - [x] windows paths
 - [x] make sure that control buttons reflect visible status
 - [x] add home button to local file browser
 - [x] unit tests
 - [x] make path editable
 - [x] disable buttons when they don't make sense
 - [x] use properties over methods
 - [x] try to abstract widget from functionality
 - [x] choose from defined plots


martindurant:
Good to see it coming along!
For the final point of installing it in `intake.gui`, this could be a config option defaulting to Panel and the ipywidgets one can maybe be deprecated later.

By the way, the attribute `intake.gui` conflicts with the package `intake.gui`, which is bad design on my part and maybe this is a good time to fix it. I don't know what you think of `InstanceMaker`, whose job is not to create the GUI instance until it is accessed. At the very least, it should have docstrings to say that this is what it's for.

martindurant:
Please can you provide the latest look of the UI?

jsignell:
Here is what it looks like right now **on master panel and latest dev release bokeh**. The vision is that each section will have meaningful inputs and outputs so that the full gui can be very composable:

#### With 1 catalog - 0 selected  and no extra sections visible
screen shot 2019-03-07 at 3 13 00 pm

#### With 1 catalog - 1 selected  and no extra sections visible
screen shot 2019-03-07 at 3 13 11 pm

#### With 1 catalog - 1 selected  and add catalog from local visible
screen shot 2019-03-07 at 3 13 27 pm

#### With 2 catalogs - 1 selected  and add catalog from remote visible
screen shot 2019-03-07 at 3 13 56 pm

#### With 3 catalogs - 1 selected and search catalog visible
screen shot 2019-03-07 at 3 22 19 pm

#### With 2 catalogs - 2 selected and both search and add catalog visible
screen shot 2019-03-07 at 3 25 42 pm

jsignell:
Note that I have the plot section controller in the layout, but it is disabled for now.

martindurant:
Thanks, @jsignell !

intake/gui/source_select.py
```diff
@@ -0,0 +1,154 @@
...
+    options = None
+
+    def callback(self, *events):
+        print(self.panel.name, events)
+        for event in events:
+            if event.name == 'value' and event.new != self.selected:
+                self.select(event.new)
+
+    def update_selected(self):
+        if self.widget.value != self.selected:
```
philippjfr: Is the condition here necessary? I'd expect this not to trigger an event if the value hasn't changed. Not saying to remove the check, just checking you didn't add it to get around a param bug.

philippjfr: Also it seems like these ``update_X`` methods could be implemented by making the ``options`` and ``selected`` parameters of the class and using ``param.depends('selected', watch=True)``. I understand that you are avoiding these param patterns though, so maybe just ignore me or leave it as a suggestion for the future.

martindurant: Thanks for keeping an eye on this @philippjfr . The choice of the correct pattern here is interesting, to discuss how the param model fits into a GUI-app model. Would be happy to see alternate patterns, but that would likely mean discarding some work later on.

jsignell: yeah I did this one first and I think it is more convoluted then it needs to be. I was trying to make sure that a widget or a call could set the value on the class. But it might be overkill

jsignell:
Some errors that I see when running the GUI as a server:

```python-traceback
RuntimeError: _pending_writes should be non-None when we have a document lock, and we should have the lock when the document changes
```

```python-traceback
2019-03-08 15:49:17,153 Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x120d96510>, <Future finished exception=KeyError('examples/',)>)
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/tornado/ioloop.py", line 758, in _run_callback
    ret = callback()
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/tornado/stack_context.py", line 300, in null_wrapper
    return fn(*args, **kwargs)
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/tornado/ioloop.py", line 779, in _discard_future_result
    future.result()
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/tornado/gen.py", line 1147, in run
    yielded = self.gen.send(value)
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/bokeh/server/session.py", line 70, in _needs_document_lock_wrapper
    result = yield yield_for_all_futures(func(self, *args, **kwargs))
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/bokeh/server/session.py", line 191, in with_document_locked
    return func(*args, **kwargs)
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/bokeh/document/document.py", line 1126, in wrapper
    return doc._with_self_as_curdoc(invoke)
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/bokeh/document/document.py", line 1112, in _with_self_as_curdoc
    return f()
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/bokeh/document/document.py", line 1125, in invoke
    return f(*args, **kwargs)
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/bokeh/document/document.py", line 915, in remove_then_invoke
    return callback(*args, **kwargs)
  File "/Users/jsignell/panel/panel/viewable.py", line 766, in _change_event
    self.set_param(**self._process_property_change(events))
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/param/parameterized.py", line 1136, in inner
    return fn(*args, **kwargs)
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/param/parameterized.py", line 2477, in set_param
    return self_or_cls.param.set_param(*args,**kwargs)
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/param/parameterized.py", line 1282, in set_param
    self_._batch_call_watchers()
  File "/Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/param/parameterized.py", line 1395, in _batch_call_watchers
    watcher.fn(*events)
  File "/Users/jsignell/panel/panel/viewable.py", line 704, in param_change
    msg = self._process_param_change({event.name: event.new})
  File "/Users/jsignell/panel/panel/widgets/select.py", line 75, in _process_param_change
    msg['value'] = [hashable(mapping[v]) for v in msg['value']]
  File "/Users/jsignell/panel/panel/widgets/select.py", line 75, in <listcomp>
    msg['value'] = [hashable(mapping[v]) for v in msg['value']]
KeyError: 'examples/'
```

philippjfr:
>RuntimeError: _pending_writes should be non-None when we have a document lock, and we should have the lock when the document changes

Is this one new? I recently removed some guards against this because it seemed like they weren't necessary but I may have been too ambitious on that front.

jsignell:
> Is this one new? 

I hadn't tried to run it as a server yet and I've only seen it in that context - so I'm not sure.

jsignell:
The KeyError issue can be seen in the notebook by doing:

```python
intake.gui.selector.visible = True
print('airline_flights/' in intake.gui.selector.fs.main.options)
intake.gui.selector.fs.main.value = ['airline_flights/']
```

philippjfr:
I can reproduce, but if I look at ``intake.gui.selector.fs.main.options`` after the error I see that it doesn't include ``'airline_flights/'``:

```
{'airline.yaml': 'airline.yaml', 'meta.yaml': 'meta.yaml'}
```

It looks to me like it's executing ``FileSelector.move_down`` before the value change has been fully processed. I'll have to think about what that means but it may be the case that we have to give different watchers different precedences so that it executes the ones internal to panel before it executes user-defined ones like yours. Alternatively we'd have to move to a global event queue as I had suggested in https://github.com/pyviz/param/issues/322.

philippjfr:
I haven't been able to reproduce the ``_pending_writes`` error on bokeh server. If you can come up with a set of steps to reproduce that that would be super helpful.

philippjfr:
On a separate note, I'd very much like to see a FileBrowser widget much like the one you've build in panel itself.

martindurant:
> I'd very much like to see a FileBrowser widget

One of the objectives here is to produce compound widgets that can be of general use to Panel!

philippjfr:
Fixed the KeyError in the file browser issue in this PR: https://github.com/pyviz/panel/pull/291

I still have to think about perhaps allowing watchers with different precedences to avoid this kind of thing in general but I'll make that a separate issue.

jsignell:
@philippjfr - I am still running into KeyErrors. For instance when I do:

```python 
intake.gui

intake.gui.searcher.visible = True
intake.gui.browser.cat.add('./airline_flights/airline.yaml')
intake.gui.searcher.text.value = 'flight'
intake.gui.searcher.widget.clicks += 1
print('airline_search' in intake.gui.browser.cat.options)
intake.gui.browser.cat.select('airline_search')
```

philippjfr:
>@philippjfr - I am still running into KeyErrors. For instance when I do:

Should be fixed now.

jsignell:
Screen Shot 2019-03-13 at 6 00 47 PM

martindurant:
OOoooohhh!

(not sure you need the text of the plot visually, but it makes it more clear what the example is doing)

martindurant:
Is there a way to have a border around the "control" region above, and the "output" region below?

jsignell:
I did a background - I had been feeling like it was getting too washed out.

Screen Shot 2019-03-14 at 9 02 35 AM

intake/gui/tests/test_source_view.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def test_defined_plots_set_source_to_empty_list_and_visible_to_false(defined_plots):
+    defined_plots.visible = False
+    defined_plots.source = []
+    assert_is_empty(defined_plots, visible=False)
+
+
+def test_defined_plots_select_a_different_plot(defined_plots):
+    """ This one is a genuine failure """
```
jsignell: @philippjfr any ideas why setting selected doesn't stick? It works in when selecting the widget with clicks, but not programatically.

jsignell:
BLOCKED: waiting on bokeh and panel releases and UI feedback

hdsingh:
Hi, I am interested in [Issue #14 · ESIPFed/gsoc · GitHub](https://github.com/ESIPFed/gsoc/issues/14).
I wanted to have a look at your GUI to get better idea. I am trying to run this GUI by doing following:
1. `git checkout -b jsignell/panel_gui`
2. `conda install -c pyviz/label/dev -c bokeh/label/dev panel hvplot bokehjupyter labextension install @pyviz/jupyterlab_pyviz` in a new environment.
3. Running jupyter in outer intake folder in created environment. --> did not worked.
4. Installing intke via `python setup.py develop` . --> did not worked.
However both step 3 and 4 are giving me same error in `GUI.ipynb` : `cannot import name 'EntryGUI'`.

Can you please help me know how to run `examples/GUI.ipynb` correctly ?

jsignell:
I guess I was being sloppy with the setup. I made an environment file with the requirements. I also rebased on master - so if you have the branch you'll need to reset your local version.

```
conda env create -f examples/environment.yml
conda activate intake-gui
pip install -e . --no-deps

# if you want to use jupyterlab
conda install -c conda-forge jupyterlab
jupyter labextension install @pyviz/jupyterlab_pyviz
```

martindurant:
> jupyter labextension install @pyviz/jupyterlab_pyviz

Isn't this automatically done when installing pyviz?

jsignell:
> Isn't this automatically done when installing pyviz?

Not that I know of. It's only for jupyterlab though

philippjfr:
As far as I know there is no recommended or supported approach to
installing Jupyterlab extensions as part of a conda install.

jsignell:
Note that a similar step is required to use ipywidgets in jupyterlab - https://ipywidgets.readthedocs.io/en/stable/user_install.html#installing-the-jupyterlab-extension

jsignell:
I verified that the GUI works in Edge - although it won't work in IE until https://github.com/pyviz/holoviews/issues/3103 is resolved.

hdsingh:
@jsignell Thanks a lot for your help!!
I have been able to run `GUI.ipynb` using the above insturctions. There were some additional dependencies which I have added in  #307 .

jsignell:
## To Do
- [x] make control buttons toggles 
- [x] don't pass around lists of widgets
- [x] nested catalog
- [ ] ~~look for unicode space~~  **did not work**
- [x] make removal work and for parents
- [ ] ~~indicate parent catalogs~~ **too time-consuming** 
- [ ] user parameter passing - **save this for next piece of work**
- [x] make plot description an optional area
- [x] scrolling in entry yaml output
- [x] clean up and add tests
- [ ] good datasets - anaconda download data, image classifier, landsat (ndvi, temp)

jsignell:
I have finished the work that I am planning to do for this first iteration of the new GUI. I decided to not do some of the bullet points that we had discussed. 

### Get rid of | in favor unicode space
I tried. The selector seems to exist in a weird area between str and markdown. In the end I decided that | is fine for now, and we should just focus on implementing a tree widget in bokeh if that is really what we need (see https://github.com/bokeh/bokeh/issues/8830).

### Indicate parent catalogs
In order to indicate parent catalogs, I'd need to inspect all the entries in every catalog to check what their container type is. This seemed like a large burden to handle up-front, so I decided that it might make sense to wait until the functionality is requested. Again it seemed very possible that there is a whole class of users that never use nested catalogs (maybe they have very flat and long catalogs) and checking for nestedness might just hurt them. Now I am thinking that there could be a checkbox to toggle this on and off.

### User parameter passing
I wasn't sure how common a pattern this is and didn't know where to put it in the GUI. I think part of this lack of clarity comes from conflating entries and sources. User parameters are important when initializing an entry to a source, so they should be at that transition, but the GUI doesn't ever initialize the source, so maybe it doesn't belong anywhere?  I did some experiments using `panel.interact` that I think will be helpful once we are ready to implement this.

```python
user_params =entry._user_parameters
if len(user_params) > 0:
    up_args = {}
    for up in entry_user_parameters:
        if up.min is not None and up.max is not None:
            up_args[up.name] = (up.min, up.max, None, up.default)

@pn.interact(**up_args)
def func(**kwargs):
    source = entry(**kwargs)
```

### Good datasets - anaconda download data, image classifier, landsat (ndvi, temp)
I am still using the crime catalog in the GUI notebook, but will likely write the blog post about some other dataset.

martindurant:
- `import intake; intake.gui` doesn't display because panel extension is not yet loaded; is there a way to do that automatically?
- selecting a catalog which refreshes the entry list (i.e., almost every time) causes the plot button to be momentarily toggled on then off, some event propagation
- perhaps, for now, the plots icon should be disabled when an entry has no predefined plots
- the test cat "catalog1" contains "nested" which refers to itself. If you load the cat and select "nested" you get the next level of nested but cannot select it, perhaps because the two cats have the same name

Screen Shot 2019-04-17 at 09 26 39

- if you search a cat, you get a new cat with name_search. If you search again, it gets overwritten
- when yaml loading fails while adding cat, nothing happens
- when a new cat is added, the list scrolls to the top and the new cat may be off the bottom
- in the image below, selecting the entry two below the current one causes the selection to pop back the where it is now; the details momentarily update. Note that the selector truncates the text of the current selection

Screen Shot 2019-04-17 at 09 40 49

- horizontal scroll is applied to the whole UI, not just the entry details
- if search would produce multiple entries of the same name, only one shows up (this might be a limitation of the search function)
- between selecting an entry and its details popping up, there is noticeable lag; perhaps this is fixed after my PR

flutefreak7:
Just read the blog entry and came to look at the progress. Damn! I'm exhausted just reading about all you've done, @jsignell! Good stuff!

PR #287, opened 2019-03-06
Separate out parameter munging

### LINKED ISSUE #257: Resolve getenv everywhere?
I was setting up gcloud authentication tokens following [this](https://github.com/ContinuumIO/intake/issues/195#issuecomment-437504787) and thought, why are `getenv`s limited to default field for the parameters? If there were expanded everywhere, I could setup the tokens easier:

```
sources:
  gcs_csv:
    description: "CSV data on GCS; requires gcsfs and a token"
    args:
      urlpath: gcs://private_bucket/path/*.csv
      storage_options:
        token: "getenv(GOOGLE_APPLICATION_CREDENTIALS)"
```

Does more broad expansion of `getenv` have some drawbacks?

martindurant:
Makes parameter evaluation clearer, and allows for user override of
any source input, with optional validation

Fixes #257

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -10,11 +10,11 @@ datasets, called *catalog entries*.  A catalog entry for a dataset includes info
 
 In addition, Intake allows the arguments to data sources to be templated, with the variables explicitly
 expressed as "user parameters". The given arguments are rendered using ``jinja2`` and the named user
-parameters, and validation is also provided for the allowed types and values for both the template
+parameters. Those parameters are also offer validation of the allowed types and values, for both the template
 values and the final arguments passed to the data source. The parameters are named and described, to
-indicate to the user what they are for.
-This kind of structure can be used to, for example,
-choose between two parts of a given data source, like "latest" and "stable".
+indicate to the user what they are for. This kind of structure can be used to, for example,
+choose between two parts of a given data source, like "latest" and "stable", see the `entry1_part` entry in
+the example below.
 
 
 YAML Format
@@ -45,10 +45,10 @@ Intake catalogs are typically described with YAML files.  Here is an example:
         description: entry1 part
         parameters: # User parameters
           part:
-            description: part of filename
+            description: section of the data
             type: str
-            default: "1"
-            allowed: ["1", "2"]
+            default: "stable"
+            allowed: ["latest", "stable"]
         driver: csv
         args:
           urlpath: '{{ CATALOG_DIR }}/entry1_{{ part }}.csv'
@@ -137,21 +137,19 @@ Templating
 
 Intake catalog files support Jinja2 templating for driver arguments. Any occurrence of
 a substring like ``{{field}}`` will be replaced by the value of the user parameters with
-that same name, or explicitly provided by the user. For how to specify these user parameters,
+that same name, or the value explicitly provided by the user. For how to specify these user parameters,
 see the next section.
 
 Some additional values are available for templating. The following is always available:
-
-- ``CATALOG_DIR``: The full path to the directory containing the YAML catalog file.  This is especially useful
-  for constructing paths relative to the catalog directory to locate data files and custom drivers.
-
-so that the search for CSV files for the two "entry1" blocks, above, will happen in the same directory as
+``CATALOG_DIR``, the full path to the directory containing the YAML catalog file.  This is especially useful
+for constructing paths relative to the catalog directory to locate data files and custom drivers.
+For example, the search for CSV files for the two "entry1" blocks, above, will happen in the same directory as
 where the catalog file was found.
 
 The following functions `may` be available. Since these execute code, the user of a catalog may decide
 whether they trust those functions or not.
 
-- ``env("USER")``: look in the environment for the named variable
+- ``env("USER")``: look in the set environment variables for the named variable
 - ``client_env("USER")``: exactly the same, except that when using a client-server topology, the
   value will come from the environment of the client.
 - ``shell("get_login thisuser -t")``: execute the command, and use the output as the value. The
@@ -189,12 +187,12 @@ A parameter may look as follows:
 
     parameters:
       name:
-        description: human-readable text for what this parameter means
-        type: optional, one of bool, str, int, float, list[str], list[int], list[float], datetime
-        default: optional, value to assume if user does not override
-        allowed: optional, list of values that are OK, for validation
-        min: optional, minimum allowed, for validation
-        max: optional, maximum allowed, for validation
+        description: name to use  # human-readable text for what this parameter means
+        type: str  # optional, one of bool, str, int, float, list[str], list[int], list[float], datetime
+        default: normal  # optional, value to assume if user does not override
+        allowed: ["normal", "strange"]  # optional, list of values that are OK, for validation
+        min: "n"  # optional, minimum allowed, for validation
+        max: "t"  # optional, maximum allowed, for validation
 
 A parameter, not to be confused with an :term:`argument`,
 can have one of two uses:
@@ -209,7 +207,7 @@ can have one of two uses:
   coerced to the given type of the parameter and validated against the allowed/max/min. It is therefore possible
   to use the string templating system (e.g., to get a value from the environment), but pass the final value as,
   for example, an integer. It makes no sense to provide a default for this case (the argument already has a value),
-  but will no raise an exception.
+  but providing a default will not raise an exception.
 
 Note: the ``datetime`` type accepts multiple values:
 Python datetime, ISO8601 string,  Unix timestamp int, "now" and  "today".
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -8,12 +8,14 @@ datasets, called *catalog entries*.  A catalog entry for a dataset includes info
 * Arguments to the ``__init__()`` method of the driver
 * Metadata provided by the catalog author (such as field descriptions and types, or data provenance)
 
-In addition, Intake allows datasets to be *parameterized* in the catalog.  This is most commonly used to allow the
-user to filter down datasets at load time, rather than having to bring everything into memory first.  The data
-parameters are defined by the catalog author, then templated into the arguments for ``__init__()`` to modify the data
-being loaded.  This approach is less flexible for the end user than something like the
-`Blaze expression system <https://blaze.readthedocs.io/en/latest/expr-compute-dev.html>`_, but also significantly
-reduces the implementation burden for driver authors.
+In addition, Intake allows the arguments to data sources to be templated, with the variables explicitly
+expressed as "user parameters". The given arguments are rendered using ``jinja2`` and the named user
+parameters, and validation is also provided for the allowed types and values for both the template
+values and the final arguments passed to the data source. The parameters are named and described, to
+indicate to the user what they are for.
+This kind of structure can be used to, for example,
+choose between two parts of a given data source, like "latest" and "stable".
+
 
 YAML Format
 -----------
@@ -41,7 +43,7 @@ Intake catalogs are typically described with YAML files.  Here is an example:
 
       entry1_part:
         description: entry1 part
-        parameters: # User defined parameters
+        parameters: # User parameters
           part:
             description: part of filename
             type: str
@@ -51,15 +53,6 @@ Intake catalogs are typically described with YAML files.  Here is an example:
         args:
           urlpath: '{{ CATALOG_DIR }}/entry1_{{ part }}.csv'
 
-Templating
-''''''''''
-
-Intake catalog files support Jinja2 templating for driver arguments. Any occurrence of
-a substring like ``{{field}}`` will be replaced by the value of the user parameters with
-that same name. Some additional values are always available:
-
-- ``CATALOG_DIR``: The full path to the directory containing the YAML catalog file.  This is especially useful
-  for constructing paths relative to the catalog directory to locate data files and custom drivers.
 
 Metadata
 ''''''''
@@ -68,6 +61,8 @@ Arbitrary extra descriptive information can go into the metadata section. Some f
 claimed for internal use and some fields may be restricted to local reading; but for now the only
 field that is expected is ``version``, which will be updated when a breaking change is made to the
 file format. Any catalog will have ``.metadata`` and ``.version`` attributes available.
+
+Note that each source also has its own metadata.
 
 Extra drivers
 '''''''''''''
@@ -125,7 +120,7 @@ returned data.  Each data source has several attributes:
 - ``driver``: Name of the Intake :term:`Driver` to use with this source.  Must either already be installed in the current
   Python environment (i.e. with conda or pip) or loaded in the ``plugin`` section of the file.
 
-- ``args``: Keyword arguments to the ``open()`` method of the driver.  Arguments may use template expansion.
+- ``args``: Keyword arguments to the init method of the driver.  Arguments may use template expansion.
 
 - ``metadata``: Any metadata keys that should be attached to the data source when opened.  These will be supplemented
   by additional metadata provided by the driver.  Catalog authors can use whatever key names they would like, with the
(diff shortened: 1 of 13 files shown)


martindurant:
@jsignell , any chance you can look this over? Particularly, there are a number of possible combinations of arguments (in the spec), parameter blocks (lists of UserParameters) and instantiation-time inputs (the things in `e(..)`) in the tests. Does the behaviour make sense? Are there more obvious things we can call the components?

When finalised, I'll clean up the documentation to the new state of affairs.

( cc @seibert )

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+    s = e()
+    assert s.kwargs['arg1'] == 'oi'
+
+    del os.environ['INTAKE_TEST_VAR']
+
+    s = e()
+    assert s.kwargs['arg1'] == ''
+
+
+def test_up_averride():
```
danielballan: Is this test identical to the test after it but with a typo in its name?

martindurant: No, in this case the kwarg passed by the "user" matches the name of the UserParameter; in the following test, it matches one of the args (i.e., as may have been read from a cat definition `args:` block). This first version is the only usage which was enabled before this PR.

intake/catalog/local.py
```diff
@@ -91,10 +92,9 @@ def describe(self):
         desc = {
             'name': self.name,
             'description': self.description,
-            'type': self.type,
-            'default': self.default
+            'type': self.type or 'unknown',
```
jsignell: Does this allow the passing of arbitrary types?

intake/catalog/tests/test_local.py
```diff
@@ -60,7 +60,7 @@ def test_local_catalog(catalog1):
     assert catalog1['entry1'].get().container == 'dataframe'
     md = catalog1['entry1'].get().metadata
     md.pop('catalog_dir')
-    assert md == dict(foo='bar', bar=[1, 2, 3], cache=[])
+    assert md == dict(foo='bar', bar=[1, 2, 3])
```
jsignell: :+1:

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+
+def test_parameter_default():
+    up = UserParameter('name', default='oi')
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"},
+                          parameters=[up])
+    s = e()
+    assert s.kwargs['arg1'] == 'oi'
+
+
+def test_maybe_par_env():
```
jsignell: I don't understand this name

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+    up = UserParameter('name', default='env(INTAKE_TEST_VAR)')
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"},
+                          parameters=[up], getenv=False)
+    s = e()
+    assert s.kwargs['arg1'] == 'env(INTAKE_TEST_VAR)'
+
+    os.environ['INTAKE_TEST_VAR'] = 'oi'
+    up = UserParameter('name', default='env(INTAKE_TEST_VAR)')
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"},
+                          parameters=[up])
```
jsignell: Maybe explicitly set `getenv=True` to make it clear that that changed?
```suggestion
                          parameters=[up], getenv=True)
```

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def test_maybe_par_env():
+    up = UserParameter('name', default='env(INTAKE_TEST_VAR)')
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"},
+                          parameters=[up], getenv=False)
+    s = e()
+    assert s.kwargs['arg1'] == 'env(INTAKE_TEST_VAR)'
+
+    os.environ['INTAKE_TEST_VAR'] = 'oi'
```
jsignell: can you repeat lines 34 and 35 here?

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+    up = UserParameter('name', default='env(INTAKE_TEST_VAR)')
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"},
+                          parameters=[up], getenv=False)
+    s = e(name='other')
+    assert s.kwargs['arg1'] == 'other'
+
+
+def test_override():
+    up = UserParameter('name', default='env(INTAKE_TEST_VAR)')
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"},
```
jsignell: It might make this clearer if you didn't use a template for the arg1 value in the catalog entry:

```suggestion
    e = LocalCatalogEntry('', '', driver, args={'urlpath': "https://foo.fake.org"},
```

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+                          parameters=[up], getenv=False)
+    s = e(arg1='other')
```
jsignell:
```suggestion
    s = e(urlpath='https://a.different.site')
```

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+    assert s.kwargs['arg1'] == 'other'
```
jsignell:
```suggestion
    assert s.kwargs['urlpath'] == 'https://a.different.site'
```

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+    del os.environ['INTAKE_TEST_VAR']
+
+    s = e()
+    assert s.kwargs['arg1'] == ''
+
+    e = LocalCatalogEntry('', '', driver,
+                          args={'arg1': "{{env(INTAKE_TEST_VAR)}}"},
+                          parameters=[], getenv=False)
+    s = e()
+    assert s.kwargs['arg1'] == '{{env(INTAKE_TEST_VAR)}}'
```
jsignell: I'd like to see this tested with the env var set as well

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+                          parameters=[up], getenv=False)
+    e()  # OK
+    with pytest.raises(ValueError):
+        e(name='oi')
+
+    up = UserParameter('name', type='int')
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"},
+                          parameters=[up], getenv=False)
+    s = e()  # OK
+    assert s.kwargs['arg1'] == '0'  # default default for int
```
jsignell: The arg value isn't an int?

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+                          parameters=[], getenv=False)
+    s = e()
+    assert s.kwargs['arg1'] == '{{env(INTAKE_TEST_VAR)}}'
+
+
+def test_validate_up():
+    up = UserParameter('name', default=1, type='int')
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"},
+                          parameters=[up], getenv=False)
+    e()  # OK
```
jsignell: Check that the default is 1 right?

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+    assert s.kwargs['arg1'] == 1  # a number, not str
+
+
+def test_overrides():
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "oi"})
+    s = e(arg1='hi')
+    assert s.kwargs['arg1'] == 'hi'
+
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "{{name}}"})
+    s = e(name='hi')
```
jsignell: what happens if you set `name` **and** `arg1`?

intake/catalog/tests/test_parameters.py
```diff
@@ -0,0 +1,138 @@
...
+    assert s.kwargs['arg1'] == 'hi'
+
+    os.environ['INTAKE_TEST_VAR'] = 'another'
+    e = LocalCatalogEntry('', '', driver, args={'arg1': "oi"})
```
jsignell: explicitly set getenv to True

intake/catalog/utils.py
```diff
@@ -133,6 +174,39 @@ def expand_defaults(default, client=False, getenv=True, getshell=True):
     return default
 
 
+def merge_pars(params, user_inputs, spec_pars, client=False, getenv=True,
+               getshell=True):
```
jsignell: doc string?

intake/catalog/utils.py
```diff
@@ -133,6 +174,39 @@ def expand_defaults(default, client=False, getenv=True, getshell=True):
...
+    context = params.copy()
+    for par in spec_pars:
+        val = user_inputs.get(par.name, par.default)
+        if val is not None:
+            if isinstance(val, six.string_types):
```
jsignell: are we moving towards supporting python2?

danielballan: It would be nice to set a direction on this soon. I have been using `six` in my PRs also, since I see it used in various places in the codebase, but not bothering to actually test it since there is no CI for Python 2. https://github.com/intake/intake/issues/228

jsignell: Right. I have been using Python 3 features when they are convenient to me - but I'll stop if we plan to ever support Python 2 :)

intake/catalog/local.py
martindurant: Yes, that's the idea - if the type is not set, there is no coersion

intake/catalog/tests/test_parameters.py
martindurant: maybe fill in parameter default from the env, depending on getenv

martindurant: I can add exactly that text as a comment

jsignell: Sorry I meant to generalize this comment. I think you could stand to have longer test names that are more explicit.

intake/catalog/tests/test_parameters.py
martindurant: Isn't that what happens a few lines above?

intake/catalog/tests/test_parameters.py
martindurant: No! The value of "name" is an int, but it gets rendered into a string. The final value of "arg1" is not given.

intake/catalog/tests/test_parameters.py
martindurant: arg1 wins

intake/catalog/utils.py
martindurant: Yep, lots of detailed documentation to come...

intake/catalog/tests/test_parameters.py
jsignell: Ah I see. So you are showing that the default for name isn't used.

intake/catalog/tests/test_parameters.py
jsignell: ok yes I get it now. Maybe a comment?

intake/catalog/utils.py
martindurant: OK, final: no py2. The six things can migrate away, although they do no harm.

docs/source/catalog.rst
```diff
@@ -8,12 +8,14 @@ datasets, called *catalog entries*.  A catalog entry for a dataset includes info
...
-In addition, Intake allows datasets to be *parameterized* in the catalog.  This is most commonly used to allow the
-user to filter down datasets at load time, rather than having to bring everything into memory first.  The data
-parameters are defined by the catalog author, then templated into the arguments for ``__init__()`` to modify the data
-being loaded.  This approach is less flexible for the end user than something like the
-`Blaze expression system <https://blaze.readthedocs.io/en/latest/expr-compute-dev.html>`_, but also significantly
-reduces the implementation burden for driver authors.
+In addition, Intake allows the arguments to data sources to be templated, with the variables explicitly
+expressed as "user parameters". The given arguments are rendered using ``jinja2`` and the named user
+parameters, and validation is also provided for the allowed types and values for both the template
+values and the final arguments passed to the data source. The parameters are named and described, to
```
jsignell: Split up that sentence if possible.

docs/source/catalog.rst
```diff
@@ -8,12 +8,14 @@ datasets, called *catalog entries*.  A catalog entry for a dataset includes info
...
+indicate to the user what they are for.
+This kind of structure can be used to, for example,
+choose between two parts of a given data source, like "latest" and "stable".
```
jsignell: Maybe give a catalog/code example?

docs/source/catalog.rst
```diff
@@ -136,11 +131,88 @@ returned data.  Each data source has several attributes:
...
-or reduce the data in specific ways defined by the catalog author.  The parameters defined for a given data source are
-available for use in template strings, which can be used to alter the arguments provided to the driver.  For example,
-a data source might accept a "postal_code" argument which is used to alter a database query, or select a particular
-group within a file.  Users set parameters with keyword arguments to the ``get()`` method on the catalog object.
+
+Templating
+''''''''''
+
+Intake catalog files support Jinja2 templating for driver arguments. Any occurrence of
+a substring like ``{{field}}`` will be replaced by the value of the user parameters with
```
jsignell: Do we normally format that with spaces inside the double brackets?

docs/source/catalog.rst
```diff
@@ -136,11 +131,88 @@ returned data.  Each data source has several attributes:
...
+that same name, or explicitly provided by the user. For how to specify these user parameters,
```
jsignell:
```suggestion
that same name, or the value explicitly provided by the user. For how to specify these user parameters,
```

docs/source/catalog.rst
```diff
@@ -136,11 +131,88 @@ returned data.  Each data source has several attributes:
...
+see the next section.
+
+Some additional values are available for templating. The following is always available:
+
+- ``CATALOG_DIR``: The full path to the directory containing the YAML catalog file.  This is especially useful
+  for constructing paths relative to the catalog directory to locate data files and custom drivers.
+
+so that the search for CSV files for the two "entry1" blocks, above, will happen in the same directory as
```
jsignell: Missing some text?

docs/source/catalog.rst
```diff
@@ -136,11 +131,88 @@ returned data.  Each data source has several attributes:
...
+where the catalog file was found.
+
+The following functions `may` be available. Since these execute code, the user of a catalog may decide
+whether they trust those functions or not.
+
+- ``env("USER")``: look in the environment for the named variable
```
jsignell:
```suggestion
- ``env("USER")``: look in the set environment variables for the named variable
```

docs/source/catalog.rst
```diff
@@ -136,11 +131,88 @@ returned data.  Each data source has several attributes:
...
+- ``client_env("USER")``: exactly the same, except that when using a client-server topology, the
+  value will come from the environment of the client.
+- ``shell("get_login thisuser -t")``: execute the command, and use the output as the value. The
+  output will be trimmed of any trailing whitespace.
```
jsignell: and preceding whitespace?

docs/source/catalog.rst
```diff
@@ -136,11 +131,88 @@ returned data.  Each data source has several attributes:
...
+.. code-block:: yaml
+
+    parameters:
+      name:
+        description: human-readable text for what this parameter means
+        type: optional, one of bool, str, int, float, list[str], list[int], list[float], datetime
+        default: optional, value to assume if user does not override
+        allowed: optional, list of values that are OK, for validation
+        min: optional, minimum allowed, for validation
+        max: optional, maximum allowed, for validation
```
jsignell: This section is slightly confusing because it is formatted like an example but it isn't an example. Could the explanatory text be in comments?

docs/source/catalog.rst
```diff
@@ -136,11 +131,88 @@ returned data.  Each data source has several attributes:
...
+  any of the source arguments, it will be replaced by the value of the parameter. If the user provides
+  a value (e.g., ``source = cat.entry(name='something")``), that will be used, otherwise the default value. If
+  there is no user input or default, the empty value appropriate for type is used. The ``default`` field allows
+  for the same function expansion as listed for arguments, above.
+
+- *If* an argument with the same name as the parameter exists, its value, after any templating, will be
+  coerced to the given type of the parameter and validated against the allowed/max/min. It is therefore possible
+  to use the string templating system (e.g., to get a value from the environment), but pass the final value as,
+  for example, an integer. It makes no sense to provide a default for this case (the argument already has a value),
+  but will no raise an exception.
```
jsignell:
```suggestion
  but providing a default will not raise an exception.
```

martindurant:
Missed this question:

> Can we guess type from the default value?

I'm not sure there's a need - the allowed type from the user need not be the same as the default, *except* in the case that the catalogue author wished to constrain it so.

docs/source/catalog.rst
martindurant: It makes no difference to jinja - I think it's fine if people don't worry about spacing here.

docs/source/catalog.rst
martindurant: That is required to get it to belong to the bullet

jsignell: confusing comment - I meant:

```suggestion
  output will be trimmed of any trailing or preceding whitespace.
```

martindurant: OK, sorry, I'll put it in next time.

Issue #289, opened 2019-03-10
Unable to ignore items when loading a compressed file


gramhagen:
I created a yaml spec pointing to a compressed file online, however there are additional (e.g. readme) files included in the tarball that confuse the data loader, which as far as I can understand tries to load all the decompressed files as dataframes.

I've tested a small code change that modifies the cache spec to include a regex_filter parameter which can be used to filter the desired item(s) from the compressed file. This seems like a pretty common scenario to support. Let me know if there's already an easier way to do this, Otherwise my suggestion is to have something like this:

```
sources:
  test:
    driver: csv
    args:
      urlpath: test.tar.gz
    cache:
      - type: compressed
        argkey: urlpath
        regex_filter: '.*sample.txt'
```

gramhagen:
closing this as it is addressed by #290

PR #290, opened 2019-03-10
adding regex_filter to cache spec with test case

gramhagen:
Addressing issue #289 

To support compressed files with extraneous files that should be ignored I'm adding an additional parameter to the cache spec.

```
sources:
  test:
    driver: csv
    args:
      urlpath: test.tar.gz
    cache:
      - type: compressed
        argkey: urlpath
        regex_filter: '.*sample.csv'
```

this is used to filter out the file(s) that are desired, one thing that doesn't seem ideal here is that the prefix will contain the full path to the cached files which were extracted, so you have to include wildcards for the beginning of the filter.

i created a test tarball with the calvert_uk data plus a readme file and added a test of this functionality


martindurant:
This sounds very reasonable to me.
Am I right in thinking that all files are still extracted from the archive, but only the matching ones are returned to the driver for loading? That is a fine pattern (it can be useful to have any README locally), just wanted to be sure.

gramhagen:
Yes, the files are extracted from the archive into the cache folder, but they are not cpatured in the cache metadata or visible to the driver.

martindurant:
Cool, thanks for the clarification. Where do you think this parameter could be documented? Perhaps the docstring of `CompressedCache` should describe al the inputs.

gramhagen:
yeah, that sounds good, i can update that. also perhaps the documentation can be updated, though I haven't look at that much. Are there are specific files / sections that you think would be helpful to change?

martindurant:
The cache mechanism in general is mentioned in `catalog.rst`, under the YAML format section and says how to find the classes, so you can read their docstrings. The base cache class is in `api_base.rst`, but the actual implementations are not. It could have its own section there, perhaps.

gramhagen:
Ok, I updated catalog.rst and added info in the docstring for the compressed cache class

docs/source/catalog.rst
```diff
@@ -311,6 +311,33 @@ The "types" of caching are that supported are listed in ``intake.source.cache.re
 the docstrings of each for specific parameters that should appear in the cache block.
 
 
+Compressed Files
+----------------
```
martindurant: Sorry, I think this should be `'''''''''`, else it'll be a higher-level heading than intended

gramhagen: ok, no worries

martindurant:
That'll do nicely, thank you very much.

PR #299, opened 2019-03-18
Use yaml safe-load everywhere

### LINKED ISSUE #295: Quickstart "unsafe" warning
I just tried running the Intake QuickStart, and it throws a warning that, while isn't really anything to worry about, includes the words `default Loader is unsafe` which new users might find concerning:
 ```python
import intake
ds = intake.open_csv('states_*.csv')
print(ds)
```
produces the warning:
```
/opt/conda/lib/python3.6/site-packages/dask/config.py:168: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  data = yaml.load(f.read()) or {}
/opt/conda/lib/python3.6/site-packages/distributed/config.py:20: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  defaults = yaml.load(f)
```

martindurant:
Fixes #295 and errors with pyyaml 5.1

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -616,6 +616,9 @@ class RemoteCatalog(Catalog):
         for name in source:
             entry = source[name]
             out[name] = entry.__getstate__()
+            out[name]['parameters'] = [up._captured_init_kwargs for up
+                                       in entry._user_parameters]
+            out[name]['kwargs'].pop('parameters')
         fn = posixpath.join(path, 'cat.yaml')
         with open_files([fn], 'wt')[0] as f:
             yaml.dump({'sources': out}, f)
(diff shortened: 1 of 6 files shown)


dotlambda:
When can we expect a release with this fix? It prevents me from upgrading to PyYAML 5.1.

martindurant:
Good point. I'll do it when I am back home, so after Monday.

martindurant:
https://github.com/conda-forge/intake-feedstock/pull/8

dotlambda:
Thanks a lot!

PR #303, opened 2019-03-21
Tab style GUI

jsignell:
I am trying to capture some different options for how to style the GUI. 

Screen Shot 2019-03-21 at 2 41 01 PM
Screen Shot 2019-03-21 at 2 41 15 PM
Screen Shot 2019-03-21 at 2 41 26 PM

### DIFF
--- a/intake/gui/base.py
+++ b/intake/gui/base.py
@@ -8,7 +8,7 @@ import panel as pn
 from param.parameterized import Event
 
 MAX_WIDTH = 1200
-BACKGROUND = '#eeeeee'
+BACKGROUND = '#ffffff'
 
 
 class Base(object):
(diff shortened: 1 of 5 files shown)


jsignell:
This is more of the pipeline approach to things. Each tab has a discrete task, with an outcome that is used in the subsequent tab. 

This is the conversation that Martin and I have had so far about it:

> You don’t necessarily want to start on the add-cat tab, for instance; but we could give more information about the currently-selected cat (type, args…).

you could start on the middle tab and there could be an add cat button next to the remove cat button that hops you to the first tab

> btw: no need to display the details of the selected plot, it shows up in the graph labels etc. Eventually it will be [‘plot1’, ‘plot2’, ’custom…’] in the selector, where the last one gives your the full customisable UI. If you don’t display details, the button can be above the plot and left-justified.

I was imagining that the custom plot would be a separate tab - so the details could be a useful guide to get people started setting up their own plot.

> Also, if you put a border around the cat-list and remove button, the button doesn’t need more text than “remove”; furthermore, you could argue that the source details “belong” to the source list, and that the plot should have the cat and source name/description as a label above

martindurant:
Maybe close this for now, given our discussions.

jsignell:
Agree

PR #306, opened 2019-03-22
Get catalog names from several inputs

### LINKED ISSUE #305: Best practices for naming catalogs
I've been wondering what the best practices should be for naming catalogs. I've been leaning towards: 

 - a single catalog in a dir - catalog.yml 
 - more than one catalog in a dir - name them based on their contents

I like naming catalogs catalog.yml because it makes it clear what the role for the file is (it functions like an environment.yml). If that is recommended though - then we might want to make some attempt to get the name of the catalog from the dirname.

jsignell:
Fixes #305 

Target is to implement naming from:

 1. name given by open function `intake.open_catalog(fn, name='top_priority')`
 2. name given in the spec
 3. name given by the caller, e.g., the entry name of the super cat
 4. name of the file/directory or something else derived from the path

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -50,9 +50,9 @@ class Catalog(DataSource):
     container = 'catalog'
     name = 'catalog'
 
-    def __init__(self, *args, name=None, metadata=None, auth=None, ttl=1,
-                 getenv=True, getshell=True, persist_mode='default',
-                 storage_options=None):
+    def __init__(self, *args, name=None, description=None, metadata=None,
+                 auth=None, ttl=1, getenv=True, getshell=True,
+                 persist_mode='default', storage_options=None):
         """
         Parameters
         ----------
@@ -60,6 +60,9 @@ class Catalog(DataSource):
             A single URI or list of URIs.
         name : str, optional
             Unique identifier for catalog. This is primarily useful when
+            manually constructing a catalog. Defaults to None.
+        description : str, optional
+            Description of the catalog. This is primarily useful when
             manually constructing a catalog. Defaults to None.
         metadata: dict
             Additional information about this data
@@ -86,6 +89,7 @@ class Catalog(DataSource):
         """
         super(Catalog, self).__init__()
         self.name = name
+        self.description = description
         self.metadata = metadata or {}
         self.ttl = ttl
         self.getenv = getenv
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -59,11 +59,11 @@ class Catalog(DataSource):
         args : str or list(str)
             A single URI or list of URIs.
         name : str, optional
-            Unique identifier for catalog. This is primarily useful when
-            manually constructing a catalog. Defaults to None.
+            Unique identifier for catalog. This takes precedence over whatever
+            is stated in the cat file itself. Defaults to None.
         description : str, optional
-            Description of the catalog. This is primarily useful when
-            manually constructing a catalog. Defaults to None.
+            Description of the catalog. This takes precedence over whatever
+            is stated in the cat file itself. Defaults to None.
         metadata: dict
             Additional information about this data
         auth : BaseClientAuth or None
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -138,7 +138,9 @@ class LocalCatalogEntry(CatalogEntry):
         Parameters
         ----------
         name: str
-            How this entry is known, normally from its key in a YAML file
+            How this entry is known, normally from its key in a YAML file, or
+            if that is not provided then from name of file, or name of dir if
+            file name is 'catalog.yaml' or 'catalog.yml'.
         description: str
             Brief text about the target source
         driver: str
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -228,7 +228,7 @@ class LocalCatalogEntry(CatalogEntry):
 
         if len(self._plugin) == 0:
             raise ValueError('No plugins loaded for this entry: %s\n'
-                             'A listing of installable plugins can be found ' 
+                             'A listing of installable plugins can be found '
                              'at https://intake.readthedocs.io/en/latest/plugin'
                              '-directory.html .'
                              % self._driver)
@@ -491,7 +491,8 @@ class CatalogParser(object):
         return dict(
             plugin_sources=self._parse_plugins(data),
             data_sources=self._parse_data_sources(data),
-            metadata=data.get('metadata', {})
+            metadata=data.get('metadata', {}),
+            name=data.get('name'),
         )
 
 
@@ -552,8 +553,6 @@ class YAMLFileCatalog(Catalog):
             file_open = open_files(self.path, mode='rb', **options)
             assert len(file_open) == 1
             file_open = file_open[0]
-        self.name = os.path.splitext(os.path.basename(
-            self.path))[0].replace('.', '_')
         self._dir = get_dir(self.path)
 
         with file_open as f:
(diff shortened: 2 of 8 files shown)


martindurant:
Do you think that the cat file should also be able to supply its own description and metadata? The former would take precedence, and the latter use update() to merge with whatever can in the init method. Of course, this is a little more subtle, as description normally applies to an entry, not a source.  I'm wondering if we are losing the distinction or if that distinction is necessary.

jsignell:
> I'm wondering if we are losing the distinction or if that distinction is necessary.

I have never had that firm a grasp on the distinction. What is the benefit of having two separate classes?

martindurant:
The entry: holds the inputs from the spec (more or less *is* the spec), including parameters and templates that might be replace/validated. Requires absolutely no work to instantiate.

The source: holds one specific set of inputs, no user options. Instantiation may involve some work; querying schema may require some (constant time) work, has methods for read, etc.

However, metadata is generally just passed though. In some cases (e.g., xarrays), some metadata can come from the source itself.

jsignell:
Right - I think the distinction is still meaningful - especially in the context of user parameters where a particular entry might be instantiated into any number of sources. So if the distinction is useful - maybe we should think about whether there is a better way to make the distinction more apparent to users.

jsignell:
> Do you think that the cat file should also be able to supply its own description and metadata? 

You mean in the `open_catalog` call or in the yaml (isn't it already possible in the yaml?)

martindurant:
I meant the latter, but both would be good, analogous to the name.

jsignell:
That sounds reasonable. Want me to do it in this PR - or separate?

martindurant:
May as well fold it in, seems like very similar territory.
I would suggest that the built-in catalog should be called "Builtin" and have description "Generated from data packages found on your intake search path", or something like that.

jsignell:
I added the description and builtin name - wasn't quite sure I understood what was needed on the metadata. Metadata can already be passed in a catalog spec.

martindurant:
> Metadata can already be passed in a catalog spec.

Does it end up in `cat.metadata`?

jsignell:
> Does it end up in cat.metadata

Yep. 

I think it would be unclear to allow passing metadata in the open_catalog method - especially if we use it to do an update rather than to replace what's in the spec.

intake/catalog/base.py
```diff
@@ -61,6 +61,9 @@ def __init__(self, *args, name=None, metadata=None, auth=None, ttl=1,
         name : str, optional
             Unique identifier for catalog. This is primarily useful when
             manually constructing a catalog. Defaults to None.
+        description : str, optional
+            Description of the catalog. This is primarily useful when
```
martindurant: "This is primarily..." - I don't think this is required. Only need say that this takes precedence over whatever is stated in the cat file itself.

intake/catalog/local.py
```diff
@@ -491,7 +491,9 @@ def _parse(self, data):
         return dict(
             plugin_sources=self._parse_plugins(data),
             data_sources=self._parse_data_sources(data),
-            metadata=data.get('metadata', {})
+            metadata=data.get('metadata', {}),
+            name=data.get('name'),
```
martindurant: These two allowed to be missing?

intake/catalog/local.py
```diff
@@ -584,7 +584,15 @@ def _load(self):
             self._entries[entry.name] = entry
 
         self.metadata = cfg.get('metadata', {})
+        self.name = self.name or cfg.get('name') or self.name_from_path
```
martindurant: `self.name` starts as `None`?

intake/catalog/local.py
```diff
@@ -584,7 +584,15 @@ def _load(self):
...
+        self.description = self.description or cfg.get('description')
 
+    @property
+    def name_from_path(self):
+        name = os.path.splitext(os.path.basename(self.path))[0]
+        if name == 'catalog':
```
martindurant: for "catalog.yaml"? Maybe this should be documented somewhere. It could also be an option to always take the dir name.

intake/catalog/local.py
```diff
@@ -623,14 +631,16 @@ def _load(self):
...
+            self.description = self.description or f'Catalog generated from {len(files)} files'
             self.path = [make_path_posix(p) for p in self.path]
         else:
             if isinstance(self.path, str) and '*' not in self.path:
                 self.path = self.path + '/*'
             files = open_files(self.path, mode='rb', **options)
             self.path = make_path_posix(self.path)
-            self.name = self.path
+            self.name = self.name or self.path
+            self.description = self.description or f'Catalog generated from all files found in {self.path}'
```
martindurant: *here* perhaps we can just have `self.path` and be done. In the repr, you get the word catalog anyway.

intake/catalog/tests/test_local.py
```diff
@@ -79,6 +79,13 @@ def test_nested(catalog1):
     assert catalog1.nested.nested.nested().cat.cat.cat is catalog1
 
 
+def test_nested_gets_name_from_super(catalog1):
+    assert catalog1.name == 'name_in_cat'
+    assert 'nested' in catalog1
+    nested = catalog1.nested
+    assert nested.name == 'nested'
```
martindurant: Shouldn't the name in the spec trump the name of the entry? What is `nested().name` ?

intake/gui/tests/test_gui.py
```diff
@@ -15,5 +15,5 @@ def test_add_cat():
     import intake
     intake.gui.add_cat(os.path.join(here, '..', '..', 'catalog', 'tests',
                                     'catalog1.yml'))
-    assert 'catalog1' in intake.gui.cat_list.value
+    assert 'name_in_cat' in intake.gui.cat_list.value
```
martindurant: How are the names generated for the GUI exactly - they are the names of the *sources*, I suppose? This has some merit, in that when you select a catalog in the list of entries, you expect something with that name to pop up in the cat list; still, `list(cat)` will now not necessarily align with the contents of the source list.

(I forgot to ask you in the GUI PR whether you can select catalog entries in the source list and expect them to load; also, cats were visually marked in the jupyter-widgets version as being special, different from other entries)

intake/catalog/local.py
jsignell: yes.

intake/catalog/tests/test_local.py
jsignell: I guess that is what I have written at the top of this PR, but it wasn't clear to me which should win out of 2 or 3. Is it weird to not have the name match the key?

intake/catalog/local.py
jsignell: yes

intake/catalog/local.py
jsignell: Yeah for catalog.yaml or catalog.yml. I'll make a docstring and think about where to add it in the docs.

intake/catalog/local.py
martindurant: Actually, changed my mind - forget it.

intake/catalog/tests/test_local.py
jsignell: I think entry name and source name need to be the same: `nested().name == 'nested'`

jsignell: Do you have a preference?

intake/gui/tests/test_gui.py
jsignell: I think this should be discussed in the GUI PR. I haven't thought very much about nested catalogs.

martindurant:
Is there anything left to be done here?

intake/catalog/tests/test_local.py
jsignell: What about this one?

intake/catalog/local.py
jsignell: @martindurant where to add in the docs?

jsignell:
I pinged you on a couple that I think are still up for discussion. Most important is adding some docs

intake/catalog/tests/test_local.py
martindurant: I would go with what you suggest. Perhaps the docstring for YAMLFile(s)Catalog can say that the cat object gets a generated name *if* it wasn't already supplied.

intake/catalog/local.py
martindurant: In class docstring, since this is basically something that is decided by the inputs to `__init__`

martindurant:
Merged this now, so it can be included in the release later today.

PR #309, opened 2019-03-26
Pipe source dir through to entry

martindurant:
For discussion on whether this is a good idea. I am minded to keep it.

Needs tests.


jsignell:
Sorry I forget to comment on the issue - but yes this looks like a reasonable idea to me.

danielballan:
Agreed.

PR #317, opened 2019-04-02
Add items() function to catalog

ian-r-rose:
Between the `__getitem__` access and the `__iter__` implementation, catalogs feel very dictionary-esque (not to mention the internal implementation). I found myself wanting an easier way to iterate over the entries of the catalog in a similar way to dictionaries. Would there be interest in this functionality?

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -191,6 +191,10 @@ class Catalog(DataSource):
             out[n] = item
         return out
 
+    def items(self):
+        """Get an iterator over (key, value) tuples for the catalog entries."""
+        return self._get_entries().items()
+
     @reload_on_change
     def _get_entry(self, name):
         entry = self._entries[name]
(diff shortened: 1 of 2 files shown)


intake/catalog/base.py
```diff
@@ -191,6 +191,10 @@ def walk(self, sofar=None, prefix=None, depth=2):
             out[n] = item
         return out
 
+    def items(self):
+        """Get an iterator over (key, value) tuples for the catalog entries."""
+        return self._get_entries().items()
```
martindurant: @danielballan , might this be a problem for catalogs where the entries are not a simple dict?

danielballan: Thanks for thinking of that, but this should be OK. The dict-like `Entries` class used by `RemoteCatalog` implements `items()`, as do the special dict-like classes I wrote for intake-bluesky. In fact I use `cat._get_entries().items()` a couple places in intake-bluesky as a workaround and was considering asking for `cat.items()` to be added. Thanks @ian-r-rose!

Added https://github.com/intake/intake/pull/318 to document which parts of the dict interfaces intake expects custom entries containers to implement.

martindurant:
Thanks for this

PR #318, opened 2019-04-03
DOC: Document what _make_entries_container should return.

danielballan:
This documents which parts of the ``dict`` interface we require and why.

Also, I have since learned that inheriting from ``dict`` is still a fraught.
Exactly which methods call which other methods internally changed on a minor
release of Python (see https://github.com/matplotlib/matplotlib/issues/12601).
It's safer for us to implement exactly the methods use and let an
``AttributeError`` be raised if some part of the code starts to use a method we
don't implement, rather than return a wrong value.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -355,7 +355,7 @@ class Entries(collections.abc.Mapping):
                 return source
 
     def __len__(self):
-        return len(self.catalog)
+        return len(self._catalog)
 
 
 class RemoteCatalog(Catalog):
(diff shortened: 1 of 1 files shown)


martindurant:
> inheriting from dict is still a fraught

Is collections.abc.(Mutible)Mapping better?

danielballan:
I'm not aware of any problems with that one. Let me take a look....

danielballan:
Good suggestion. I think we want `Mapping` not `MutableMapping` --- the later would require us to implement `__setitem__` and `__delitem__` which do not apply here.

But, yes, this is a clear improvement. The result has the full (readonly) `dict` interface and is simpler to look at.

PR #320, opened 2019-04-05
Give BasicClientAuth a stable dask token.

danielballan:
Catalogs should have a stable `__hash__`. This means they should return a stable
result when

```py
dask.base.tokenize(catalog.__getstate__())
```

is called. In my Catalog subclasses, ``self.auth`` (a ``BaseClientAuth`` instance)
is among the contents of ``__getstate__()``. Dask falls back to its last resort
for tokenization, which is UUID4 and thus not stable.

Implementing `__dask_tokenize__` on `BaseClientAuth` seems the best place to fix
this so that it does not trip up others. Of course I could implement it on my
`Catalog` subclasses instead or as well.

### DIFF
--- a/intake/auth/base.py
+++ b/intake/auth/base.py
@@ -66,7 +66,15 @@ class BaseClientAuth(object):
         self.args = args
 
     def __dask_tokenize__(self):
-        return self.args
+        return hash(self)
+
+    @property
+    def _tok(self):
+        from dask.base import tokenize
+        return tokenize({'cls': type(self).__name__, 'args': self.args})
+
+    def __hash__(self):
+        return int(self._tok, 16)
 
     def get_headers(self):
         """Returns a dictionary of HTTP headers for the remote catalog request.
--- a/intake/auth/base.py
+++ b/intake/auth/base.py
@@ -65,6 +65,9 @@ class BaseClientAuth(object):
     def __init__(self, *args):
         self.args = args
 
+    def __dask_tokenize__(self):
+        return self.args
+
     def get_headers(self):
         """Returns a dictionary of HTTP headers for the remote catalog request.
         """
--- a/intake/auth/secret.py
+++ b/intake/auth/secret.py
@@ -60,6 +60,7 @@ class SecretClientAuth(BaseClientAuth):
     def __init__(self, secret, key='intake-secret'):
         self.secret = secret
         self.key = key
+        super(SecretClientAuth, self).__init__()
 
     def get_headers(self):
         return {self.key: self.secret}


martindurant:
Catalogs do implement hash, which should depend only only the startup arguments to the catalog class and the class itself; also seen in hex form in the `._tok` attribute.

danielballan:
I agree, but what if the start arguments to the catalog class include an instance of `BaseClientAuth`?

https://github.com/NSLS-II/intake-bluesky/blob/3e92aea11acded14074041b7da3104f4af702463/intake_bluesky/jsonl.py#L249-L257

martindurant:
OK, see what you mean; does it also work to have BaseAuth have the same _tok and hash methods like Catalog, then, rather than this dask-specific solution? Or perhaps both?

danielballan:
The problem is dask-specific so we do need `__dask_tokenize__`.

On the `master` branch, `BaseClientAuth.__hash__` is stable but `dask.tokenize` is not:

```py
In [3]: auth = intake.auth.base.BaseClientAuth('a', 1)

In [4]: hash(auth)
Out[4]: -9223363294960595733

In [5]: hash(auth)
Out[5]: -9223363294960595733

In [6]: dask.base.tokenize(auth)
Out[6]: '0116bd9e4883d7c3c31ad8437d659ce3'

In [7]: dask.base.tokenize(auth)
Out[7]: 'bf771e8380cdace324b4269fff76c922'
```

Defining `__hash__` and `_tok` like

```diff
+
+    @property
+    def _tok(self):
+        from dask.base import tokenize
+        return tokenize({'cls': type(self).__name__, 'args': self.args})
+
+    def __hash__(self):
+        return int(self._tok, 16)
```

doesn't hurt, but it doesn't help:

```py
In [4]: hash(auth)
Out[4]: 157894537581662214

In [5]: hash(auth)
Out[5]: 157894537581662214

In [6]: dask.base.tokenize(auth)
Out[6]: 'c787a9b0fe7a4cee42fa000d91885e80'

In [7]: dask.base.tokenize(auth)
Out[7]: 'e4d70f05ca7ad66533feec97c01305a4'
```

because `dask` dispatches is tokenization on _type_ and it still doesn't recognize the type `BaseClientAuth` so we end up [in this codepath](https://github.com/dask/dask/blob/5dc48d28d9c96c014789c8313a2f27f1560819a1/dask/base.py#L637-L642).

Finally, with `__dask_tokenize__`:

```py
In [4]: dask.base.tokenize(auth)
Out[4]: '02800d3fd34f82ac21875e0211754009'

In [5]: dask.base.tokenize(auth)
Out[5]: '02800d3fd34f82ac21875e0211754009'
```

Should I push a commit adding `__hash__` and `_tok` for completeness?

martindurant:
Should have been `super(BaseClientAuth, self).__init__()` - but otherwise I agree and this is good.

danielballan:
Ah, thanks. Fix force-pushed.

PR #321, opened 2019-04-06
More flexibility with dotted nested entry names

### LINKED ISSUE #316: nested catalogs and '.' characters in identifiers
In Pangeo, we are now using highly nested catalogs. We have hit an issue (https://github.com/pangeo-data/pangeo-datastore/pull/16#issuecomment-479124582), that exposes my ignorance about how `__getitem__` works with nested catalogs.

Consider the following:
```python
import intake
url = 'https://raw.githubusercontent.com/pangeo-data/pangeo-datastore/387e55f27c9a7c93975bdf6dfef58dcfb1077a2b/intake-catalogs/cmip6.yaml'
cat = intake.Catalog(url)
list(cat)
```

There are two entries (`['Omon', 'A3hr']`), which are themselves catalogs. 

`cat.walk()` gives
```
{'Omon.BCC-CSM2-MR.historical.r1i1p1f1.sos': <Catalog Entry: BCC-CSM2-MR.historical.r1i1p1f1.sos>,
 'Omon.BCC-CSM2-MR.historical.r1i1p1f1.tos': <Catalog Entry: BCC-CSM2-MR.historical.r1i1p1f1.tos>,
...
 'Omon': <Catalog Entry: Omon>,
 'A3hr.BCC-CSM2-MR.historical.r1i1p1f1.pr': <Catalog Entry: BCC-CSM2-MR.historical.r1i1p1f1.pr>,
 'A3hr.CNRM-CM6-1.historical.r1i1p1f2.huss': <Catalog Entry: CNRM-CM6-1.historical.r1i1p1f2.huss>,
...
 'A3hr': <Catalog Entry: A3hr>}
```

However, `cat['Omon.BCC-CSM2-MR.historical.r1i1p1f1.sos']  is None`.

Three things stand out:
- If there are no `.` characters in the dataset name, it works
- I would expect the keys returned by `walk` would be valid dataset identifiers 
- I would expect an error if I do not use a valid identifier, rather than returning `None`. For example, `cat['foo']` raises a `KeyError` but `cat['foo.bar']` returns `None`.

martindurant:
Fixes #316 

@naomi-henderson , @rabernat , please see the list of assertions in the new test and the included catalog file, to see if this now works in the way that you would like to see.

Note that, as with pandas, this complex dotted-name usage will work with getitem style, but not getattr, because of python syntax mangling. It would be possible, and already was, to mix `cat.thing['sub.thing'].more`, but that seems like poor style.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -249,14 +249,27 @@ class Catalog(DataSource):
             return e
         if isinstance(key, str) and '.' in key:
             key = key.split('.')
-        if isinstance(key, (tuple, list)):
-            if len(key) > 1 and key[0] in self._entries:
-                out = self._entries[key[0]].__getitem__(key[1:])
-                return out
-            if len(key) == 1:
-                return self[key[0]]
-        else:
-            raise KeyError(key)
+        if isinstance(key, list):
+            parts = list(key)[:]
+            prefix = ''
+            while parts:
+                bit = parts.pop(0)
+                prefix = prefix + ('.' if prefix else '') + bit
+                if prefix in self._entries:
+                    rest = '.'.join(parts)
+                    try:
+                        out = self._entries[prefix][rest]
+                        return out
+                    except KeyError:
+                        # name conflict like "thing" and "think.oi", where it's
+                        # the latter we are after
+                        continue
+        elif isinstance(key, tuple):
+            out = self
+            for part in key:
+                out = self[part]
+            return out
+        raise KeyError(key)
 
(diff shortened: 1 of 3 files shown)


martindurant:
Actually, it occurs to me that that attribute access would be possible, but maybe not when there is a name conflict, such as 'thing' and 'thing.thing' both being names in a catalog.

martindurant:
I'll merge this, as I think it's useful as it stands, and maybe attempt the attribute thing later.

PR #322, opened 2019-04-10
revert odd change to required python version

martindurant:
exclude py2 again, and even py35?


martindurant:
@danielballan

danielballan:
Context: https://github.com/intake/intake/issues/228#issuecomment-481838981

Issue #242, opened 2019-01-28
Dynamic catalogs


jacobtomlinson:
I'm currently doing some work with a data API which provides multiple CSV datasets. I ended up writing a quick notebook which gets the file manifest from the API, converts the format into the intake catalog style and outputs a yaml file. We can then use the intake catalog normally.

This API is adding new datasets all the time and I would like to keep my catalog up to date. I could set up a cron job somewhere to periodically generate the manifest. However I was thinking perhaps I could go a step further and write an intake plugin which adds the sources on import.

The API I'm working with is a pretty standard one, so I could envisage writing some kind of manifest which points to the specific implementation of the API. But it would then get the manifest, generate the catalog and add all the datasets to the built in catalog.

Could anyone provide me with some guidance on how I could implement this?

martindurant:
This is very similar to the example workflow in https://github.com/NCAR/intake-siphon/pull/2 , and elsewhere, where a query is made when the cat is instantiated, and then catalogue entries are generated from that. Dynamic catalogues are very important to Intake!

jacobtomlinson:
Ah awesome! I'll take a look at that example.

martindurant:
@jacobtomlinson , how are things going with catalogs and your various plugins? Are you still developing, do you need any pointers?

jacobtomlinson:
Things are progressing, although we are focusing on the data sets themselves and zarr at the moment.

ian-r-rose:
I am also working on a similar idea to what @jacobtomlinson describes over at [intake-dcat](https://github.com/CityOfLosAngeles/intake-dcat), and have run into some similar issues. I would like to be able to read in a remote catalog, convert it to something intake-friendly, and then re-export that intake catalog for use. Some specific ideas that I think might be useful to have on the base `Catalog` class:
* I was a bit surprised to not be able to find a function on the catalog to export itself to a `.yaml` file. Individual catalog entries can do that, so I can accomplish this by iterating over the entries, calling `yaml()` on them, and assembling those into a new catalog, but I think it would be nice to have a convenience method to do this automatically. I haven't really thought through how this would work for nested or remote catalogs, so perhaps there are some complicating considerations that would need to be addressed.
* I like the `search()` functionality for producing new catalogs. In a similar vein, I think users will want to produce new catalogs in a more targeted fashion than `search()`. I think it would be useful to implement some more general-purpose functional-flavored operations to the base `Catalog`. Something like
```
[...]

PR #323, opened 2019-04-12
Add cat methods

martindurant:
Fixes https://github.com/intake/intake/issues/242

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -131,8 +131,10 @@ class Catalog(DataSource):
         -------
         Catalog instance
         """
+        from dask.base import tokenize
         cat = cls(**kwargs)
         cat._entries = entries
+        cat._tok = tokenize(kwargs, entries)
         return cat
 
     @property
@@ -184,24 +186,26 @@ class Catalog(DataSource):
 
     @reload_on_change
     def search(self, text, depth=2):
+        import copy
         words = text.lower().split()
-        cat = Catalog(name=self.name + "_search",
-                      ttl=self.ttl,
-                      getenv=self.getenv,
-                      getshell=self.getshell,
-                      auth=self.auth,
-                      metadata=(self.metadata or {}).copy(),
-                      storage_options=self.storage_options)
+        entries = {k: copy.copy(v)for k, v in self.walk(depth=depth).items()
+                   if any(word in str(v.describe().values()).lower()
+                   for word in words)}
+        cat = Catalog.from_dict(
+            entries, name=self.name + "_search",
+            ttl=self.ttl,
+            getenv=self.getenv,
+            getshell=self.getshell,
+            auth=self.auth,
+            metadata=(self.metadata or {}).copy(),
+            storage_options=self.storage_options)
         cat.metadata['search'] = {'text': text, 'upstream': self.name}
-        cat._entries = {k: v for k, v in self.walk(depth=depth).items()
-                        if any(word in str(v.describe().values()).lower()
-                               for word in words)}
         cat.cat = self
-        for e in cat._entries.values():
+        for e in entries.values():
             e._catalog = cat
         return cat
 
-    def filter(self, predicate):
+    def filter(self, func):
         """Create a Catalog of a subset of entries based on a condition
 
         Note that, whatever specific class this is performed on, the return
@@ -211,7 +215,7 @@ class Catalog(DataSource):
 
         Parameters
         ----------
-        predicate : function
+        func : function
             This should take a CatalogEntry and return True or False. Those
             items returning True will be included in the new Catalog, with the
             same entry names
@@ -220,8 +224,8 @@ class Catalog(DataSource):
         -------
         New Catalog
         """
-        return Catalog.from_dict({name: entry for name, entry in self.items()
-                                  if predicate(entry)})
+        return Catalog.from_dict({key: entry for key, entry in self.items()
+                                  if func(entry)})
 
     @reload_on_change
     def walk(self, sofar=None, prefix=None, depth=2):
@@ -331,7 +335,7 @@ class Catalog(DataSource):
                 raise AttributeError(item)
         raise AttributeError(item)
 
-    def __setitem__(self, name, entry):
+    def __setitem__(self, key, entry):
         """Add entry to catalog
 
         This relies on the `_entries` attribute being mutable, which it normally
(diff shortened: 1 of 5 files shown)


ian-r-rose:
Nice, thanks for the quick turnaround!

martindurant:
I am posting now just to get your thoughts, @ian-r-rose , on whether this API looks OK. Tests are still needed, of course.

I would also like more dict-like actions such as to add and remove sources by key. Perhaps the YAMLFleCat should also have a "parse" method separate from reload (or classmethod `.from_string`?), with which you supply explicit YAML string. 

@danielballan , any thoughts on interacting with paginated cats?

ian-r-rose:
The serialization commands look good to me. I don't really have any special insight into how well they would work for paginated or remote catalogs, but as long as `items` works, it seems fine.

To clarify some of my thinking about creating new catalogs in a functional-ish way, this would be an axample of filtering a catalog, filter function vs comprehension:
```python
newCat = cat.filter(lambda key, entry: entry['metadata']['org'] == 'myorg')
newCat = Catalog.from_dict(
    {key: entry for for key,entry in cat.items() if entry['metadata']['org'] == 'myorg' }
)
```
The former is a bit terser, the latter is a bit more idiomatic.
One thing I like a lot about what you have proposed here (if I understand it correctly) is that it will work well for subclasses of `Catalog`. That is, subclasses that use `from_dict` will produce the right type of catalog.

I would be in favor of more dict-like actions on the catalog. I am less interested in `from_string` methods, even just scoped to the `YAMLFileCatalog`. Not against it really, but I don't think it's particularly onerous to ask for dict-like entries, and be more agnostic to the specific serlialization format (yaml, json, etc).

martindurant:
>  it will work well for subclasses of Catalog

indeed, that is true - and possibly problematic, as you could then end up with a SQLCatalog whose entries are not actually SQL sources. With the filter method, this would not happen, you get out whichever cat type you started with. Perhaps the class method is the right way to go, but with also `.filter()` provided to use the first form of syntax to execute the second?

ian-r-rose:
Good point -- sometimes `from_dict` approach is appropriate, but sometimes not. I would have no problem with providing both. Having a `filter` method is conceptually similar to the `search` method that already exists.

martindurant:
@jsignell , you now have a hook for setting the `.text` of a YAMLFileCatalog directly, and calling `parse()` - so editing the source. You can also call `cat._load(reload=True)` to revert.

martindurant:
Fixes #324

intake/catalog/base.py
```diff
@@ -259,6 +331,36 @@ def __getattr__(self, item):
...
+    def __delitem__(self, name):
+        """Remove entry from catalog
+
+        This relies on the `_entries` attribute being mutable, which it normally
+        is. Note that if a catalog automatically reloads, any entry removed here
+        may soon reappear
+
+        Parameters
+        ----------
+        name : str
```
jsignell: I think we need to work hard to differentiate between key and name. Also seems like it would be safer to use the value rather than the name for this also I mildly prefer `remove`. Maybe `pop` could be used for key?

intake/catalog/entry.py
```diff
@@ -125,14 +125,24 @@ def plots(self):
 
     def __getattr__(self, attr):
         # TODO: only consider attr not starting with "_"?
```
jsignell: Is this still a TODO?

intake/catalog/tests/test_local.py
```diff
@@ -594,3 +594,49 @@ def test_dot_names():
...
+    from intake.catalog.base import Catalog
+    entry1 = LocalCatalogEntry(name='trial', description='get this back',
+                              driver='csv')
+    entry2 = LocalCatalogEntry(name='trial', description='pass this through',
+                              driver='csv')
+    cat = Catalog.from_dict({'trial1': entry1,
+                             'trial2': entry2}, name='mycat')
+    cat2 = cat.filter(lambda e: 'pass' in e._description)
+    assert list(cat2) == ['trial2']
+    assert cat2.trial2 is entry2
```
jsignell: I guess this is why we wouldn't want to change the name.

intake/catalog/local.py
```diff
@@ -534,37 +534,44 @@ class YAMLFileCatalog(Catalog):
...
         Parameters
         ----------
         path: str
             Location of the file to parse (can be remote)
+        reload : bool
+            Whether to watch the source file for changes; make False if you want
+            an editable Catalog
         """
         self.path = path
+        self.autoreload = autoreload  # set this to False to prevent
```
jsignell: Is this comment complete?

intake/catalog/local.py
```diff
@@ -534,37 +534,44 @@ class YAMLFileCatalog(Catalog):
...
-        with file_open as f:
-            text = f.read().decode()
-        if "!template " in text:
-            logger.warning("Use of '!template' deprecated - fixing")
-            text = text.replace('!template ', '')
-
-        data = yaml_load(text)
+    def _load(self, reload=False):
+        if self.autoreload or reload:
+            # First, we load from YAML, failing if syntax errors are found
```
jsignell: I think this should be a docstring on the method since you've split the method. Same with the one in `parse`

intake/catalog/local.py
```diff
@@ -534,37 +534,44 @@ class YAMLFileCatalog(Catalog):
...
+
+            with file_open as f:
+                self.text = f.read().decode()
+            if "!template " in self.text:
+                logger.warning("Use of '!template' deprecated - fixing")
+                self.text = self.text.replace('!template ', '')
+            self.parse()
+
+    def parse(self):
+        data = yaml_load(self.text)
```
jsignell: I think it would be cleaner to pass `text` into `parse` as an arg or name it `_parse` otherwise it looks like a public method but it depends on other things having happened first.

intake/catalog/base.py
```diff
@@ -114,6 +114,27 @@ def __init__(self, *args, name=None, description=None, metadata=None,
...
+            e.g., is derived from ``collections.abc.Mapping``.
+        kwargs : passed on the constructor
+            Things like metadata, name; see ``__init__``.
+
+        Returns
+        -------
+        Catalog instance
+        """
+        cat = cls(**kwargs)
+        cat._entries = entries
```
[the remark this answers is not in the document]
martindurant: The entries have `_name` and `_catalog` attributes which the filter function can make use of, it it desires. Here, entries get whatever key was specified in the input entries dict-like. Are you saying you think we should reassign `_name` to them depending on the keys they might happen to have now?

intake/catalog/tests/test_local.py
martindurant: Yeah, I don't think we want. If the user does Catalog.from_dict({key: item}) where the item's name doesn't match they key, we should assume they know what they are doing.

martindurant:
@jsignell , 0c46837 should fix the search issue we were talking about - please have a careful look at it.

jsignell:
The change looks right to me. I imagine a test might be useful.

PR #325, opened 2019-04-18
Reorganizing GUI

jsignell:
Reorganizing GUI around the concept of a separation between catalog actions and sources actions after discussion with @jbednar and @jlstevens

Screen Shot 2019-04-18 at 4 57 04 PM

PR #327, opened 2019-04-24
Better repr

### LINKED ISSUE #311: Improve reprs for catalogs and catalog entries
Currently the reprs for catalogs and catalog entries are pretty spartan, which makes discovery and exploration more challenging. For instance, a typical entry repr looks something like
```
<Catalog Entry: some-url...>
```

Is there any interest in outputting more information here, at least in an IPython/Jupyter environment?

Catalogs could include:
* name
* length (at least for local catalogs)
* version

Entries could include
* markdown formatted description
* name
* intake driver
* target container

I see there is some effort towards a GUI for catalog exploration. I would view this as a lighter-weight, complementary feature.

### LINKED ISSUE #326: Performance issues with displaying catalog entries
I'm a bit puzzled as to what is going on here, but the repr for catalog entries is extremely slow.
If I have a catalog (in an IPython context), the following takes almost a second on my machine per entry:
```python
for entry_id, entry in catalog.items():
    display(entry)
```
The following is much faster:
```python
for entry_id, entry in catalog.items():
    display(entry.describe_open())
```

~~Some rooting around with cprofile suggests that the issue may be in `intake.catalog.utils._expand`, where there is the possibility for quadratic expansion of variables in templates. I'm not familiar enough with the codebase to know what might cause the paths to be so different for these two display variants, however~~.

Edit: seems like the problem may be with some unwanted interactions with IPython. It does a fair amount of attribute testing when deciding how to display a given object. The intake catalog entries overload `__getattr__`, and create new sources on the fly. This may turn what is meant to be a cheap attribute test into a much more expensive one with lots of jinja2 templating:

https://github.com/intake/intake/blob/b50ab5ca58e3c99eed159dc124db3862bf4ab03e/intake/catalog/entry.py#L128-L132

ian-r-rose:
Fixes #326, fixes #311.

This adds an `_ipython_display_` hook to catalog entries to better display them in a rich context. It sends over two mimetypes:
1. `text/plain` which should be viewable by all frontends, and uses the `pretty_describe` utility function,
2. `application/json`, which should be renderable with more modern frontends (JupyterLab, nteract), and allows for a collapsible and filterable view of the data:

image

We could also consider rendering something nicely with html or markdown, but I thought this was okay for the time being. The collapsible JSON rendering also allows for the flexibility of showing arbitrary metadata.

### DIFF
--- a/intake/catalog/entry.py
+++ b/intake/catalog/entry.py
@@ -28,7 +28,8 @@ class CatalogEntry(DictSerialiseMixin):
         """Get a dictionary of attributes of this entry.
 
         Returns: dict with keys
-
+          name: str
+              The name of the catalog entry.
           container : str
               kind of container used by this data source
           description : str
@@ -133,7 +134,7 @@ class CatalogEntry(DictSerialiseMixin):
             'application/json': contents,
             'text/plain': pretty_describe(contents)
         }, metadata={
-            'application/json': { 'root': self._name }
+            'application/json': { 'root': contents["name"]}
         }, raw=True)
         if warning:
             display(warning) # noqa: F821
@@ -145,7 +146,7 @@ class CatalogEntry(DictSerialiseMixin):
         try:
             contents.update(self.describe_open())
         except ValueError:
-            warning = f'Need additional plugin to use {self._driver} driver'
+            warning = f'Need an additional plugin to load {contents["name"]}.'
         return contents, warning
 
     def __getattr__(self, attr):
--- a/intake/catalog/entry.py
+++ b/intake/catalog/entry.py
@@ -5,8 +5,9 @@
 # The full license is in the LICENSE file, distributed with this software.
 #-----------------------------------------------------------------------------
 
+from copy import deepcopy
 import time
-from ..utils import DictSerialiseMixin
+from ..utils import DictSerialiseMixin, pretty_describe
 from .gui import EntryGUI
 
 
@@ -124,6 +125,32 @@ class CatalogEntry(DictSerialiseMixin):
     def plots(self):
         """List custom associated quick-plots """
         return list(self._metadata.get('plots', {}))
+
+    def _ipython_display_(self):
+        """Display the entry as a rich object in an IPython session."""
+        contents, warning = self._display_content()
+        display({
+            'application/json': contents,
+            'text/plain': pretty_describe(contents)
+        }, metadata={
+            'application/json': { 'root': self.name }
+        }, raw=True)
+        if warning:
+            display(warning)
+
+    def _display_content(self):
+        """Create a dictionary with content to display in reprs."""
+        contents = deepcopy(self.describe())
+        warning = None
+        try:
+            contents.update(self.describe_open())
+            if 'plots' in contents['metadata']:
+                contents['metadata'].pop('plots')
+            if 'plots' in contents['args'].get('metadata', {}):
+                contents['args']['metadata'].pop('plots')
+        except ValueError:
+            warning = f'Need additional plugin to use {self.source._driver} driver'
+        return contents, warning
 
     def __getattr__(self, attr):
         if attr in self.__dict__:
--- a/intake/catalog/entry.py
+++ b/intake/catalog/entry.py
@@ -129,14 +129,14 @@ class CatalogEntry(DictSerialiseMixin):
     def _ipython_display_(self):
         """Display the entry as a rich object in an IPython session."""
         contents, warning = self._display_content()
-        display({
+        display({ # noqa: F821
             'application/json': contents,
             'text/plain': pretty_describe(contents)
         }, metadata={
             'application/json': { 'root': self.name }
         }, raw=True)
         if warning:
-            display(warning)
+            display(warning) # noqa: F821
 
     def _display_content(self):
         """Create a dictionary with content to display in reprs."""
--- a/intake/catalog/entry.py
+++ b/intake/catalog/entry.py
@@ -144,10 +144,6 @@ class CatalogEntry(DictSerialiseMixin):
         warning = None
         try:
             contents.update(self.describe_open())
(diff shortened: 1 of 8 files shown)


ian-r-rose:
Likely to conflict with #325, happy to deal with those however you see fit.

intake/catalog/entry.py
```diff
@@ -125,6 +126,32 @@ def plots(self):
...
+        if warning:
+            display(warning) # noqa: F821
+
+    def _display_content(self):
+        """Create a dictionary with content to display in reprs."""
+        contents = deepcopy(self.describe())
+        warning = None
+        try:
+            contents.update(self.describe_open())
+            if 'plots' in contents['metadata']:
```
jsignell: The removal of the plots dict is specific to the use in the GUI and should just be left in for this case.

intake/catalog/entry.py
```diff
@@ -125,6 +126,32 @@ def plots(self):
...
+                contents['metadata'].pop('plots')
+            if 'plots' in contents['args'].get('metadata', {}):
```
jsignell: @martindurant I have seen some duplication of content  between metadata and args[metadata]. I don't think this is a particularly resilient way of handing that de-deduplication. Can you think of a better way?

intake/catalog/entry.py
ian-r-rose: Okay, I've moved the plots logic back to the GUI.

jsignell:
This looks very good! I am a little worried about mapping between yaml, json and dicts too much. But the collapsing feature is very nice.

martindurant:
Well, it turns out that the JSON representation happens to look exactly like YAML, so that's alright with me. Does this require a browser plugin, or particular browser platform/version, though?

ian-r-rose:
The JSON renderer should work fine in JupyterLab and nteract (not sure about colab), and across all supported browsers. For the classic Jupyter notebook it should use the pretty print fallback.

intake/catalog/entry.py
```diff
@@ -125,6 +126,28 @@ def plots(self):
...
         return list(self._metadata.get('plots', {}))
 
+    def _ipython_display_(self):
+        """Display the entry as a rich object in an IPython session."""
+        contents, warning = self._display_content()
+        display({ # noqa: F821
+            'application/json': contents,
+            'text/plain': pretty_describe(contents)
+        }, metadata={
+            'application/json': { 'root': self.name }
```
jsignell:
```suggestion
            'application/json': { 'root': self._name }
```
You'll want to use the underscore name to avoid initializing the entry I think

martindurant: Do you think we should have a test driver specifically to guard against this?

```
class DoNotInitialise(Source):

    def __init__(self, *args, **kwargs):
        raise RuntimeError
```

ian-r-rose: I see that `_name` is defined on `LocalCatalogEntry` and `name` is defined on `RemoteCatalogEntry`, and I don't see either on the base class. Is it safe to use `_name` here?

martindurant: It should consistently be `_name`. I don't know how the attributes all ended up underscored for the entries, but I suppose we should stick with it for now.

ian-r-rose: I suppose I am missing some context/motivation here. For my own edification, what is the reason to avoid initializing?

martindurant: It can take some non-zero time. In theory, sources ought to defer any serious processing until at least `.discover()` (i.e., inferring the schema and other important details) but in practice some stuff can happen in the driver `__init__` too.

jsignell:
It's fine will me if you want to land this first and I can deal with any merge conflicts on the #325 side

intake/catalog/entry.py
```diff
@@ -125,6 +126,28 @@ def plots(self):
...
         return list(self._metadata.get('plots', {}))
 
+    def _ipython_display_(self):
+        """Display the entry as a rich object in an IPython session."""
+        contents, warning = self._display_content()
+        display({ # noqa: F821
+            'application/json': contents,
+            'text/plain': pretty_describe(contents)
+        }, metadata={
+            'application/json': { 'root': self._name }
```
martindurant: Does this now error on remote entries, then?
A couple of ways to fix this: 
- use `_name` in remote entries throughout
- make sure `.describe` contains a "name" key and don't use attribute at all
- move these basic attributes to `Entry` base class and define only once (and move `describe()` implementation)

martindurant:
I think given the conversation around attributes and the like, there had better be a test that shows that the display method can produce something sensible for both a local entry and a remote entry.

intake/catalog/entry.py
ian-r-rose: Happy to take a crack at whichever of these seems most sensible to you.

ian-r-rose: It seems to me that if the base `Entry` class contains the `display` logic, then it should not use attributes specific to subclasses. This would mean moving at least `_name` and `_driver` to `Entry`.

ian-r-rose: As in, option 3 of your list.

martindurant: Fair enough - although the remote ones don't actually have a `plugin` (it get resolved only at open time), but `None` would be a fine default.

ian-r-rose: I'm actually now taking a closer look at number two. It involves a bunch of changes to test outputs, but is otherwise rather straightforward. Thoughts?

martindurant: Certainly seems the simplest change, and we can maybe refactor as number 3 later

ian-r-rose:
Okay, ready for another look.

martindurant:
Thanks for this.

ian-r-rose:
Thanks for the reviews @jsignell and @martindurant

PR #329, opened 2019-04-25
Improving pretty_describe func

jsignell:
Add a newline for every dict
Screen Shot 2019-04-25 at 1 03 43 PM

PR #330, opened 2019-04-25
Simplify describe and describe_open

martindurant:
Following #329 , this would be something like "option 3"; this fails on a couple of tests to do with evaluating argument templates on the server when opening directly. I don't have a good answer for what is the right thing to do yet, so posting this for conversation.

### DIFF
--- a/intake/gui/source/description.py
+++ b/intake/gui/source/description.py
@@ -61,12 +61,12 @@ class Description(BaseView):
         """String representation of the source's description"""
         if not self._source:
             return ' ' * 100  # HACK - make sure that area is big
-        contents, warning = self.source._display_content()
+        contents = self.source.describe()
         if 'metadata' in contents and 'plots' in contents['metadata']:
             contents['metadata'].pop('plots')
         if 'args' in contents and 'plots' in contents['args'].get('metadata', {}):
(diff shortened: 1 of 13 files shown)


martindurant:
@danielballan , thought you might have some comments on this, and it ought to be tried with Bluesky, which is so far the only cat I know of so far that is designed for remote.

intake/gui/source/description.py
```diff
@@ -61,12 +61,12 @@ def contents(self):
         """String representation of the source's description"""
         if not self._source:
             return ' ' * 100  # HACK - make sure that area is big
-        contents, warning = self.source._display_content()
+        contents = self.source.describe()
         if 'metadata' in contents and 'plots' in contents['metadata']:
             contents['metadata'].pop('plots')
         if 'args' in contents and 'plots' in contents['args'].get('metadata', {}):
```
jsignell: Does this  duplication still happen? Might want to remove otherwise.

jsignell:
Sorry that was a false alarm. This PR seems good.

martindurant:
Will wait until after #331

ian-r-rose:
:+1:

PR #331, opened 2019-04-26
Don't pop plots from metadata

jsignell:
This fixes bug where no plots were showing up in the GUI

### DIFF
--- a/docs/source/api_user.rst
+++ b/docs/source/api_user.rst
@@ -13,7 +13,7 @@ These are reference class and function definitions likely to be useful to everyo
    intake.source.zarr.ZarrArraySource
    intake.catalog.local.YAMLFileCatalog
    intake.catalog.local.YAMLFilesCatalog
-   intake.gui.DataBrowser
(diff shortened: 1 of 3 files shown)


martindurant:
Perhaps needs a change to the test which specifically looks at the output of the description panel?

martindurant:
I notice that this is still installing ipywidgets: https://travis-ci.org/intake/intake/jobs/525040505#L675

martindurant:
I think it's because of test_requirements.txt

jsignell:
Isn't ipywidgets still required for tqdm?

martindurant:
Also (sorry), can you update docs/source/api_user.rst to swap out `intake.gui.DataBrowser`? It occurs in the file twice.

martindurant:
> Isn't ipywidgets still required for tqdm?

Even there it ought to be optional, and our tests don't depend on it.

martindurant:
That could have been skip-ci :)
Merging this now.

PR #334, opened 2019-04-30
DCAT plugin

ian-r-rose:
`intake-dcat` now works as a data driver in its own right. Also adding a link to a demo binder.


martindurant:
Thanks for this!

PR #335, opened 2019-05-01
Use LRU to cache latest source for an entry

### LINKED ISSUE #333: Catalogs as catalog entries and _load() performance
Over at [`intake-dcat`](https://github.com/CityOfLosAngeles/intake-dcat/tree/master/intake_dcat) I've got things set up so that DCAT catalogs are a data source themselves, as described in the intake docs. I was a bit surprised to see that every access of the catalog source calls the `_load()` function on it. In this case, that involves a network request for a few-megabyte file, so inspecting the catalog contents becomes pretty laggy.

1. Is this behavior intended? I might expect the catalog source to not need to re-initialize/reload itself unless explicitly forced to.
2. Is there a recommended way around this? I expect these catalogs to be slowly moving, so I can cache the entries after the first load, and then shortcut `_load()` thereafter. But perhaps there is a way that is considered more idiomatic for `intake`?

martindurant:
Fixes #333

### DIFF
--- a/intake/catalog/entry.py
+++ b/intake/catalog/entry.py
@@ -61,7 +61,6 @@ class CatalogEntry(DictSerialiseMixin):
         """
         raise NotImplementedError
 
-    @functools.lru_cache(1)
     def __call__(self, persist=None, **kwargs):
         """Instantiate DataSource with given user arguments
 
@@ -92,7 +91,9 @@ class CatalogEntry(DictSerialiseMixin):
 
     def _get_default_source(self):
         """Instantiate DataSource with default agruments"""
-        return self()
+        if self._default_source is None:
+            self._default_source = self()
+        return self._default_source
 
     @property
     def has_been_persisted(self, **kwargs):
@@ -108,7 +109,7 @@ class CatalogEntry(DictSerialiseMixin):
         """Display the entry as a rich object in an IPython session."""
         from IPython.display import display
         contents = self.describe()
-        display({  # noqa: F821
+        display({
             'application/json': contents,
             'text/plain': pretty_describe(contents)
         }, metadata={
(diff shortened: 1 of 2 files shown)


intake/catalog/entry.py
```diff
@@ -102,6 +106,7 @@ def plots(self):
 
     def _ipython_display_(self):
         """Display the entry as a rich object in an IPython session."""
+        from IPython.display import display
         contents = self.describe()
         display({  # noqa: F821
```
ian-r-rose: With this import the lint directive is no longer relevant. I had slid on the import because this function is only meant to be invoked in an `IPython` context, when the `display` function is automatically in the global namespace.

martindurant: Right, I see what it was doing now.

martindurant:
This actually - surprisingly - produces a whole load of failures, so will need more work.

ian-r-rose:
yikes, would you look at all that red.

intake/catalog/entry.py
ian-r-rose: No harm in importing though, it's more explicit.

ian-r-rose:
Thanks for the quick fix!

danielballan:
This is well timed. In intake-bluesky we were finding it very easy to accidentally recreate the Catalog source in a loop, standing up a large number of redundant `MongoClients` and eventually running out of file descriptors. This should fix that.

martindurant:
hurray!

intake/catalog/entry.py
```diff
@@ -6,6 +6,7 @@
 #-----------------------------------------------------------------------------
 
 from copy import deepcopy
+import functools
```
ian-r-rose: This import is no longer required.

PR #339, opened 2019-05-02
split requirements

### LINKED ISSUE #97: Slim down requirements
Currently the requirements file looks like the following:

```
jinja2
msgpack-numpy
msgpack-python
numpy
pandas
pytest
python-snappy
ruamel.yaml >= 0.15.0
requests
appdirs
six
tornado >= 4.5.1
dask[complete]
holoviews
```

This makes it difficult for downstream deployments to easily depend on it.  I wonder if perhaps some of these requirements could be made optional for minimal deployments.

martindurant:
Supersedes #253 
Fixes #97 

To do: add extra sections to setup.py
- server (tornado, msgpack, maybe snappy)
- dataframe (dask-dataframe)
- plot (panel, hvplot)

### DIFF
--- a/intake/container/serializer.py
+++ b/intake/container/serializer.py
@@ -37,14 +37,19 @@ class GzipCompressor(object):
             return f.read()
 
 
+try:
+    import msgpack_numpy
+except ImportError:
+    msgpack_numpy = None
+
+
 class MsgPackSerializer(object):
     # TODO: This is ugly, should maybe transition to
     #  distributed.protocol.serialize
     name = 'msgpack'
 
     def encode(self, obj, container):
-        if container in ['ndarray', 'xarray']:
-            import msgpack_numpy
+        if container in ['ndarray', 'xarray'] and msgpack_numpy:
             return msgpack.packb(obj, default=msgpack_numpy.encode)
         elif container == 'dataframe':
             return obj.to_msgpack()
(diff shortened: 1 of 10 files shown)


martindurant:
How would I include the server CLI endpoint only if install is [server] or [complete]?
How does one invoke the pip install sections locally?

danielballan:
> How would I include the server CLI endpoint only if install is [server] or [complete]?

As far as I can tell you do you not have access to that. In any case it could arguably (just arguably) be better to install the `intake-server` endpoint always but have it check its dependencies at the top of the module and produce an informative error message if they are missing.

> How does one invoke the pip install sections locally?

Do you mean as in:

```sh
pip install -e .[server]
```

intake/container/serializer.py
```diff
@@ -102,10 +91,28 @@ def decode(self, bytestr, container):
...
+        name = 'snappy'
+
+        def compress(self, data):
+            return snappy.compress(data)
+
+        def decompress(self, data):
+            return snappy.decompress(data)
+
+
+    compressors = [SnappyCompressor(), GzipCompressor(),
```
danielballan: A somewhat minor thing, but it would avoid repetition to set `compressors = [GzipCompressor(), NoneCompressor()]` once, above, and then _append_ to the `compressors` here.

martindurant:
> pip install -e .[server]

exactly!

requirements.txt
```diff
@@ -1,15 +1,8 @@
 # This file is also processed by conda-build, so keep spaces as "pkg >=1.2"
 appdirs
-cloudpickle
-dask >=0.17.0
-holoviews
+dask[bag] >=1.0
 jinja2
-msgpack-numpy
-msgpack-python
```
danielballan: Was this dropped intentionally? Will it be installed an indirect dependency? I can't find it as a dependency of either msgpack-numpy or dask.

martindurant: It's just `msgpack` in pypi... I've added back `msgpack-numpy` in the dataframe section, and also made it optional. Note that the conda dependencies have not changed at all yet.

martindurant:
This will need some careful monitoring at release time

PR #349, opened 2019-05-08
Allow local cat entry to specify explicit driver class

### LINKED ISSUE #348: LocalCatalogEntry should raise an error if unexpected driver type specified
Consider this example:

```python
LocalCatalogEntry(name='foo', description='bar', driver=object)
```

This raises:

```python-traceback
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-20-68571aef1b6a> in <module>
----> 1 LocalCatalogEntry(name='foo', description='bar', driver=None)

~/miniconda3/envs/intake/lib/python3.7/site-packages/intake/catalog/local.py in __init__(self, name, description, driver, direct_access, args, cache, parameters, metadata, catalog_dir, getenv, getshell, catalog)
    187                             if v is not None}
    188             containers = set(p.container for p in self._plugin.values())
--> 189         if len(containers) > 1:
    190             # this is an error, because cat is poorly specified, even if other
    191             # plugins are OK

UnboundLocalError: local variable 'containers' referenced before assignment
```

A friendly error message would be nice to see here. Something like:

```python
if not isinstance(driver, (str, list, dict)):
    raise TypeError('Expected driver to be a str, list, or dict but got %s' % type(driver))
```

Alternatively, you may just set `containers = []` at the start of this routine. 

BTW, I got originally got here by passing a Catalog type as a driver.

martindurant:
Fixes #348

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -5,6 +5,7 @@
 # The full license is in the LICENSE file, distributed with this software.
 #-----------------------------------------------------------------------------
 
+import inspect
 import logging
 import os
 import posixpath
@@ -13,7 +14,7 @@ from jinja2 import Template
 from dask.bytes import open_files
 
 from .. import __version__
-from .base import Catalog
+from .base import Catalog, DataSource
 from . import exceptions
 from .entry import CatalogEntry
 from ..source import registry as global_registry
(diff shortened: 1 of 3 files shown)


martindurant:
@jhamman , does this work for you?

jhamman:
@martindurant - it does seem to work as expected. See https://github.com/pangeo-data/intake-stac/pull/8 for the implementation I've come up with.

PR #350, opened 2019-05-09
add intake-stac to plugins list

jhamman:
closes https://github.com/pangeo-data/intake-stac/issues/7


docs/source/plugin-directory.rst
```diff
@@ -27,6 +27,7 @@ contains in parentheses:
 * `intake-postgres <https://github.com/intake/intake-postgres>`_: PostgreSQL database (``postgres``)
 * `intake-s3-manifests <https://github.com/informatics-lab/intake-s3-manifests>`_ (``s3_manifest``)
 * `intake-solr <https://github.com/intake/intake-solr>`_: Apache Solr search platform (``solr``)
+* `intake-stac <https://github.com/pangeo/intake-stac>`_: Intake Driver for `SpatioTemporal Asset Catalogs (STAC) <https://stacspec.org/>`_.
```
martindurant: Typically, the driver packages list the names of the drivers they contain, but I see that none of the intake-stac classes has the name class attribute. It is not necessary to have such a name, but the purpose is to enable including a STAC cat inside another cat; when a driver isn't found at load time, the message points users to this page so that they can figure out which package they need to install.

I'll merge this anyway, it is for your information only.

Issue #353, opened 2019-05-13
Change in v0.5.0 requires args to be msgpack-serializable.


danielballan:
Commits 2d69d1a and 1b8913b, released in v0.5.0, had the possibly-unintentional effect of adding keys to `source.describe()`, and that now includes an Entry's `args`. When the server calls `source.describe()` here

https://github.com/intake/intake/blob/95177644620378de34618cb6b4c39d8ec3f1bf95/intake/cli/server/server.py#L208

and attempts to serialize it, it will fail if any of those args are not msgpack-serializable. In intake-bluesky, we pass into `args` various callables that close over a MongoDB client, which are of course not serializable. Should we avoid passing callables to `args`? Or can intake go back to being more narrow about what `describe()` returns and/or what the server attempts to serialize?

xref https://github.com/bluesky/intake-bluesky/issues/32

martindurant:
Hm, that is a problem for you, I see. I'll have to think about it a little. Indeed, it must be possible to pass all the args at some point, so that opening a remote source directly (bypassing the server) can be possible, but maybe it can be deferred. Passing the arg values early does allow, though, for the user to override some at source instantiation time (POST) even when opening remote as opposed to direct.

martindurant:
The obvious, but perhaps suboptimal, option would be to allow "cloudpickle" as an allowed serialiser. Of course, that comes with a possible security problem...

martindurant:
Is it possible to refer to functions by a string `package.mod.func` rather than using concrete function objects? Allowing remote triggering of imports seems ok (we would do this to get a driver class anyway), preferable to unpickling arbitrary objects.

danielballan:
My functions are closures over a `pymongo.MongoClient` instance, so they can't be recreated by importing. I think I could do something clever with strings nonetheless, like passing in a key in some global registry of `pymongo.MongoClient` instances.

martindurant:
So I don't suppose they could be pickled either... I wonder if we should do something different in these cases where direct access is never possible.

danielballan:
I was mulling something along those lines. I wouldn't want to ask intake to carve out an overly-specialized case, but maybe adding an option at server startup time would make sense. For my use case that would be sufficient. For potential other use cases where a server might serve a mix of directly-accessible and not-directly-accessible Entries, maybe there should be an way for specific Entries to opt in or out of that setting. Of course, if the Entries are _not_ being accessed locally with no server, that option would not be meaningful.

danielballan:
[...]

PR #354, opened 2019-05-14
Don't attempt to serialise functions in server response

martindurant:
ref: #353

### DIFF
--- a/intake/cli/server/server.py
+++ b/intake/cli/server/server.py
@@ -207,8 +207,22 @@ class ServerSourceHandler(tornado.web.RequestHandler):
             if self.auth.allow_access(head, source, self._catalog):
                 info = source.describe()
                 info['name'] = name
+
                 source_info = dict(source=info)
-                self.write(msgpack.packb(source_info, use_bin_type=True))
+                try:
+                    out = msgpack.packb(source_info, use_bin_type=True)
+                except TypeError:
+                    info['direct_access'] = 'forbid'
+                    args = {}
+                    for k, v in info['args'].items():
+                        try:
+                            msgpack.packb(v, use_bin_type=True)
+                            args[k] = v
+                        except TypeError:
+                            args[k] = 'UNSERIALIZABLE_VALUE'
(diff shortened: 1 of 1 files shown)


martindurant:
This is a bit of a hack... I don't think there's any legitimate use for functions to be passed to the client, and the client won't pass them back, but perhaps somehow we should be careful not to accidentally override functions that are defined only server-side.

danielballan:
Ha, the branch name made me laugh. Agreed that this is a bit of hack, but maybe a good last-resort fallback even if we come up with something more graceful higher up.

The same approach would need to be taken in `ServerInfoHandler.get` as well.

danielballan:
I opened a PR against this PR https://github.com/martindurant/intake/pull/1

That _seems_ to have resolved this issue for the intake-bluesky tests, but solving it has exposed a second breakage that seems not related to serialization. Investigating.

danielballan:
In fact, the "second breakage" was just a hard-to-trace side effect of double nesting the response [here](https://github.com/intake/intake/pull/354/commits/09dcb7c6d39477537af72566dbca0a6be605f24c#diff-92c1a6d6dd4df7c92054d816cd1d4475R211) and [here](https://github.com/intake/intake/pull/354/commits/09dcb7c6d39477537af72566dbca0a6be605f24c#diff-92c1a6d6dd4df7c92054d816cd1d4475R223) in 09dcb7c. With https://github.com/martindurant/intake/pull/2, the intake-bluesky tests pass.

Thanks for indulging a bit of a hack to fix us. :-) I have set up a nightly Travis-CI Cron Job running intake-bluesky against intake `master` so that intake-bluesky should act as a more prompt "canary" in future.

martindurant:
Good idea on the cron. I was thinking of doing that against other intake repos too.

martindurant:
CI passed, but didn't update here for some reason - sorry to keep you waiting.

danielballan:
Not at all, thanks again for jumping on this!

PR #357, opened 2019-05-16
Fixing up the GUI

jsignell:
Fixes 2 small GUI issues:
 1. When using server version there was no logo
 2. The catalog and entry guis got broken in the import work

Adds:
1. More panel-like methods to the GUI (`show` and `servable`). So now you can do `python -c "import intake; intake.gui.show()"`
2. Tests of cat.gui and entry.gui so they can't get broken again.

### DIFF
--- a/intake/catalog/gui.py
+++ b/intake/catalog/gui.py
@@ -8,7 +8,14 @@ from functools import partial
 
 try:
     import panel as pn
-    from ..gui import SourceGUI, MAX_WIDTH
+    from ..gui.gui import SourceGUI, MAX_WIDTH
+    css = """
+    .scrolling {
+        overflow: scroll;
+    }
+    """
+    pn.config.raw_css.append(css)  # add scrolling class from css (panel GH#383, GH#384)
+    pn.extension()
 
     class EntryGUI(SourceGUI):
         def __init__(self, source=None, **kwargs):
@@ -51,7 +58,6 @@ try:
                 return None
             return self.sources[0]
 
-
 except ImportError:
 
     class GUI(object):
@@ -62,7 +68,6 @@ except ImportError:
 
     EntryGUI = GUI
     CatalogGUI = GUI
-
 
 except Exception as e:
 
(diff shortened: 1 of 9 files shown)


intake/gui/base.py
```diff
@@ -68,6 +68,24 @@ def __init__(self, visible=True, visible_callback=None, logo=False):
...
+                      margin=0)
+
+    @panel.setter
+    def panel(self, panel):
+        self._panel = panel
+
+    def servable(self, *args, **kwargs):
+        return self.panel.servable(*args, **kwargs)
+
+    def show(self, *args, **kwargs):
```
martindurant: May as well document these user methods?

intake/catalog/gui.py
```diff
@@ -8,7 +8,14 @@
...
     import panel as pn
-    from ..gui import SourceGUI, MAX_WIDTH
+    from ..gui.gui import SourceGUI, MAX_WIDTH
+    css = """
+    .scrolling {
+        overflow: scroll;
+    }
+    """
+    pn.config.raw_css.append(css)  # add scrolling class from css (panel GH#383, GH#384)
+    pn.extension()
```
jsignell: One thing that was broken is that the pn.extension() was never being called.

intake/catalog/gui.py
```diff
@@ -8,7 +8,14 @@
 
 try:
     import panel as pn
-    from ..gui import SourceGUI, MAX_WIDTH
+    from ..gui.gui import SourceGUI, MAX_WIDTH
```
jsignell: SourceGUI wasn't actually importable from `..gui`

martindurant:
OK, got it.
Were you intending to add any docs to those methods? I gather you'd rather this was merged soon and release.

jsignell:
Unless you think they need docs, I'd rather just merge.

martindurant:
Done. Sorry I'm slow today, officially I am off for Victoria Day.

PR #359, opened 2019-05-23
Let GUI docs be visible; don't ever error on repr

martindurant:
Fixing https://intake.readthedocs.io/en/latest/api_user.html GUI section.

### DIFF
--- a/intake/gui/base.py
+++ b/intake/gui/base.py
@@ -88,17 +88,15 @@ class Base(object):
 
     def __repr__(self):
         """Print output"""
-        try:
-            return self.panel.__repr__()
-        except:
-            raise RuntimeError("Panel does not seem to be set up properly")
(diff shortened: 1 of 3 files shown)


intake/gui/base.py
```diff
@@ -88,17 +88,15 @@ def show(self, *args, **kwargs):
...
-            raise RuntimeError("Panel does not seem to be set up properly")
+        return self.panel.__repr__()
 
     def _repr_mimebundle_(self, *args, **kwargs):
         """Display in a notebook or a server"""
         try:
             return self.panel._repr_mimebundle_(*args, **kwargs)
         except:
-            raise RuntimeError("Panel does not seem to be set up properly")
+            raise NotImplementedError("Panel does not seem to be set "
```
martindurant: Falls back to repr in basic python session

PR #360, opened 2019-05-23
More api

martindurant:


martindurant:
I am adding all the many GUI classes!

PR #362, opened 2019-05-26
Avoid accidental mutation.

danielballan:
Follow-up on #354. While adding some new functionality to intake-bluesky I
found some additional protection against mutating the internal state of
``source`` is needed.


martindurant:
This is fine, and I'll merge it, but I feel like there must be a better way to do this!

danielballan:
Yes. I pushed a commit that makes one copy instead of two, the key point being that it is safe to iterate over the original. This is at least marginally better.

You might also mean that there should be a better way to handle unserializable parameters in _general_. I've been mulling that over but don't have any new ideas yet.

martindurant:
OK, let's stick with this for now. I'll merge it when it goes green.

danielballan:
Thank you!

PR #364, opened 2019-05-30
Hide plot when changing source selection

martindurant:
cc @jsignell

### DIFF
--- a/intake/gui/source/gui.py
+++ b/intake/gui/source/gui.py
@@ -62,7 +62,7 @@ class SourceGUI(Base):
         self.description = Description(source=self.sources)
 
         self.plot = DefinedPlots(source=self.sources,
-                                 visible=False,
+                                 visible=self.plot_widget.value,
                                  visible_callback=partial(setattr, self.plot_widget, 'value'))
 
         super().__init__(**kwargs)
--- a/intake/gui/source/gui.py
+++ b/intake/gui/source/gui.py
@@ -62,7 +62,7 @@ class SourceGUI(Base):
         self.description = Description(source=self.sources)
 
         self.plot = DefinedPlots(source=self.sources,
-                                 visible=self.plot_widget.value,
+                                 visible=False,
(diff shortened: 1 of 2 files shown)


intake/gui/source/gui.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, cats=None, sources=None, done_callback=None, **kwargs):
         self.description = Description(source=self.sources)
 
         self.plot = DefinedPlots(source=self.sources,
-                                 visible=self.plot_widget.value,
+                                 visible=False,
```
jsignell: I don't think this is needed.

intake/gui/source/gui.py
```diff
@@ -112,9 +112,11 @@ def visible(self, visible):
     def callback(self, sources):
         """When a source is selected, enable widgets that depend on that condition
         and do done_callback"""
+        if hasattr(self, 'plot'):
+            # guard since this cannot happen until plot is ready
+            self.plot.visible = False
```
jsignell: I don't think this needs to be here. Setting `self.plot_widget.value = False` should do it.

intake/gui/source/gui.py
martindurant: The value is always False at this point (it's defined just above), thought it good to avoid the obfuscation

intake/gui/source/gui.py
martindurant: That is exactly what already exists two lines below, but doesn't have the right effect.

intake/gui/source/gui.py
jsignell: I think it is good to make sure it is always the same. It'd be weird if you set it to True here.

intake/gui/source/gui.py
jsignell: Huh I thought that would be enough. But if not I guess it's fine.

intake/gui/source/gui.py
martindurant: reverted

PR #365, opened 2019-05-31
Params editor

martindurant:
Create a pane in the GUI with which user-parameters can be edited for the selected entry. The instantiated data source is then available as an attribute of the source gui or the overall gui. 

(selections are not yet reflected in the plot)


martindurant:
Screen Shot 2019-06-04 at 11 18 43
Screen Shot 2019-06-04 at 11 19 06

PR #369, opened 2019-06-04
WIP: cache redux

martindurant:
- [x] key for hashing off URL, so files can be shared between sources
- [ ] allow for multiple locations
- [ ] allow for expiry and auto-update, like persist
- [ ] ~allow for incremental updates of some files~
- [ ] ~allow for versions/checkpointing~


martindurant:
Screen Shot 2019-06-04 at 11 18 43
Screen Shot 2019-06-04 at 11 19 06

PR #373, opened 2019-06-10
Upload

martindurant:


martindurant:
@seibert @mmccarty , you might have thoughts about this

intake/container/__init__.py
```diff
@@ -20,3 +20,40 @@
...
+def upload(data, path, **kwargs):
+    """Given a concrete data object, store it at given location return Source
+
+    Use this function to publicly share data which you have created in your
+    python session. Intake will try each of the container types, to see if
+    one of them can handle the input data, and write the data to the path
+    given, in the format most appropriate for the data type, e.g., parquet for
+    pandas or dask data-frames.
+
+    With the DataSource instance you get back, you can add this to a catalog,
```
martindurant: Note that YAMLFileCatalog.add_source() will come later in this PR. I don't think it's meaningful for any other type of cat class (unless specifically developed for whatever service supplies the cat).

danielballan:
Have you given much thought to how update semantics should work? Does `add` mean `upsert`?

intake/catalog/local.py
```diff
@@ -578,6 +578,23 @@ def _load(self, reload=False):
                 text = text.replace('!template ', '')
             self.parse(text)
 
+    def add(self, **sources):
```
danielballan: I would be inclined to do one `source` at a time here, akin to `set.add`.

martindurant: I had wanted to give an explicit name; but I suppose a source may well have a name, so could be `add(self, source, name=None)` to provide the optional override.

martindurant:
Right, `add`, as written, will clobber a source of the same name

PR #375, opened 2019-06-13
catalog.rst plugin selection typo

talebzeghmi:
It seems that plugin and driver is sometimes used interchangeably introducing confusion.  For example multi_plugins.yaml has a list of drivers but to select a driver one must use the `plugin` parameter (ex: `plugin='myplug2'` in test_local.py).

Another place where plugin/driver is used interchangeably is in `LocalCatalogEntry` (one example: `self._plugin = [driver]`).

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -225,8 +225,7 @@ intake-iris and the intake-xarray packages contain drivers with the name ``"netc
 are capable of reading the same files, but with different backends. Here we will describe the
 various possibilities of coping with this situation. Intake's plugin system makes it easy to encode such choices.
 
-It may be
-acceptable to use any driver which claims to handle that data type, or to give the option of
+It may be acceptable to use any driver which claims to handle that data type, or to give the option of
 which driver to use to the user, or it may be necessary to specify which precise driver(s) are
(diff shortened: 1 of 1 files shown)


martindurant:
They all used to be called "plugins", but there are several places in the code that could be considered pluggable, so I decided to use the more specific term "driver" for classes deriving from DataSource; this was refactored in the code in most places where it was convenient, but clearly there are some leftovers.

docs/source/catalog.rst
```diff
@@ -256,7 +255,7 @@ instantiation.
         args: {}
 
 These alternative plugins can also be given data-source specific names, allowing the
-user to choose at load time with `driver=` as a parameter. Additional arguments may also
+user to choose at load time with `plugin=` as a parameter. Additional arguments may also
```
martindurant: This is the code being referenced: https://github.com/intake/intake/blob/master/intake/catalog/local.py#L146 , where indeed the parameter is called "driver". I believe it would be better to change the docstring to refer to drivers only (which happen to be a type of plugin). 

Have you used `plugin=` to any effect? Perhaps that might exist elsewhere.

Issue #374, opened 2019-06-12
intake-server "can not serialize 'Timestamp' object" when we have datetime parameters


sliem:
First of all, really cool project!

I'm trying to run an intake-server with an time series data set. It uses parameters of `type: datetime`. The catalog yaml file looks something like this

```yaml
sources:
  timeseriessource:
    driver: sometimeseriesapi
    description: get some time series data
    args:
      label: "{{label}}"
      start: "{{start}}"
      end: "{{end}}"
    parameters:
      label:
        description: data identifier
        type: str
      start:
        description: start time of the time series
        type: datetime
      end:
        description: start time of the time series
        type: datetime
```

It works locally but when using `intake-server` it fails with a `RemoteCatalogError: Failed to fetch metadata.` In the `intake-server` logs we see this traceback
```python
Traceback (most recent call last):
  File "/home/sliem/intake/.venv/lib/python3.7/site-packages/tornado/web.py", line 1697, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/home/sliem/intake/.venv/lib/python3.7/site-packages/intake/cli/server/server.py", line 132, in get
    self.write(msgpack.packb(server_info, use_bin_type=True))
  File "/home/sliem/intake/.venv/lib/python3.7/site-packages/msgpack/__init__.py", line 46, in
packb
    return Packer(**kwargs).pack(o)
  File "msgpack/_packer.pyx", line 282, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 288, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 285, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 232, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 261, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 232, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 261, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 232, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 279, in msgpack._cmsgpack.Packer._pack
TypeError: can not serialize 'Timestamp' object
```

[...]

PR #376, opened 2019-06-25
Add support for msgpack to handle datetime object, 

sliem:
Cf. issue #374, allowing intake-server to serve datasets with datetime-typed user parameters.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -21,8 +21,7 @@ from ..auth.base import BaseClientAuth
 from .remote import RemoteCatalogEntry
 from .utils import flatten, reload_on_change, RemoteCatalogError
 from ..source.base import DataSource
-from ..compat import unpack_kwargs
-from ..utils import encode_datetime, decode_datetime
+from ..compat import unpack_kwargs, pack_kwargs
 logger = logging.getLogger('intake')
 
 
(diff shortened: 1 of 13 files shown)


intake/catalog/base.py
```diff
@@ -608,7 +609,7 @@ def fetch_page(self, page_offset):
             raise RemoteCatalogError(
                 "Failed to fetch page of entries {}-{}."
                 "".format(page_offset, page_offset + self._page_size)) from err
-        info = msgpack.unpackb(response.content, **unpack_kwargs)
+        info = msgpack.unpackb(response.content, object_hook=decode_datetime, **unpack_kwargs)
```
danielballan: Could this be added to `unpack_kwargs` instead of added to each call individually?

intake/catalog/base.py
```diff
@@ -721,12 +722,12 @@ def search(self, *args, **kwargs):
                    'source_id': self._source_id}
         response = requests.post(
             url=self.source_url, **self._get_http_args({}),
-            data=msgpack.packb(request, use_bin_type=True))
+            data=msgpack.packb(request, default=encode_datetime, use_bin_type=True))
```
danielballan: I wonder if it is worth adding `pack_kwargs`? Since we use the same parameters every time, might as well make it easy to keep all the calls in sync.

sliem:
Now it uses the unpack_kwargs and a new pack_kwargs to pass the {de|en}code_datetime to all msgpack calls. Also added np_pack_kwargs, and np_unpack_kwargs. Not terribly sure those are useful.

danielballan:
Thanks!

> Also added np_pack_kwargs, and np_unpack_kwargs. Not terribly sure those are useful.

If someone ever tries to put a numpy array in _metadata_ that would come up. Apparently it hasn't yet, but it seems plausible that someone could trip that, so as long as the overhead isn't high it seems worth having from my point of view.

martindurant:
Nice changes, thank you!

PR #379, opened 2019-07-08
Make test_which more robust.

danielballan:
On my system, ``test_which`` get confused by the soft-link between
``python3`` and ``python``.

```python
    def test_which():
        p = default.which('python')
>       assert p == sys.executable
E       AssertionError: assert '/home/dallan...nl/bin/python' == '/home/dallan/...l/bin/python3'
E         - /home/dallan/venv/bnl/bin/python
E         + /home/dallan/venv/bnl/bin/python3
E         ?                                 +

intake/catalog/tests/test_default.py:15: AssertionError
```

This PR uses ``Path.resolve()`` to avoid false-positive failure.

### DIFF
--- a/intake/catalog/tests/test_default.py
+++ b/intake/catalog/tests/test_default.py
@@ -5,6 +5,7 @@
 # The full license is in the LICENSE file, distributed with this software.
 #-----------------------------------------------------------------------------
 
+from pathlib import Path
 import sys
 from intake.catalog import default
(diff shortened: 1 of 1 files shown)


martindurant:
I generally dislike `Path`s, but in this case, it really seems to make a succinct solution.

danielballan:
Fair enough. `os.path.realpath` would also have worked.

PR #381, opened 2019-07-11
rebasing intake on fsspec instead of dask.bytes

martindurant:


martindurant:
Still need to dereference CATALOG_DIR

martindurant:
This must not be released before dask is.

PR #394, opened 2019-07-26
Add a load of tests

martindurant:
Pushes coverage over 90%

Anyone know why the coveralls upload isn't working?

### DIFF
--- a/intake/cli/client/subcommands/drivers.py
+++ b/intake/cli/client/subcommands/drivers.py
@@ -29,6 +29,7 @@ from intake import __version__
 from intake.cli.util import Subcommand
 from intake.source.discovery import autodiscover, autodiscover_all, enable, disable
 from intake.config import confdir
+import intake
 
 #-----------------------------------------------------------------------------
 # API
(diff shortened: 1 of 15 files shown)


intake/cli/client/subcommands/drivers.py
```diff
@@ -67,15 +72,24 @@ def _list(self, args):
             fmt = '{name:<30}{cls.__module__}.{cls.__name__}'
         drivers_by_name = autodiscover()   # dict mapping name to driver
         all_drivers = autodiscover_all()  # listof (name, driver)
-        print("Enabled:", file=sys.stderr)
+        direct = {k: v for k, v in intake.registry.items()
+                  if k not in all_drivers and k not in drivers_by_name}
+
+        print("Direct:", file=sys.stderr)
```
martindurant: @danielballan - these are all the drivers that simply assign themselves to the registry. Should it be possibel to disable them? I'd say yes; perhaps that means that Intake itself should declare its own entrypoints.
Also, I do you think that "enable" should be able to use a name alone, if that name is in the current disable list? Obviously, that only works for unique names, so perhaps not.

danielballan: >  I'd say yes; perhaps that means that Intake itself should declare its own entrypoints.
I agree. That's what `nbconvert` does.

> Also, I do you think that "enable" should be able to use a name alone, if that name is in the current disable list? Obviously, that only works for unique names, so perhaps not.

Yeah, I lean towards no for that reason. Let's see if people ask for it much.

danielballan: > I'd say yes; perhaps that means that Intake itself should declare its own entrypoints.

I agree. That's what nbconvert does.

> Also, I do you think that "enable" should be able to use a name alone, if that name is in the current disable list? Obviously, that only works for unique names, so perhaps not.

Yeah, I lean towards no for that reason. Let's see if people ask for it much.

martindurant: OK, let me try to add entrypoints to intake itself, and see if those "direct" entries all go away.

Issue #347, opened 2019-05-07
Possibility for ibis container


ian-r-rose:
Has there been any discussion of adding an [`ibis`](https://github.com/ibis-project/ibis) container for intake sources? This might be a nice option for deferred loading of large datasets with SQL-like backends.

martindurant:
Certainly a fine idea, but no, it hasn't specifically come up in conversations with me before (although I did momentarily misread that as https://github.com/informatics-lab/intake-iris ). I'm not sure quite how you would express the dataframe-like syntax in a way that could be conveniently packed into a yaml spec.

ian-r-rose:
I guess I was envisioning that you would get an ibis connection object, rather than an already-constructed expression, so the yaml specification would look much like the existing SQL drivers.

martindurant:
Yes, that would be fine, and in-keeping with what other drivers do. Probably `.read()` would give you the pandas equivalent, and another method gives you the ibis object. However, the ability to prescribe the full expression is also interesting - this is currently done in intake-spark only.

Do you know if ibis outputs are chunkable for dask?

ian-r-rose:
I don't think they are chunkable as-is, though once they are evaluated they are regular dataframes.

martindurant:
Are you volunteering to make such a driver, @ian-r-rose ?

ian-r-rose:
Possibly, but not on any immediate timescale :)

I'm also unsure about whether it should be considered a new driver or a new container type.

martindurant:
It seems to me that this would be fine as a dataframe-type driver.

ian-r-rose:
But if it immediately executes the expression, then you lose the benefit of the more lazy expression building, and might as well use `intake-sql`.

martindurant:
>  if it immediately executes the expression

providing an alternate method to expose the expression object is fine, just like `.to_dask` produces lazy dask dataframes, but the type of data really is dataframe-like, and I would expect `.read()` to produce pandas.

costrouc:
I am looking to implement this feature or some datasource that has ibis compatibility. We currently have intake catalogs with some dataframes taking up around 50GB of memory due to complex joins. Would like to be able to apply filters to limit the number of rows fetches (for example last 2 months of data instead of 5 years). Right now I am think that I will just implement this in an adhok way but would like to contribute it back once we have an idea of what this would look like

martindurant:
[...]

PR #395, opened 2019-07-26
initial pass at a to_ibis function.

ian-r-rose:
A first pass at implementing #347.  This PR is mostly about establishing an interface for drivers to implement -- it doesn't really do anything useful here.

The idea is that driver authors implement `to_ibis()` which knows how to produce ibis clients for lazy evaluation of expressions on SQL like backends.

### DIFF
--- a/intake/container/dataframe.py
+++ b/intake/container/dataframe.py
@@ -51,6 +51,12 @@ class RemoteDataFrame(RemoteSource):
         self._load_metadata()
         return self.dataframe
 
+    def to_ibis(self):
+        import ibis.pandas
+        df = self.read()
+        con = ibis.pandas.connect({'source': df })
+        return con.table('source')
+
     def _close(self):
         self.dataframe = None
 
@@ -177,6 +183,12 @@ class GenericDataFrame(DataSource):
     def to_dask(self):
         self._load_metadata()
         return self.dataframe
+
+    def to_ibis(self):
+        import ibis.pandas
+        df = self.read()
+        con = ibis.pandas.connect({'source': df })
+        return con.table('source')
 
     def _close(self):
         self.dataframe = None
--- a/intake/source/base.py
+++ b/intake/source/base.py
@@ -203,12 +203,21 @@ class DataSource(DictSerialiseMixin):
         """Return a dask container for this data source"""
         raise NotImplementedError
 
+    def to_ibis(self):
+        """Return an ibis expression for this data source
+
+        Ibis expressions allow construction and lazy evaluation of queries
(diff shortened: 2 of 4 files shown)


intake/container/dataframe.py
```diff
@@ -51,6 +51,12 @@ def to_dask(self):
         self._load_metadata()
         return self.dataframe
 
+    def to_ibis(self):
+        import ibis.pandas
```
ian-r-rose: These are cute, but perhaps not very useful. We can wrap an dataframe in an ibis client and perform lazy operations on them. It all happens in memory, however. I'm agnostic to whether they should actually be included here.

intake/source/base.py
```diff
@@ -203,12 +203,21 @@ def to_dask(self):
         """Return a dask container for this data source"""
         raise NotImplementedError
 
+    def to_ibis(self):
```
ian-r-rose: The actual proposed interface.

intake/container/dataframe.py
martindurant: I take it, there's no way to wrap a Dask lazy dataframe

intake/source/textfiles.py
```diff
@@ -109,6 +109,12 @@ def to_dask(self):
         return db.from_delayed([dfile(f, self.decoder, self._read)
                                 for f in self._files])
 
+    def to_ibis(self):
+        import ibis.pandas
+        df = self.read()
```
martindurant: This is a list of text strings, not a dataframe, so you probably can't ibis this

intake/container/dataframe.py
ian-r-rose: I don't think so. Or, at least, it is not designed to be able to do that.

intake/source/textfiles.py
ian-r-rose: Ah, good point, I'll remove this.

ian-r-rose:
> Still thinking about this. I don't think the SQL portion requires this code, right?

Right, this PR is almost a no-op -- it's more a suggested interface that drivers *may* want to implement.

martindurant:
How many drivers do you think might be compatible with ibis?

ian-r-rose:
I'm mostly thinking of drivers for SQL databases. I know this would be useful for `intake-sql` and `intake-omnisci` (and I personally have some interest in using it with PostGIS). It may also be useful for `intake-spark`, though I haven't really run that to ground.

martindurant:
Let's see how it goes with the other repos first. I don't see it for spark, unless the user happens to be doing SQL only (which I don't think you have an easy way to know).

PR #396, opened 2019-07-29
Add Entrypoints

martindurant:


martindurant:
Ref #394

intake/cli/client/subcommands/drivers.py
```diff
@@ -76,20 +81,34 @@ def _list(self, args):
                   if k not in all_drivers and k not in drivers_by_name}
 
         print("Direct:", file=sys.stderr)
-        for name, cls in direct.items():
+        none = True
```
danielballan: Have you considered the pattern:

```python
for ...:
    ...
else:
    print("<none>")
```

The `for...else` usage is slightly obscure, which is a downside of this approach, but it is more succinct and saves you the `none` variable.

martindurant: I did exactly that, but didn't get the right behaviour, especially below, where not every element of `for` counts.

danielballan: Ohhh. Got it. Sorry for the noise, carry on! :-)

PR #398, opened 2019-08-02
add viz layer to GUI

martindurant:


martindurant:
Code to reproduce slowness:
```
"""
param 1.9.1
bokeh 1.3.4
panel 0.6.2
python 3.7
intake-xarray 0.3.1
zarr 2.3.2

> pip install git+https://github.com/intake/xrviz
> pip install git+https://github.com/martindurant/intake.git@viz
"""
import xarray as xr
xarr = xr.tutorial.open_dataset('air_temperature')
xarr.to_zarr('temp.zarr')

# simple invocation
import xrviz
xarr = xr.open_zarr('temp.zarr')
xrviz.XrViz(xarr).show()


# via intake
with open('cat.yaml', 'w') as f:
    f.write("""
sources:
  z1:
    driver: zarr
    args:
      urlpath: "{{CATALOG_DIR}}/temp.zarr"
"""
            )

import intake
cat = intake.open_catalog("cat.yaml")
cat.gui.select.widget.value = [cat.z1]  # select source
cat.gui.plot_widget.value = True  # open plots
cat.gui.plot.interact(None)  # open xrviz
cat.gui.show()
```

martindurant:
^ @philippjfr

martindurant:
(also @hdsingh )

philippjfr:
Looking at this now, sorry for the delay.

intake/gui/source/defined_plots.py
```diff
@@ -4,14 +4,27 @@
 #
 # The full license is in the LICENSE file, distributed with this software.
 #-----------------------------------------------------------------------------
-from copy import deepcopy
+from distutils.version import LooseVersion
+
+try:
+    import dfviz
+    assert LooseVersion(dfviz.__version__) >= LooseVersion("0.1.0")
```
philippjfr: Confused by this, the highest version of dfviz seems to be 0.0.3, so this isn't allowing me to display the GUI at all.

philippjfr:
Having some issues running stuff with latest intake-xarray and fsspec:

```
/Users/philippjfr/miniconda3/envs/pyviz/lib/python3.6/site-packages/fsspec/core.py open_files L157
/Users/philippjfr/miniconda3/envs/pyviz/lib/python3.6/site-packages/fsspec/core.py get_fs_token_paths L315
	AttributeError: type object 'S3FileSystem' has no attribute '_get_kwargs_from_urls'
```

intake/gui/source/defined_plots.py
martindurant: https://anaconda.org/conda-forge/dfviz/files version is 0.1.0

philippjfr: Ah, thanks, I just cloned the repo.

philippjfr:
No I struggled making this behave any better. Last thing I tried was trying to come up with minimal example with similar performance issues but couldn't come up with anything. One easy solution would be to use a custom template for a deployed dashboard but this wouldn't help in the notebook. I've asked @jsignell to try to replicate the performance issues in a more minimal example so we can narrow down the issue. I'm on vacation for the next two weeks.

philippjfr:
Okay, had another go at looking at this. By removing parts of the xrviz GUI bit by bit I worked out that there are two particular components that slow things down **a lot**. The first is the `Describer.panel` but the worst offender is the `Fields.panel`. My guess is that it's the deep nesting in the `Fields.panel` that's giving the layout solver such a headache and that the `Describer` `HTML` pane is also a bit of an issue. I'll see what I can do to change those to make them a bit easier on the solver. It's also the case that the browser has much bigger issues solving these inside JupyterLab than it does in the classic notebook and a standalone dashboard is fastest of all.

In addition to trying to optimize these so the solver has an easier time I'll probably try to provide some custom templates, which will allow us to put together a much more polished standalone GUI. The unfortunate thing is that custom templates do not currently work inside a notebook.

philippjfr:
Just for reference, here's what bokeh's layout solver is trying to do. Each little block is it trying to measure the size of a component, the deeper the components are nested the harder it has to work:

Screen Shot 2019-09-27 at 2 34 26 AM

Looking at that I think getting templates to work inside the notebook so that the layout solver can solve each part of the GUI independently is our only real hope but I will have a go going through the xrviz GUI and providing as many hints to the solver as possible to see what difference that can make.

philippjfr:
Another part of the problem is that selecting a variable seems to trigger a cascade of events which trigger multiple rerenders. We can hopefully improve that as well.

martindurant:
This last point sounds like the clincher - if the layout is being redone multiple times, that's far more impactful than the layout being slow in the first place. In xrviz alone, this would be many times a small number, in the gui in a notebook, many times a bigger number. Sounds like I was wrong to assume the python side had no problem...

philippjfr:
>This last point sounds like the clincher - if the layout is being redone multiple times, that's far more impactful than the layout being slow in the first place.

True, although it makes a big difference if the layout takes 15 seconds per rerender or 100 ms per rerender. By using custom templating we can render the intake GUI and xrviz GUI as separate bokeh layouts which should get us closer to the 100 ms render. I've also got promising news on that front, I've just opened PR open that can render custom templates inside the notebook: https://github.com/pyviz/panel/pull/666

Screen Shot 2019-09-27 at 3 27 38 PM

Once I get comms working this will provide a really good alternative.

jsignell:
Oh! The template inside the notebook seems like it'll be very useful outside of xrviz as well. Especially if the grid layout isn't working too well.

philippjfr:
>Sounds like I was wrong to assume the python side had no problem...

There's definitely an issue but we can probably solve it pretty easily by using bokeh's ability to queue and merge events. Once I've merged the required changes into panel I'll open PRs on intake and xrviz to take advantage of the new features.

jsignell:
This now depends on the next release of panel.

jsignell:
I added templates for the guis that are defined on catalogs, entries, and on intake itself. It is slightly strange to use templates because the gui classes have the concept of clearing all the children out of the panel and I don't think that is doable with templates. Maybe that whole concept is not really appropriate for the top level gui components. 

Items left to do on this PR:

 - [ ] add yaml to the yaml tab. I am not sure whether this should be done on xrviz and dfviz or on intake itself. Probably the easiest way would be when the yaml tab becomes the active tab, populate it from `<df/xrviz.object>.kwargs`.
 - [ ] do more templating. Currently only the most basic templating is completed, but probably if templating were implemented more thoroughly it would reduce the render time.

Additionally there is still real work to do on xrviz. When the data is at all big (even the air temperature example), things get very slow. Even selecting a variable leads to the kernel being busy for many seconds. There must be some read process going on that is probably not fully intentional.

martindurant:
> I am not sure whether this should be done on xrviz and dfviz or on intake itself

I can see it both ways: you want to get the parameters out of df/xrviz (but you already have the kwargs); but the yaml format is intake-specific.

philippjfr:
I've just merged some patches for the bokeh layout solver into Panel that gives dramatic speedups so I'm hoping we can abandon the templating approach and just go for the original naive approach. I'll give this a try as soon as I get a chance.

martindurant:
much appreciated, @philippjfr

philippjfr:
Okay, so after removing the templating stuff and using Panel dev here's what the performance looks like:

intake_gui

I think that's pretty good, as the delay also involves loading the data. Also there are a number of low hanging fruit here since it still seems to do some thing fairly inefficiently. I'll try to rebase and then push my changes up as long as I have permissions.

philippjfr:
That being said, maybe we still want a template, at least for the standalone app to make the whole thing look a bit more polished. What do you think @martindurant?

philippjfr:
Looks like I don't have permissions.

martindurant:
I invited you to the org, which should give you permissions

martindurant:
@philippjfr , please remind me: where rae we up to here?

philippjfr:
We're only waiting on me to have sufficient bandwidth to finish this. Sorry this has dragged so long, bokeh 2.0 has taking a significant chunk of my time and so has HoloViews 1.13. Both are due out this week.

martindurant:
Is there anything I can do to help?

philippjfr:
Probably not much, I don't think this is much effort at this point but I do need to spend an hour or two to double check things.

philippjfr:
@martindurant I now have permissions for this repo but the branch is on your fork which it seems I don't have permissions for.

martindurant:
You are now added to my branch too. I thought collaborators to the main project could push to forks, or perhaps I didn't tick that box.

philippjfr:
Fixing those unit tests tomorrow morning and then hopefully we can get this merged finally.

martindurant:
Hurray!

philippjfr:
Maybe can we meet next week to discuss the future of dfviz and xrviz?

martindurant:
Certainly, book me in.

philippjfr:
@martindurant Okay I've figured out the issue here. The problem is that because we didn't pin bokeh<2.0 in earlier versions of HoloViews the imports break and the GUI fails. I can pin bokeh 2.0 but my plan is to finally finish the HoloViews 1.13.0 release which does support bokeh 2.0. Up to you if you want to merge with the bokeh<2.0 pins.

martindurant:
Done!

Docs to follow...

PR #400, opened 2019-08-06
Don't necessarily pass storage_options

### LINKED ISSUE #399: Extra kwargs passed to drivers
The filesystem work in #381 is causing some problems with `intake_geopandas`. So far as I can tell, it tries to add a `'storage_options'` kwarg to the open args, which is not recognized by the plugin. Requiring this kwarg seems a bit strict to me (though we could go ahead and ignore it downstream).

```text.python.traceback
/opt/conda/lib/python3.7/site-packages/intake/catalog/entry.py in __getattr__(self, attr)
    119             return self.__dict__[attr]
    120         else:
--> 121             return getattr(self._get_default_source(), attr)
    122 
    123     def __dir__(self):

/opt/conda/lib/python3.7/site-packages/intake/catalog/entry.py in _get_default_source(self)
     91         """Instantiate DataSource with default agruments"""
     92         if self._default_source is None:
---> 93             self._default_source = self()
     94         return self._default_source
     95 

/opt/conda/lib/python3.7/site-packages/intake/catalog/entry.py in __call__(self, persist, **kwargs)
     76             raise ValueError('Persist value (%s) not understood' % persist)
     77         persist = persist or self._pmode
---> 78         s = self.get(**kwargs)
     79         if s.has_been_persisted and persist is not 'never':
     80             s2 = s.get_persisted()

/opt/conda/lib/python3.7/site-packages/intake/catalog/local.py in get(self, **user_parameters)
    279         """Instantiate the DataSource for the given parameters"""
    280         plugin, open_args = self._create_open_args(user_parameters)
--> 281         data_source = plugin(**open_args)
    282         data_source.catalog_object = self._catalog
    283         data_source.name = self.name

TypeError: __init__() got an unexpected keyword argument 'storage_options'
```

martindurant:
Fixes #399

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -369,6 +369,12 @@ should specify this.  Further parameters
 may be necessary for communicating with the storage backend and, by convention, the driver should take
 a parameter ``storage_options`` containing arguments to pass to the backend.
 
+The special template variable "CATALOG_DIR" may be used to construct relative URLs in the arguments to
+a source. In such cases, if the filesystem used to load that catalog contained arguments, then
+the ``storage_options`` of that file system will be extracted and passed to the source. Therefore, all
+sources which can accept general URLs (beyond just local paths) must make sure to accept this
+argument.
+
 As an example of using ``storage_options``, the following
 two sources would allow for reading CSV data from S3 and GCS backends without
 authentication (anonymous access), respectively
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -383,12 +383,14 @@ authentication (anonymous access), respectively
 
    sources:
      s3_csv:
+       driver: csv
        description: "Publicly accessible CSV data on S3; requires s3fs"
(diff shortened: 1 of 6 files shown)


martindurant:
@ian-r-rose , since we're chatting ( :-) ), any idea why the new release of numpy seeks to have broken NPySource?

ian-r-rose:
Thanks for the quick fix @martindurant, I've confirmed that this fixes the issue seen in `intake_geopandas`. Do you recommend any `storage_options` related updates to `intake_geopandas`?

ian-r-rose:
@martindurant, unfortunately, I don't have any idea what's going on with NPySource

martindurant:
> Do you recommend any storage_options related updates to intake_geopandas?

None, to keep the current behaviour. If you want to be able to accept general file-like objects, it would be nice to follow the pattern but, as you said, if everything is handled by GDAL, then you don't have that option.

PR #403, opened 2019-08-12
update to docs to distinguish between a catalog and catalog file

gwbischof:


docs/source/glossary.rst
```diff
@@ -19,6 +19,9 @@ Glossary
         server or another third-party data service, like a SQL database. Thus, catalogs form a hierarchy: any
         catalog can contain other, nested catalogs.
 
+    Catalog file
+        A specification file which contains a list of named entries describing how to load data sources.
```
martindurant: Would you mind specifying that this is in YAML format and link to the docs where that syntax is described?

gwbischof:
I think this part of docs would be a bit clearer if `Catalog` was replaced by `Catalog file`
https://intake.readthedocs.io/en/latest/quickstart.html#opening-a-catalog

docs/source/glossary.rst
gwbischof: Sure i'll add that now

gwbischof:
Also fixed a broken link to CSVSource here: https://intake.readthedocs.io/en/latest/quickstart.html#opening-a-catalog

martindurant:
OK, that's great, thank you.

Issue #404, opened 2019-08-13
use entrypoints for catalog discovery


gwbischof:
In addition to using entrypoints for driver discovery, if we use entrypoints for catalog discovery we could also pip install catalogs in addition to using conda.

martindurant:
That is a *great* idea. Would the target of each entry point be a URL, or a function/constant within the package?

gwbischof:
I'm not sure of the best way to do that. I'll investigate a bit. Do you have an opinion?

martindurant:
I imagine the main use case will be to point to material contained within the package, so a function or constant would be fine (perhaps distinguished by parentheses). It would be nice to allow URLs too, perhaps by requiring that they use ":" in the string

danielballan:
I like this idea too.

> Would the target of each entry point be a URL, or a function/constant within the package?

Is the “target” here the right side of the equals sign in the entry point definition? If so, I believe is is restricted to refer to an object in a package, not just any string. It might be natural to standardize that every object in the ‘intake.catalogs’ entry point should be a Catalog instance, leaving the details of how that instance is constructed up to the package.

martindurant:
Yes, that's what I meant by target
Ok, that does make sense - then there's only one way to declare a catalog, but of course, within the code itself, it can refer to anything that Intake can handle, even a bare Catalog instance with entries attached dynamically.

gwbischof:
I agree with @danielballan that string targets aren't a feature of entrypoints. Other than registering URLs, there may be some nice uses of registering other strings, like API keys. What if we make a new package, that behaves like entrypoints but it is for strings (maybe JSON?) Maybe it has an encryption feature?  Then we could add it later as a second way to discover catalogs?

danielballan:
> Other than registering URLs, there may be some nice uses of registering other strings, like API keys.

I usually think of API keys as something to be distributed per _user_, not per package, and the lifecycle for a key might not line up with the release cycle for a package.

> What if we make a new package, that behaves like entrypoints but it is for strings (maybe JSON?) 

Adding other things like entrypoints is an interesting thought. Note that it would take more than writing a new package to achieve it. The `entry_points` keyword has special meaning to `setuptools` and in fact [has been elevated to a formal part of the Python packaging specification](https://twitter.com/ncoghlan_dev/status/1154548054524977152?s=20).

martindurant:
> Note that it would take more than writing a new package to achieve it.

[...]

PR #406, opened 2019-08-13
minor fixes to the docs

gwbischof:


martindurant:
Actually https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.open_files

martindurant:
Thanks for these changes! 
You should note the conversation in https://github.com/intake/intake/issues/404 to provide an alternative, non conda-specific way to "install data".

Issue #407, opened 2019-08-14
LocalFileSystem has no attribute 'storage_options'


gwbischof:
LocalCatalogEntry tries to access self._filesystem.storage_options.
self.filesystem is a fsspec.implementations.local.LocalFileSystem.
fsspec LocalFileSystem does not have this attribute `storage_options'
I think the fix is to pin fsspec >= 0.3.6.

```../../anaconda3/envs/garrett2/lib/python3.7/site-packages/intake/catalog/entry.py:78: in __call__
    s = self.get(**kwargs)
../../anaconda3/envs/garrett2/lib/python3.7/site-packages/intake/catalog/local.py:282: in get
    plugin, open_args = self._create_open_args(user_parameters)

self = name: xyz
container: catalog
plugin: ['bluesky-jsonl-catalog']
description: Some imaginary beamline
direct_access: for...amline: 00-ID
args: 
  paths: /tmp/tmp9oda5p8s/data/*.jsonl
  handler_registry: 
    NPY_SEQ: ophyd.sim.NumpySeqHandler
user_parameters = {}

    def _create_open_args(self, user_parameters):
        plugin = user_parameters.pop('plugin', None)
    
        md = self._metadata.copy() if self._metadata is not None else {}
        md['catalog_dir'] = self._catalog_dir
        if user_parameters.pop('cache', None) or self._cache:
            md['cache'] = user_parameters.pop('cache', None) or self._cache
        params = {'metadata': md,
                  'CATALOG_DIR': self._catalog_dir,
                  }
        params.update(self._open_args)
        if ('storage_options' not in params
                and self._filesystem is not None
                and self._filesystem.storage_options
                and _has_catalog_dir(params)):
E               AttributeError: 'LocalFileSystem' object has no attribute 'storage_options'```

PR #408, opened 2019-08-14
fsspec dependancy

gwbischof:
Fixes https://github.com/intake/intake/issues/407


martindurant:
Yep, that's fair. You may also want to put a PR to intake-feedstock.

PR #402, opened 2019-08-09
Use styles

martindurant:
ref #397

PR #413, opened 2019-09-11
Use a generator for pagination.

danielballan:
The `walk()` method returns a list. The `items()` method, added in
#317, returns a generator.

Tested on a large catalog (70k entries) this change makes a big
difference to overall performance, as one would expect.

### DIFF
--- a/intake/cli/server/server.py
+++ b/intake/cli/server/server.py
@@ -105,7 +105,7 @@ class ServerInfoHandler(tornado.web.RequestHandler):
             else:
                 start = int(page_offset)
                 stop = int(page_offset) + int(page_size)
-            page = itertools.islice(cat.walk(depth=1).items(), start, stop)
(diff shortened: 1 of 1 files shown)


martindurant:
Yeah, OK, makes sense.

martindurant:
PS: Would like to hear about this catalog of 70k entries!

danielballan:
Yeah! We've made good progress here... we have finished refactoring our legacy project, databroker, on top of intake. All our old tests are passing. Now we're seeing how it fares on our 4+ years of experimental data.

image

We now think of databroker an intake "distribution" --- when our users `pip install --pre databroker` they get intake, relevant intake drivers, and perhaps in the future relevant intake _catalogs_ aimed at facility-maintained intake-servers.  The long-promised blog post is in progress...

martindurant:
Would love to see some text and images describing this in #402 !

PR #414, opened 2019-09-11
Refine entrypoints docs

danielballan:
This addresses a point of confusion raised in #405 and a couple other potential points of confusion. It also fixes a crucial typo in the example.

### DIFF
--- a/docs/source/making-plugins.rst
+++ b/docs/source/making-plugins.rst
@@ -220,7 +220,7 @@ should add the following to the package's ``setup.py``:
        ...
        entry_points={
            'intake.drivers': [
-               'some_format_name = some_package.and_maybe_a_submodule.YourDriverClass',
+               'some_format_name = some_package.and_maybe_a_submodule:YourDriverClass',
                ...
            ]
        },
--- a/docs/source/making-plugins.rst
+++ b/docs/source/making-plugins.rst
@@ -226,6 +226,19 @@ should add the following to the package's ``setup.py``:
        },
    )
 
+.. important::
+
+   Some critical details of Python's entrypoints feature:
+
+   * Note the unusual syntax of the entrypoints. Each item is given as one long
+     string, with the ``=`` as part of the string. Modules are separated by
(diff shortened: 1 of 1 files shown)


docs/source/making-plugins.rst
```diff
@@ -220,12 +220,25 @@ should add the following to the package's ``setup.py``:
        ...
        entry_points={
            'intake.drivers': [
-               'some_format_name = some_package.and_maybe_a_submodule.YourDriverClass',
+               'some_format_name = some_package.and_maybe_a_submodule:YourDriverClass',
```
martindurant: I actually think we may want to support both, since the classes references, e.g., in `driver:` tags use dots only.

danielballan: I would be happier if it were dots only, but we don't have a choice. The entrypoints syntax is what it is, at the Python level. I don't think there is any way to make this work with the colon, short of a PR to CPython.

martindurant: It *looks* just like a string - what happens if you don't have a ":"?

danielballan: That would result in the package containing `entry_points.txt` that breaks the specification, so you would get errors if you tried to parse it with `importlib` [which uses a regex that includes a colon](https://github.com/python/cpython/blob/17499d82702432955d8e442a1871ff276ca64bc5/Lib/importlib/metadata.py#L46-L59) or with `entrypoints` [which does the same](https://github.com/takluyver/entrypoints/blob/82d8113ace6fb43a5b82bd14b51986b3683ecd6d/entrypoints.py#L21-L27).

martindurant: I see...

danielballan: Reading that first link, I saw the motivation for the colon. It enables referencing attributes on objects `module.submodule:object.attribute_on_object` while being explicit about which part of that path refers to modules and which part to attributes on an object in a module.

martindurant: Which is not our situation, but if there is no choice, then so be it.

PR #416, opened 2019-09-18
Don't replace "intake" in remote server name

martindurant:


martindurant:
That would require some DNS or at least /hosts configuration; could maybe be done with docker? In any case, too hard to do for now.

PR #419, opened 2019-09-26
don't check meta when merging dataframes

CJ-Wright:


CJ-Wright:
This depends on https://github.com/dask/dask/pull/5436

CJ-Wright:
@intake-dev ready for review!

martindurant:
Can you think of any case in which verifying is useful, i.e., should be an option for the user, with a default to not verify?
Should this wait for Dask release?

CJ-Wright:
It might make sense to define verify on a source by source basis, since some dataframes may want verification and some might not. A user override would also be good.

Any idea when the next Dask release will be?

martindurant:
> Any idea when the next Dask release will be?

Not as such, but minor releases often come every fortnight or so, depending on activity. It would be OK to merge here so long as we don't release before they do or - perhaps this is the right thing to do - explicitly check the Dask version to decide whether to use the kwarg or not.

CJ-Wright:
Would an `inspect` of the `from_delayed` signature be too heavy?

martindurant:
I would do `if loose_version(dask.__version__) <= loose_version("..."): no kwarg`

CJ-Wright:
Are there tests for remote catalogs? Can I add a test for this?

martindurant:
Sure - they are under intake/catalog/tests/test_remote_integration, and some more specific tests of the server in intake/cli/server/tests

CJ-Wright:
Would you accept this without a test?

martindurant:
The effect of not using the flag is to create more work, but the same result? Then yes - but I do think the Dask version check should be there, so that updating Intake does not necessarily require latest Dask (even though, in practice, most will be on latest).

martindurant:
Ah, you already did this. Assuming you can confirm that what I say above is correct, I'll merge.

CJ-Wright:
Yes, although there are some corner cases where setting the flag to True will cause errors, since the metadata across the dataframe chunks are not the same. If we put the flag into the sources themselves this should work out, since the source should know if the dataframe metadata is consistent.

martindurant:
OK .Since this is remote, the schema should be always consistent. Merging this now.

CJ-Wright:
Thank you very much!

PR #421, opened 2019-10-01
fix dask version constraint

CJ-Wright:


CJ-Wright:
@martindurant ready for review

PR #424, opened 2019-10-07
Catalog discovery with entrypoints

gwbischof:
Fixes https://github.com/intake/intake/issues/404

@danielballan worked on this with me.


martindurant:
@gwbischof , let us know if you need help here

gwbischof:
@martindurant thanks, any feedback so far? I want to think more about nesting vs flattening the merged catalogs before I change it from a draft PR. I'll add some updates to the docs too.

martindurant:
Sorry, I haven't had the chance yet. I intend to come back to it! I wonder what you think about whether all the potential catalogs should be merged into the top-level intake.cat or not...

martindurant:
@talebzeghmi , @danielballan any thoughts here?

talebzeghmi:
Some documentation on how to register a Catalog with an entry point would be helpful.  This would allow us to register non `YAMLFileCatalog` to the global Intake catalog.

danielballan:
@martindurant I think the behavior proposed in this PR is right: it makes sense to put all the sources listed in any YAMLs on the search path or any entrypoints at top level. This is how intake catalog discovery works currently (minus the entrypoints, of course).

I admit that initially I was against the flattening, but now I think it makes sense. How the package was distributed is an implementation detail. That is, the name shouldn't change if it comes from a YAML file vs a package. We have been trying this out in databroker 1.0.0beta, and it the usage seems nice.

danielballan:
I rebased on the `master` and got the tests passing (locally, at least, TBD if Travis-CI agrees). Are we ready to move forward with this?

danielballan:
Test failures are related to

```
AttributeError: 'DataFrame' object has no attribute 'to_msgpack'
```

caused by pandas choosing to [remove msgpack support](https://pandas.pydata.org/pandas-docs/version/1.0.0/user_guide/io.html?highlight=to_msgpack#msgpack) in order to guide users toward Arrow.

martindurant:
Please merge from master, this should be good to go.

danielballan:
Rebased, CI passes as expected. Thanks for the nudge and the :+1:, @martindurant. Nice to have this in!

martindurant:
Hurray, thank you everyone

PR #427, opened 2019-10-21
add search key function

### LINKED ISSUE #426: Advanced search options
Should there be more advanced search options? Would it be possible to incorporate something like [mongoquery](https://github.com/kapouille/mongoquery)? Or use elasticsearch for the metadata?

CJ-Wright:
Closes #426

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -202,6 +202,8 @@ class Catalog(DataSource):
         elif hasattr(key, '__call__'):
             entries = {k: copy.copy(v) for k, v in self.walk(depth=depth).items()
                        if key(v)}
+        else:
+            raise TypeError("Expected either string or callable")
         cat = Catalog.from_dict(
             entries, name=self.name + "_search",
             ttl=self.ttl,
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -196,7 +196,7 @@ class Catalog(DataSource):
         import copy
         if hasattr(key, '__call__'):
             entries = {k: copy.copy(v) for k, v in self.walk(depth=depth).items()
-                       if key(v)}
+                       if key(v.describe())}
         else:
             words = key.lower().split()
             entries = {k: copy.copy(v) for k, v in self.walk(depth=depth).items()
@@ -260,7 +260,7 @@ class Catalog(DataSource):
         out = sofar if sofar is not None else {}
         prefix = [] if prefix is None else prefix
         for name, item in self._entries.items():
-            if item._container == 'catalog' and depth > 1:
+            if getattr(item, '_container', None) == 'catalog' and depth > 1:
                 # recurse with default open parameters
                 try:
                     item().walk(out, prefix + [name], depth-1)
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@martindurant ready for review!

martindurant:
Please bear with me, my connectivity is very poor (thank you Rogers, inc)

CJ-Wright:
Friendly bump 😄

martindurant:
Might I ask for some kind of test, please?

CJ-Wright:
So this seems to have issues with remote catalogs, since we need to send the function to the remote catalog and have it run the function. Is there any nice way to do this (currently msgpack is trying and failing to serialize a function). Would it be acceptable to send a function in the first place? I guess we could use cloudpickle.

martindurant:
No, we should in no way send a function (a server should not execute arbitrary code). Either the function should run locally on the list of entries, or the remote defines its own search method (e.g., the mongo case for intake-bluesky).

CJ-Wright:
What is the best way to make catalogs from entries in memory?

martindurant:
`intake.catalog.Catalog.from_dict` is best

CJ-Wright:
Ahhh, it seems that doesn't work for remote catalogs, since remote catalogs get their catalogs via `_make_entries_container` to make `Entries`

martindurant:
Sorry, I don't understand what you want to do

CJ-Wright:
My approach was to:
1. Pull the entry metadata from the server
2. Run the search function on the metadata to filter the entries
3. Make a new `RemoteCatalog` from those entries that passed the filter

However it seems that `RemoteCatalog.from_dict` doesn't produce the needed `Entries`.

martindurant:
What happens if you make an ordinary Catalog (not remote) from the set of RemoteCatalogEntries which pass the filter?

CJ-Wright:
That works, although the return type of `RemoteCatalog.search` would then depend on if you pass in a function or a string.

martindurant:
Right, corresponding to whether the search is done locally or remotely. The default text version could also be done locally, but it requires fetching all of the entries, which may not be desirable.

CJ-Wright:
I guess the APIs are similar enough that one shouldn't build something that behaves differently for a remote or local catalog.

intake/catalog/base.py
```diff
@@ -717,6 +730,14 @@ def _load(self):
                  for source in info['sources']})
 
     def search(self, *args, **kwargs):
+        # if callable do special callable thing
+        if hasattr(args[0], '__call__'):
+            # check function against all of them
+            # TODO: run proper walk when walk doesn't take ages
+            entries = {k: self[k] for k, v in self._entries.items() if args[0](v.describe())}
```
CJ-Wright: This doesn't do walk like the local one. I left this out because `walk` was rather slow. I think this is because it is hitting the catalog refresh each time, maybe?

martindurant: I think that's OK: your function could filter whole catalogs in/out too.

CJ-Wright:
Should this be put under `filter` instead?

martindurant:
Ah, exactly! I knew I was repeating myself!

CJ-Wright:
Great! I'm going to close this PR and open a new one with some additional docs?

martindurant:
I don't mind whether you change tack here or make a new PR

CJ-Wright:
Sorry for all the noise!

PR #439, opened 2019-11-08
Make sure that ttl is numeric

### LINKED ISSUE #438: persist ttl argument should check that it's an int.
I just tried to use persist for the first time and passed a file name to the `ttl` argument (don't ask me why but I kind of thought `ttl` might be a terrible abbreviation of title). Passing a string to ttl did not through an error, but when I try to call `is_persisted` on the source, I am getting a comparison failure. 

As part of this, might be nice to add to the persist docstring.

jsignell:
Closes #438

### DIFF
--- a/intake/source/base.py
+++ b/intake/source/base.py
@@ -259,13 +259,26 @@ class DataSource(DictSerialiseMixin):
         return self.plot
 
     def persist(self, ttl=None, **kwargs):
-        """Save data from this source to local persistent storage"""
+        """Save data from this source to local persistent storage
+
+        Parameters
+        ----------
+        ttl: numeric, optional
+            Time to live in seconds. If provided, the original source will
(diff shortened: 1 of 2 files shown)


intake/source/base.py
```diff
@@ -259,13 +259,26 @@ def hvplot(self):
...
+        kargs: passed to the _persist method on the base container.
+        """
         from ..container import container_map
         from ..container.persist import PersistStore
         import time
         if 'original_tok' in self.metadata:
             raise ValueError('Cannot persist a source taken from the persist '
                              'store')
+        if ttl is not None and not isinstance(TTL, (int, float)):
+            raise ValueError('Cannot persist using a time to live that is '
```
martindurant: if you don't test this, coverage will go down :(

jsignell: Yes I am planning to add a test.

PR #442, opened 2019-11-26
add server cli kwarg for address

CJ-Wright:


martindurant:
Looks good.
Could use a test? Presumably setting this value to some arbitrary value would make the server refuse to respond.

CJ-Wright:
Are there any tests for the cli? Or do you mean a test to make certain that it is a valid address on the inside of the cli?

martindurant:
Yes, there are tests of the server, launched via the CLI, with either config of passed arguments. Actually, this value could do well with being int he config too.

CJ-Wright:
I'm sorry, the only references I see to the `port` option are in the conftest and the server's main module. Is there some other place I should be looking?

martindurant:
I would have thought there are, but I could be wrong. Yes, port was an obvious thing to check, so maybe a CLI test for that could be made at the same time?

CJ-Wright:
@martindurant I think this is ready for review!

intake/cli/server/tests/test_server.py
```diff
@@ -254,3 +254,55 @@ def test_flatten_flag(multi_server):
...
+            try:
+                yield 'intake://localhost:5001'
+            finally:
+                P.terminate()
+                P.wait()
+                shutil.rmtree(tmpdir)
+            break
+        except:
+            time.sleep(0.2)
+            assert time.time() - t < 10
```
martindurant: Move shutdown to a finally: block for the outer try? That way a hung server will get shut down too.

intake/cli/server/__main__.py
```diff
@@ -39,6 +39,9 @@ def main(argv=None):
                         help='Name of catalog YAML file')
     parser.add_argument('--flatten', dest='flatten', action='store_true')
     parser.add_argument('--no-flatten', dest='flatten', action='store_false')
+    parser.add_argument('-a', '--address', type=str, 
+                        default=conf.get('address', 'localhost'),
+                        help='hosting address')
```
martindurant: Could be a bit more verbose here? Yes, I realise the other options are also very, er, concise.
Also, you might consider including this in the docs pages.

CJ-Wright: It is in the docs pages, but I can make this more verbose.

intake/cli/server/tests/test_server.py
CJ-Wright: I think we can move it out completely, we don't need a finally since the try except will break out anyway.

CJ-Wright:
@martindurant that took longer than I'd like, but I think this is ready for another round of review!

intake/cli/server/__main__.py
```diff
@@ -39,6 +39,10 @@ def main(argv=None):
                         help='Name of catalog YAML file')
     parser.add_argument('--flatten', dest='flatten', action='store_true')
     parser.add_argument('--no-flatten', dest='flatten', action='store_false')
+    parser.add_argument('-a', '--address', type=str, 
+                        default=conf.get('address', 'localhost'),
+                        help='address to use as a host, defaults to the address '
+                        'in the configuration file, if provide otherwise localhost')
```
martindurant: small typo

CJ-Wright:
```suggestion
                        'in the configuration file, if provided otherwise localhost')
```

docs/source/tools.rst
```diff
@@ -62,18 +62,31 @@ You can see the full description of the server command with:
...
+    -p PORT, --port PORT  port number for server to listen on
+    --list-entries        list catalog entries at startup
+    --sys-exit-on-sigterm
+                          internal flag used during unit testing to ensure
+                          .coverage file is written
+    --flatten
+    --no-flatten
+    -a ADDRESS, --address ADDRESS
+                          address to use as a host, defaults to the address in
+                          the configuration file, if provide otherwise localhost
```
CJ-Wright:
```suggestion
                          the configuration file, if provided otherwise localhost
```

CJ-Wright:
@martindurant typo fixed

CJ-Wright:
Thank you very much!

Issue #444, opened 2019-12-07
GUI plot preview does not support full range of hvplot options (specifically dynamic plots)


jmosbacher:
I love this package! I came across this small annoyance and found a simple solution, tested it on a forked version and thought I would share it if anyone had experienced the same issue.

### Issue:
The current implementation of the GUI plot preview creates a HoloViews pane in the setup method:
` self.pane = pn.pane.HoloViews(self._plot_object(self.selected)) `
and then replaces the "object" attribute of the pane when a new plpot is selected:
`self.pane.object = self._plot_object(event.new)`
This requires that the new plot conforms to a limited subset of the plots returned by hvplot. By using the "by" and "groupby" keywords in the hvplot api, a dynamic plot is returned which does not display correctly (only the first item is displayed with no control widgets).

### Suggested solution
Use the Panel suggested method for dynamic layouts as shown in the gallery example [here](https://panel.pyviz.org/gallery/dynamic/dynamic_plot_layout.html). 
Setup should wrap the pane in a layout pane such as Column and use the panel helper function to automatically select the correct pane class:
`self.pane = pn.Column(pn.panel(self._plot_object(self.selected)))`
and replace its child with a new pane every time a the selected plot is changed:
`self.pane[0] = pn.panel(self._plot_object(event.new))`
This will perhaps effect the performance a bit but I think its worth it for getting full compatibility with hvplot.

PR #445, opened 2019-12-07
Allow for hvplot returning dynamic layouts

jmosbacher:
This addresses issue #444.

### DIFF
--- a/intake/gui/source/defined_plots.py
+++ b/intake/gui/source/defined_plots.py
@@ -54,7 +54,7 @@ class DefinedPlots(BaseView):
         self.select = pn.widgets.Select(options=self.options, height=30,
                                         align='center', width=200)
         self.desc = pn.pane.Str()
-        self.pane = pn.Column(pn.panel(self._plot_object(self.selected)))
+        self.pane = pn.Column(self._plot_object(self.selected))
         self.show_desc = pn.widgets.Checkbox(value=False,
                                              width_policy='min',
                                              align='center')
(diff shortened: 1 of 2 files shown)


intake/gui/source/defined_plots.py
```diff
@@ -54,7 +54,7 @@ def setup(self):
         self.select = pn.widgets.Select(options=self.options, height=30,
                                         align='center', width=200)
         self.desc = pn.pane.Str()
-        self.pane = pn.pane.HoloViews(self._plot_object(self.selected))
+        self.pane = pn.Column(pn.panel(self._plot_object(self.selected)))
```
philippjfr: Layouts will call pn.panel automatically if needed:

```suggestion
        self.pane = pn.Column(self._plot_object(self.selected))
```

philippjfr: I'd probably also rename `pane` -> `panel` throughout.

philippjfr:
Thanks for these changes. It looks like tests also need to be updated. Otherwise this looks good to me.

jmosbacher:
Yeah sorry I forgot to update the tests. I'll do that asap. Thanks

philippjfr:
@martindurant This seems ready to merge to me.

jmosbacher:
@philippjfr would you suggest already finding a better name than "pane" for the plot display in this PR or leave that for another time? since panel is already taken in this class, i guess plot_preview or plot_display would make sense. Alternatively, i guess i could make a DefinedPlot class with its own panel attribute that also inherits  from BaseView.

martindurant:
Leave it as pane, I think - too many names already. Just so long as we are consistent.

Issue #449, opened 2019-12-15
RemoteArray fails to initialize without chunks keyword


jmosbacher:
### Issue description
The default value for chunks is (-1,) 
```
self.chunks = tuple(tuple(c)
                            for c in tuple(kwargs.get('chunks', ((-1, ),))))
```
which causes a type error since tuple() is called on each item in the chunks
`TypeError: 'int' object is not iterable`
### Proposed solution
Change default value to nested tuple ((-1,),)

PR #450, opened 2019-12-15
Fix remote ndarray chunks

jmosbacher:
This addresses issue #449 .

### DIFF
--- a/intake/container/ndarray.py
+++ b/intake/container/ndarray.py
@@ -22,7 +22,7 @@ class RemoteArray(RemoteSource):
         self.metadata = kwargs['metadata']
         self.dtype = kwargs['dtype']
         self.chunks = tuple(tuple(c)
-                            for c in tuple(kwargs.get('chunks', (-1, ))))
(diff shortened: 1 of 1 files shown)


martindurant:
Would be nice to have a test that demonstrates the problem, and that it is now solved - presumably based on however you found this!

jmosbacher:
Funny, I was just thinking I should have done that. Ill add it to the PR as soon as I can (a bit busy lately).

martindurant:
ping here

jmosbacher:
Sorry, I have been super busy at work lately and havent had any time to write a test for this. Been pushing it off since i cant imagine many people suffering from this tiny bug but will definitely get around to it at some point. I need to build a test that includes a custom data source that doesnt send any "chunks" field in the schema and then set up a server etc., from what i saw in existing tests it should be pretty straight forward but will take some tinkering on my part (ie time).

martindurant:
This issue appears to have dropped into a dark hole :|

PR #452, opened 2019-12-24
Use join instead of "+"

### LINKED ISSUE #453: Extra "/" preventing exploration of more than one "subdirectory"
I'm getting a double slash introduced when I navigate using the intake gui to explore an s3 bucket, which then does not show the next level:

If I manually change `//` to `/` it works.

rsignell-usgs:
Fixes problem drilling more than one directory deep

Fixes #453

### DIFF
--- a/intake/gui/catalog/add.py
+++ b/intake/gui/catalog/add.py
@@ -144,7 +144,7 @@ class FileSelector(Base):
                 fn = event.new[0]
                 if fn.endswith('/'):
                     if self.path_text.value:
-                        self.path_text.value = self.path_text.value + '/' + fn
+                        self.path_text.value = os.path.join(self.path_text.value, fn)
                     else:
                         self.path_text.value = fn
                     self.make_options()


intake/gui/catalog/add.py
```diff
@@ -144,7 +144,7 @@ def move_down(self, *events):
                 fn = event.new[0]
                 if fn.endswith('/'):
                     if self.path_text.value:
-                        self.path_text.value = self.path_text.value + '/' + fn
+                        self.path_text.value = os.path.join(self.path_text.value, fn)
```
martindurant: This will fail on Windows, where the separator is "\\". Perhaps `posixpath.join`?

rsignell-usgs: @martindurant, isn't this the situation?  This is what I get on Windows:
```python
a = 's3://bucket/key1/'
b = 'key2/'
print(os.path.join(a,b))

's3://bucket/key1/key2/'
```

martindurant: What if `a = 'bucket'` ?

rsignell-usgs: Can that be the case?  I thought the GUI puts a "/" on the end.

martindurant: I suppose you are right, if the path ends with "/", it amounts to the same thing. This is a surprise to me! I never work with windows paths for a reason.

PR #455, opened 2020-01-13
tokenize performance

gwbischof:
This was a joint effort from @danielballan, @tacaswell, and myself.

We converted `dict` of the `DictSerializeMixin` to `OrderedDicts`, so `dask.base.tokenize()` doesn't sort them.  This change gave us a significant performance improvement for our use case.

When dask tokenizes a plain `dict` it sorts the keys, and it turns out that this sort operation dominates the call time, even for very small `dicts`. Using an  `OrderedDict` steers dask toward a different and faster tokenization.


martindurant:
Looks like you'll have to add a parser to YAML for ordered dicts like the [tuple](https://github.com/intake/intake/blob/master/intake/utils.py#L43) one

tacaswell:
This is a minor API change in that `cat(a=1, b=2)` and `cat(b=2, a=1)` will now tokenize differently, however I think that this is the correct behavior. As of py36 the order of iteration on kwargs is the same as the user passed in and it is plausible that some classes may rely on order for resolving the precedence of overlapping kwargs (ex color, edgecolor, facecolor).

martindurant:
^ if that was directed at me, my comment was about how to get the tests to pass, since at some points, the dicts underwent YAML deserialisation. Now the dicts are ordered, and if you want YAML to make sense of that, you need to tell it what to do. **Or** you convert to/from dicts before handing to YAML.

tacaswell:
@martindurant It was meant to be a general comment about the consequences of this change and should be documented someplace.

I could also see someone arguing that `cat(a=1, b=2)` and `cat(b=2, a=1)` should be interchangeable (because from python <3.6 they were from a technical POV and practically the number of classes that are going to make use of the kwarg order is going to be (vanishingly?) small) and must tokenize the same.  Better to have that discussion now rather than in 6mo when you get a bug report about it ;)

danielballan:
Thanks!

PR #462, opened 2020-02-10
Use pyarrow for dataframe serialization

### LINKED ISSUE #460: Incompatible with pandas 1.0.0
Pandas chose to [remove support for msgpack](https://pandas.pydata.org/pandas-docs/version/1.0.0/user_guide/io.html?highlight=to_msgpack#msgpack) in order to guide users toward Arrow instead. This breaks this codepath in intake

https://github.com/intake/intake/blob/33a096721765fc7fe79e958d3aa5f050a4c60937/intake/container/serializer.py#L56-L57

which now raises an `AttributeError` because `Data.Frame.to_msgpack` no longer exists.

I'm not immediately sure what the right fix is here, but maybe we should take up the suggestion in the TODO comment in this same function and transition to relying on `distributed`.

https://github.com/intake/intake/blob/33a096721765fc7fe79e958d3aa5f050a4c60937/intake/container/serializer.py#L48-L49

TomAugspurger:
Closes #460

### DIFF
--- a/intake/cli/server/server.py
+++ b/intake/cli/server/server.py
@@ -353,8 +353,6 @@ class ServerSourceHandler(tornado.web.RequestHandler):
             msg = dict(format=chunk_encoder.format_name,
                        compression=chunk_encoder.compressor_name,
                        container=source.container, data=data)
-            # Is this double serializing / copying?
-            # We're passing a Buffer object here.
             self.write(msgpack.packb(msg, **pack_kwargs))
             self.flush()
             self._cache.touch(source_id)  # keep source alive
--- a/intake/cli/server/server.py
+++ b/intake/cli/server/server.py
@@ -353,6 +353,8 @@ class ServerSourceHandler(tornado.web.RequestHandler):
             msg = dict(format=chunk_encoder.format_name,
                        compression=chunk_encoder.compressor_name,
                        container=source.container, data=data)
+            # Is this double serializing / copying?
+            # We're passing a Buffer object here.
             self.write(msgpack.packb(msg, **pack_kwargs))
             self.flush()
             self._cache.touch(source_id)  # keep source alive
--- a/intake/container/serializer.py
+++ b/intake/container/serializer.py
@@ -44,6 +44,15 @@ except ImportError:
     msgpack_numpy = None
 
 
+def check_pyarrow():
+    try:
+        import pyarrow
+    except ImportError:
+        raise ImportError("Serializing DataFrames requires pyarrow.")
+    return pyarrow
+
+
+
 class MsgPackSerializer(object):
     # TODO: This is ugly, should maybe transition to
     #  distributed.protocol.serialize
@@ -54,7 +63,10 @@ class MsgPackSerializer(object):
         if container in ['ndarray', 'xarray'] and msgpack_numpy:
             return msgpack.packb(obj, **np_pack_kwargs)
         elif container == 'dataframe':
-            return obj.to_msgpack()
+            pa = check_pyarrow()
+
+            context = pa.default_serialization_context()
+            return memoryview(context.serialize(obj).to_buffer())
         else:
             return msgpack.packb(obj, **pack_kwargs)
 
@@ -64,8 +76,9 @@ class MsgPackSerializer(object):
             from ..compat import np_unpack_kwargs
             return msgpack.unpackb(bytestr, **np_unpack_kwargs)
         elif container == 'dataframe':
-            import pandas as pd
-            return pd.read_msgpack(bytestr)
+            pa = check_pyarrow()
+            context = pa.default_serialization_context()
+            return context.deserialize(bytestr)
         else:
             return msgpack.unpackb(bytestr, **unpack_kwargs)
 
--- a/intake/container/serializer.py
+++ b/intake/container/serializer.py
@@ -63,9 +63,14 @@ class MsgPackSerializer(object):
         if container in ['ndarray', 'xarray'] and msgpack_numpy:
             return msgpack.packb(obj, **np_pack_kwargs)
         elif container == 'dataframe':
+            # Use pyarrow for serializing DataFrames, rather than
+            # msgpack: https://github.com/intake/intake/issues/460
             pa = check_pyarrow()
 
             context = pa.default_serialization_context()
+            # This eventually goes to msgpack.packb, which doesn't
+            # directly accept PyArrow Buffer objects. Need to wrap
+            # it in a memoryview to avoid a TypeError.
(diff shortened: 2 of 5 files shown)


martindurant:
That was quick :)

intake/cli/server/server.py
```diff
@@ -353,6 +353,8 @@ def post(self):
             msg = dict(format=chunk_encoder.format_name,
                        compression=chunk_encoder.compressor_name,
                        container=source.container, data=data)
+            # Is this double serializing / copying?
+            # We're passing a Buffer object here.
```
martindurant: I imagine msgpack doesn't do anything to the bytestring here (except perhaps copy).
It would be a larger discussion, perhaps worth having, whether this framing is useful at all. It would be far more efficient to do `context.serialize(obj).write_to(self)` here - both sides have already agreed that this is a piece of dataframe.

TomAugspurger: I could use some help understanding the flow here. At this point, I *think* data has already been serialized. I got an error about msgpack not being able to serialize a pyarrow Buffer object (which I worked around by wrapping it in a memoryview).

But that makes me think that we've already serialized the data, and are potentially serializing it again? Does that make sense?

intake/container/serializer.py
```diff
@@ -44,6 +44,15 @@ def decompress(self, data):
     msgpack_numpy = None
 
 
+def check_pyarrow():
+    try:
+        import pyarrow
+    except ImportError:
+        raise ImportError("Serializing DataFrames requires pyarrow.")
```
TomAugspurger: No tests for this yet.

setup.py
```diff
@@ -15,7 +15,7 @@
 extras_require = {
   'server': ['tornado', 'python-snappy'],
   'plot': ['hvplot', 'panel >= 0.5.1'],
-  'dataframe': ['dask[dataframe]', 'msgpack-numpy'],
+  'dataframe': ['dask[dataframe]', 'msgpack-numpy', 'pyarrow'],
```
TomAugspurger: Is this the right extras?

intake/container/serializer.py
```diff
@@ -54,7 +63,10 @@ def encode(self, obj, container):
         if container in ['ndarray', 'xarray'] and msgpack_numpy:
             return msgpack.packb(obj, **np_pack_kwargs)
         elif container == 'dataframe':
-            return obj.to_msgpack()
+            pa = check_pyarrow()
+
+            context = pa.default_serialization_context()
+            return memoryview(context.serialize(obj).to_buffer())
```
martindurant: Does `memoryview` do anything here?

TomAugspurger: See https://github.com/intake/intake/pull/462#discussion_r377119345. It's a workaround for

```python
In [5]: a = pa.array([1, 2, 3])

In [9]: msgpack.packb(a.buffers()[0])
TypeError                                 Traceback (most recent call last)
<ipython-input-9-08194270ba6f> in <module>
----> 1 msgpack.packb(a.buffers()[0])

~/miniconda3/envs/pyarrow=0.16.0/lib/python3.8/site-packages/msgpack/__init__.py in packb(o, **kwargs)
     44     See :class:`Packer` for options.
     45     """
---> 46     return Packer(**kwargs).pack(o)
     47
     48

msgpack/_packer.pyx in msgpack._cmsgpack.Packer.pack()

msgpack/_packer.pyx in msgpack._cmsgpack.Packer.pack()

msgpack/_packer.pyx in msgpack._cmsgpack.Packer.pack()

msgpack/_packer.pyx in msgpack._cmsgpack.Packer._pack()

TypeError: can not serialize 'pyarrow.lib.Buffer' object

In [10]: msgpack.packb(memoryview(a.buffers()[0]))
Out[10]: b'\xa1\x07'

```

intake/cli/server/server.py
martindurant: Right, we are serialising a few times: dataframe to bytes (arrow buffer, which is just a buffer), then to msgpack (which will contain that bytestring plus other stuff) and then to JSON to send over the connection. The latter two steps don't require much processing, but will be new memory copies.

setup.py
martindurant: I believe yes

martindurant:
The failing test suggests that pip pyarrow and/or pandas is not up to date with conda?

TomAugspurger:
I'm not sure. https://travis-ci.org/intake/intake/jobs/648446981#L396 says we got pyarrow 0.16. Trying it out locally.

TomAugspurger:
It seems to work fine locally using pyarrow from PyPI. Will look more later.

TomAugspurger:
I updated the TravisCI image from "xenial" to "bionic", which bumped the version of Python from 3.7.1 to 3.7.5. This fixed the pyarrow import issues.

Not sure about the CI failure

```
____________________________ test_display_init_gui _____________________________

    def test_display_init_gui():

        pytest.importorskip('panel')

        import intake

>       assert repr(intake.gui).startswith('Row')

intake/gui/tests/test_init_gui.py:36: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

intake/gui/__init__.py:59: in __repr__

    return repr(self._instance)

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[RuntimeError('Initialisation of GUI failed, even though panel is installed. Please update it to a more recent version (`conda install -c conda-forge panel==0.5.1`).') raised in repr()] GUI object at 0x7fe75e22cb90>

    def __repr__(self):

>       raise RuntimeError("Initialisation of GUI failed, even though "

                           "panel is installed. Please update it "

                           "to a more recent version (`conda install -c"

                           " conda-forge panel==0.5.1`).")
```

martindurant:
^ @philippjfr , are you aware of anything changing in a recent panel release that might have this effect? (pip only, but we install a different set of packages)

philippjfr:
No, but I can look into it in a minute.

philippjfr:
For now you could pin panel<0.8. I'll finish off the GUI plotting work as soon as I get a chance and will remove that pin at that point.

martindurant:
Great.

intake/container/serializer.py
danielballan: :+1: msgpack's type dispatch is very intolerant of duck types, has lots of `isinstance` checks.

PR #467, opened 2020-02-27
Improve UserParameter __repr__.

### LINKED ISSUE #465: Better repr for user parameters
Currently user parameters render like: 

```python
[<intake.catalog.local.UserParameter at 0x7f7072960d90>,
 <intake.catalog.local.UserParameter at 0x7f70723c7e80>]
```

It'd be nice if they were a dict, but if that is too hard, at least the repr for the object should contain the name.

danielballan:
In #255 we wanted a UserParameter repr that was more useful but not too
verbose. We settled on making `__str__` a complete eval-able
representation but leaving `__repr__` as the not-so-descriptive `object`
defaut.

I propose a different compromise: unify `__str__` and `__repr__` but use
a succinct non-eval-able repr, `<UserParameter 'some_name_here'>` where,
as in the standard library `<>` is used to indicate the the repr is not
eval-able but merely a summary.

Reasoning:

* An eval-able representation is nice, but our current implementation of
eval-able `__str__` is backward: normally it's the `__repr__` that is
(where possible) eval-able. If we agree that the eval-able string is too
verbose, maybe it would be better to settle on short but still useful
repr.
* Python _supports_ different `__str__` and `__repr__` but to my taste
it's nice when they can be consistent. When presented with a
not-so-useful `__repr__`, it doesn't always occur to me to try printing
it to see if the `__str__` representation is better.

Closes #465

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -79,14 +79,10 @@ class UserParameter(DictSerialiseMixin):
             self.allowed = [coerce(self.type, item)
                             for item in self.allowed]
 
-    def __str__(self):
-        return ('UserParameter(name={self.name!r}, '
-                'description={self.description!r}, '
-                'type={self.type!r}, '
-                'default={self.default!r}, '
-                'min={self.min!r}, '
-                'max={self.max!r}, '
-                'allowed={self.allowed!r})'.format(self=self))
+    def __repr__(self):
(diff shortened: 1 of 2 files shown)


martindurant:
OK, fine - we do, after all, also have the `describe()` method to get the details.

Issue #477, opened 2020-03-06
Address usage of deprecated 'imp' module.


danielballan:
This usage is deprecated and generates warnings in the tests.

https://github.com/intake/intake/blob/3678714200774efe0bd725b8d8b00f4743ed8dd7/intake/source/discovery.py#L301-L305

A downside of this approach is it bypasses the import machinery's normal caching mechanism, such that if the same plugin is loaded multiple times by intake, each load produces a new `type` instance even though they represent the same Python file and should be equivalent. (Maybe this is why `imp` is deprecated and that method is now un-documented. I'm not up to speed on Python import internals.)

We could look for a cleaner way to do this, or evaluate whether we want to do it at all.

danielballan:
Referring to #478, this:

```py
register_driver('thing', A)
register_driver('thing', A)  # OK, this is redundant but harmless
```

only works if `A` and `A` are literally the same type. `imp.load_source` causes trouble because it exec's the same file twice and produces distinct types, similar to this issue:

```py
# example.py
class A:
    ...
```

```py
In [1]: import example                                                                                                                                                                        

In [2]: a = example.A()                                                                                                                                                                       

In [3]: isinstance(a, example.A)                                                                                                                                                              
Out[3]: True

In [4]: from imp import reload                                                                                                                                                                
/home/dallan/venv/bnl/bin/ipython:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  #!/home/dallan/venv/bnl/bin/python3

In [5]: reload(example)                                                                                                                                                                       
Out[5]: <module 'example' from '/home/dallan/Repos/bnl/intake/example.py'>

In [6]: isinstance(a, example.A)                                                                                                                                                              
Out[6]: False
```

danielballan:
(Fully addressed by #478, for the record.)

PR #478, opened 2020-03-06
Do driver registration through a function.

danielballan:
Suggested by @HarinarayanKrishnan

Goals:
* Make the answer to "How do I register a driver at run time?" easier to discover.
* Ensure that two different code bases do not silently step on each other. This can cause trouble:
  ```py
  intake.registry['thing'] = A
  # ... later, perhaps in a different package...
  intake.registry['thing'] = B  # silently changed the registry under us!


This PR attempts to address both of these by introducing a new function `intake.register_driver` and laying a path toward making `intake.registry` an immutable view in the future. It works like this, a nice pattern that I have employed a couple times before to make shared registries safer and avoid confusing bugs.

```py
register_driver('thing', A)
register_driver('thing', A)  # OK, this is redundant but harmless
register_driver('thing', B)  # raises ValueError
register_driver('thing', B, overwrite= True)  # OK, since you have acknowledged that you are stepping on something
```

The code has been updated to use `register_driver` internally, never modifying `intake.registry` directly. Track is laid to make `intake.registry` an immutable view in the future; however, other libraries are likely modifying it directly now, so to start mutation is _support_ but issues a `DeprecationWarning`.

Unfortunately, as noted in code comments, we have to use `overwrite=True` internally in some situations. See #477. But this is should help avoid libraries accidentally stepping on each other and many cases.

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -19,7 +19,7 @@ from .. import __version__
 from .base import Catalog, DataSource
 from . import exceptions
 from .entry import CatalogEntry
-from ..source import registry as global_registry
+from ..source import register_driver
 from ..source import get_plugin_class
 from ..source.discovery import load_plugins_from_module
 from .utils import (expand_defaults, coerce, COERCION_RULES, merge_pars,
@@ -512,7 +512,7 @@ def register_plugin_module(mod):
         if k:
             if isinstance(k, (list, tuple)):
                 k = k[0]
-            global_registry[k] = v
+            register_driver(k, v)
 
 
 def register_plugin_dir(path):
@@ -521,7 +521,7 @@ def register_plugin_dir(path):
     for f in glob.glob(path + '/*.py'):
         for k, v in load_plugins_from_module(f).items():
             if k:
-                global_registry[k] = v
+                register_driver(k, v)
 
 
 def get_dir(path):
@@ -844,5 +844,8 @@ class EntrypointsCatalog(Catalog):
                 warings.warn(f"Failed to load {name}, {entrypoint}, {e!r}.")
 
 
-global_registry['yaml_file_cat'] = YAMLFileCatalog
-global_registry['yaml_files_cat'] = YAMLFilesCatalog
+# Register these early in the import process to support the default catalog
+# which is built at import time. (Without this, 'yaml_file_cat' is looked for
+# in intake.registry before the registry has been populated.)
+register_driver('yaml_file_cat', YAMLFileCatalog)
+register_driver('yaml_files_cat', YAMLFilesCatalog)
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -512,7 +512,11 @@ def register_plugin_module(mod):
         if k:
             if isinstance(k, (list, tuple)):
                 k = k[0]
-            register_driver(k, v)
+            # load_plugins_from_module can employ imp.load_source in which case
+            # we will re-exec the code and get a different type each time, so
+            # we have to tolerate overwriting here.
+            uses_load_source = mod.endswith('.py')
+            register_driver(k, v, overwrite=uses_load_source)
 
 
 def register_plugin_dir(path):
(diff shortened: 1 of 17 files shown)


intake/source/__init__.py
```diff
@@ -8,8 +8,60 @@
...
+        if original is driver:
+            return
+        raise ValueError(
+            f"The driver {driver} could not be registered for the "
+            f"name {name} because {_registry[name]} is already "
+            f"registered for that name. Use overwrite=True to force it.")
+    _registry[name] = driver
+
+
+def unregister_driver(name):
```
danielballan: I am not sure whether we actually need or want this one. Letting a package silently _remove_ a driver from underneath us isn't as likely to cause confusing bugs as silently _change_ a driver from underneath us, but it's still not great and I'm not sure there is a use case for this. Maybe `register_driver` with `overwrite` is enough.

martindurant: It is useful, at least for debugging. I don't expect too many end-users to need it, but I don't think it hurts.

martindurant:
On (re)importing source, we could do something a little shady like 
```
exec("import {modname}; cls = {modname}.{clsname}")
```
I'm not sure how you protect from code injection here (require only [a-zA-Z0-9.]?), but it doesn't incur multiple import costs, since it will use the module cache.

intake/source/__init__.py
danielballan: Ok, that works for me.

danielballan:
I believe that in all situations where `exec("import {modname}")` would work, `importlib.import_module(modname)` would also work. The thing is, intake currently supports exec-ing `.py` files at arbitrary locations outside of the `sys.path`, where neither of the approaches mentioned here would work.

danielballan:
The easier thing to do here would be to stop supporting that. Are you aware of an important use case?

martindurant:
Probably no reason to support it, agreed; other tricks like modifying path are not appetising.

On March 9, 2020 3:29:29 PM EDT, Dan Allan <notifications@github.com> wrote:
>The easier thing to do here would be to stop supporting that. Are you
>aware of an important use case?
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/intake/intake/pull/478#issuecomment-596736721

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

intake/catalog/local.py
```diff
@@ -512,7 +512,11 @@ def register_plugin_module(mod):
         if k:
             if isinstance(k, (list, tuple)):
                 k = k[0]
-            global_registry[k] = v
+            # load_plugins_from_module can employ imp.load_source in which case
+            # we will re-exec the code and get a different type each time, so
+            # we have to tolerate overwriting here.
+            uses_load_source = mod.endswith('.py')
```
martindurant: Yep, suggest raising an error for this case and dropping in `register_driver`. In hindsight, it's quite a security hole to automatically execute code just because you opened a catalogue, right? Also, it won't work for remote catalogues.

martindurant:
Any thoughts on `intake.container.container_map`?

danielballan:
Sure, might as well address `container_map` in the same PR. On it....

intake/catalog/local.py
danielballan: Both good points. Will do.

danielballan: Done in 6e9b2b1c2110b3bfa55a95bb5496b43d82b29e84

danielballan:
Some panel-related errors have appeared that seem unrelated to this PR. Was there a release recently?

danielballan:
Reproduced panel test-related issues locally and fixed. Summary:

- [x] Made `intake.registry` warn if directly mutated, with a path to making it immutable in a couple releases. Added `intake.register_driver` and `intake.unregister_driver`, 
- [x] Same for `intake.container.container_map`, `intake.container.register_container`, `intake.container.unregister_container`.
- [x] Removed ability to execute arbitrary Python scripts when a catalog is opened. If someone tries to use this feature, a specific error is raised explaining that this has intentionally changed and what to do installed. Closes #477; we no longer user the deprecated `imp` module.

martindurant:
intake-xarray is the only place where we need register_container?

I'm very happy to see this go in. Any last comments @jsignell or anyone else?

jsignell:
This all looks very sane and like a nice cleanup.

PR #479, opened 2020-03-13
Deprecate intake.Catalog, de-emphasise YAML files

### LINKED ISSUE #471: Unify catalog read
At the moment there is a `intake.Catalog` alias for `intake.open_catalog` and there is also `intake.load_catalog`. It'd be nice to at least consolidate around one approach in the docs. Additionally, it'd be nice to get rid of the top level `Catalog` since it is confusing that it is not a class.

martindurant:
Fixes #471


martindurant:
(a precursor to updating the docs as per the devs meeting)

PR #483, opened 2020-04-01
Data entrypoint fix

martindurant:


martindurant:
(also started docs here, will continue separately)

PR #490, opened 2020-04-29
WIP: Let users never see Entry objects

martindurant:
This is "unification light" - still have Entry objects, but the user doesn't see them. The catalog just gives you the default source directly, but you can override them with call or clone (maybe should be get??) to get a new version of the source. Note that `.describe()` will give you the original catalog definition (with user parameters), not your overridden version, but repr of the source now also gives you that YAML view with the current set of arguments.

cc @danielballan @tacaswell


martindurant:
(I'm sure a load of tests would need updating, but would like comments sooner rather than later)

intake/source/base.py
```diff
@@ -215,6 +222,27 @@ def to_spark(self):
...
+    def clone(self, **kwargs):
+        return self.entry(**kwargs)
+
+    def __call__(self, *args, **kwargs):
+        return self.clone(**kwargs)
+
+    def describe(self):
+        return self.entry.describe()
+
+    @property
```
tacaswell, line 242: This seems like an application layer thing that should not be baked into the core objects.  If this is something that was there and needs to stay for back compatibility can we deprecate it?

intake/source/base.py
```diff
@@ -215,6 +222,27 @@ def to_spark(self):
...
+        if self._entry is None:
+            raise ValueError("Source was not made by a catalog")
+        if self._entry() is None:
+            raise RuntimeError("Tried to recover entry, but it was gone")
+        return self._entry()
+
+    def clone(self, **kwargs):
+        return self.entry(**kwargs)
+
+    def __call__(self, *args, **kwargs):
```
tacaswell, line 236: I assume this needs to be here for back compatibility?  If so, can we deprecate it in favor of `clone`?

intake/source/base.py
```diff
@@ -215,6 +222,27 @@ def to_spark(self):
...
+    @property
+    def entry(self):
+        if self._entry is None:
+            raise ValueError("Source was not made by a catalog")
+        if self._entry() is None:
+            raise RuntimeError("Tried to recover entry, but it was gone")
+        return self._entry()
+
+    def clone(self, **kwargs):
+        return self.entry(**kwargs)
```
tacaswell, line 234: should this return `self.entry(**kwargs)()` so we get a `Datasource` back?

intake/source/base.py
martindurant: Right, you could construct a GUI for an entry previously (which allowed you to select the user parameters and then plot with point&click); so if the user is no longer exposed to the entry object, this is the only way to do it. Of course, we could question the usefulness of that particular GUI view, but otherwise the source no longer has any connection to the GUIs at all.

intake/source/base.py
martindurant: Yes, this is how it was done before: `entry() -> default source`, `entry(args) -> specific_source`. It could warn indeed, but this is exactly the conversation I wanted to have: what would feel the simplest for a user.
Also, it should maybe just return `self`, not a copy, if the kwargs match the kwargs self was made with.

intake/source/base.py
martindurant: "calling" the entry makes a data source :)
I know, I know....

intake/source/base.py
```diff
@@ -215,6 +222,27 @@ def to_spark(self):
...
 
+    @property
+    def entry(self):
+        if self._entry is None:
+            raise ValueError("Source was not made by a catalog")
+        if self._entry() is None:
+            raise RuntimeError("Tried to recover entry, but it was gone")
+        return self._entry()
+
+    def clone(self, **kwargs):
```
martindurant, line 233: What should the behaviour be if there is no entry, e.g., if this source was created with `intake.open_*`? Right now it will error, but I suppose it should make a new source anyway.

intake/catalog/entry.py
```diff
@@ -80,11 +81,12 @@ def __call__(self, persist=None, **kwargs):
...
-                return s2
-            if met['ttl'] < time.time() - met['timestamp']:
-                return s2
+                s = s2
+            elif met['ttl'] < time.time() - met['timestamp']:
+                s = s2
             else:
-                return store.refresh(s2)
+                s = store.refresh(s2)
+        s._entry = weakref.ref(self)
```
tacaswell, line 89: Why store a weakref?  If we are going to require the entry to still be around to clone a `DataSource` (maybe with different kwargs), it seems like a trap to make that working contingent on something else holding a cache of entries.

intake/source/base.py
tacaswell: Sorry, I had lost track of the layer here :sheep:

intake/source/base.py
tacaswell: But that implies there is always a reliably way to determine if the kwargs are "equal" and probably an escape hatch to always return a new instance.  It seems like a lot of complexity to save a few keystrokes in a minority of cases.

intake/source/base.py
tacaswell: It makes a lot more sense to me to pass the `DateSrouce` to the gui (`make_the_clicky_gui(ds)`) rather than have it be a method so that it is possible to support exposing several gui generation tactics to the user (ex "all the details" vs "executive summary").

intake/source/base.py
tacaswell: In the worst case it could merge the kwargs it was created with with these kwargs (or just go with the defaults?) and return `type(self)(**{**my_kwargs, **kwargs})`?

I'm also not sure `clone` is the best name as one of the features of this is the ability to change it, however I don't have a better name.

tacaswell:
Sorry for crashing into your code base with possibly more opinions than understanding.

martindurant:
> Sorry for crashing into your code base with possibly more opinions than understanding.

It is appreciated!

intake/catalog/entry.py
martindurant: I wanted to avoid a circular reference, but actually the entry doesn't refer to the source, so this was unnecessary.

tacaswell: That makes sense.  I can see how it would be useful for the entry to hold on to weakrefs of the `DataSources`.

danielballan:
Thanks for leading the way on this, @martindurant! I am still under a rock but should emerge next week to look at this.

martindurant:
Good to go, reviews appreciated!

(docs still need updating in some places; may have knock-on errors in dependent packages which need sorting before blanket releases)

danielballan:
I'd like to beg another week or so to look at this. Racing to prep for a big tutorial a week from tomorrow (unveiling databroker-refactored-on-top-of-intake among other things).

martindurant:
@danielballan , I can wait :)

intake/source/base.py
```diff
@@ -215,6 +231,42 @@ def to_spark(self):
         """
         raise NotImplementedError
 
+    @property
+    def entry(self):
+        if self._entry is None:
+            raise NoEntry("Source was not made from a catalog entry")
+        return self._entry
+
+    def configure(self, **kwargs):
```
danielballan, line 240: This name strongly suggests to me that it is mutating the current instance, rather than returning a new instance. Could you live with `duplicate(...)`?

intake/source/base.py
danielballan: I too would like to revisit this, but I think should not lump that into this PR.

intake/source/base.py
danielballan: :+1: for always returning a new instance

intake/source/base.py
```diff
@@ -215,6 +231,42 @@ def to_spark(self):
         """
         raise NotImplementedError
 
+    @property
+    def entry(self):
+        if self._entry is None:
+            raise NoEntry("Source was not made from a catalog entry")
```
danielballan, line 237: Raise a subclass of `AttributeError` here is nice. :+1:

intake/utils.py
```diff
@@ -112,7 +112,7 @@ def __getstate__(self):
         # OrderedDict steers dask toward a different and faster tokenization.
         kwargs = collections.OrderedDict({k: arg.__getstate__()
                   if isinstance(arg, DictSerialiseMixin) else arg
-                  for k, arg in self._captured_init_kwargs.items()})
+                  for k, arg in sorted(self._captured_init_kwargs.items(), key=lambda x: x[0])})
```
tacaswell, line 115: I do not think we should sort here.  Python guarantees that kwarg order is maintained so there can be meaningful information in that order that we will be losing by sorting.

danielballan, line 115: Keyword argument order has meaning, so these should not be sorted. Two objects with `_captured_init_kwargs` given in a different order should tokenize uniquely. This was introduced here, for reference: https://github.com/intake/intake/pull/455#issuecomment-574143114

martindurant: I think this is here because in older pythons, the dict order was *not* fixed, so sources that ought to have been equal turned out not to be.

intake/catalog/entry.py
```diff
@@ -93,6 +94,15 @@ def _get_default_source(self):
             self._default_source = self()
         return self._default_source
 
+    @property
+    def container(self):
+        return getattr(self, '_container', None)
+
+    @container.setter
+    def container(self, cont):
+        # for compat with magis source attr
```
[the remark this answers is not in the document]
martindurant: (if I could spell...)
The entry object have a `._container` attribute that can be updated, so we get and set that and tie to `.container` (which is what they have always been called on a source) to make sure that whether code is trying to access the entry version or the source version, it sees the same thing.

intake/source/base.py
martindurant: Duplicate isn't right either, because the new one is not the same as the old one. 
Other suggestions? `change`, `copy_and_configure`? I don't yet know if the old `get`/`__call__` will be deprecated, so perhaps this is not the ones users will actually type.

intake/source/base.py
martindurant: The GUI code can probably do with a refactor following improvements in https://github.com/intake/filesystem_spec/blob/master/fsspec/gui.py#L14

intake/source/base.py
danielballan: That's true. I have the same issue with `change` that I have with `configure`. I think `copy_and_configure` is pretty good. Other ideas, maybe not better than `copy_and_configure`:

* `new_instance(...)`
* `variation(...)`
* `new_variation(...)`
* `with_params(...)`

intake/utils.py
danielballan: Isn't it better to have some cache misses than to have erroneous cache collisions? If we want to an optimized path for Python 3.5 and lower, we can sort conditional on `sys.version_info < (3, 6)`?

martindurant: Certainly agree that code shouldn't change just for the sake of what is testable...

danielballan:
This requires some small changes to databroker but I now have a branch https://github.com/bluesky/databroker/pull/569 that passes against this intake PR branch, intake master, and intake 0.5.5. I also took it for a spin interactively, and it behaves as expected.

Remaining concerns:

- [ ] There is a back-compatibility issue with `.get()`. See this example using `data-us-states`. With intake 0.5.5:

  ```py
  In [3]: intake.cat.states.get()                                                                                                                                                               
  Out[3]: <intake.source.csv.CSVSource at 0x7ff9771b50d0>
   ```

  With this PR branch

  ```py
  In [3]:  intake.cat.states.get()                                 
  AttributeError                            Traceback (most recent call last)
  <ipython-input-5-e1dc9e0715d2> in <module>
  ----> 1 intake.cat.states.get()

  AttributeError: 'CSVSource' object has no attribute 'get'
  ```

- [ ] Settling on a name for `configure` (I am fine with `copy_and_configure`, or other suggestions listed in line above.)
- [ ] Removing `sorted` or making it conditional on Python version (again see comment above)

Thanks again @martindurant for taking this on.

intake/source/base.py
martindurant: `configure_new` ?

martindurant:
Last call for comments

martindurant:
(updates to docs will follow merger and release)

danielballan:
I'm seeing a regression from the last commit but I don't understand it yet. Please hold merging while I investigate.

danielballan:
Looks like it was just a confusion on my end, I must have had different branches installed than I thought. Full steam ahead!

danielballan:
Rebased to resolve conflicts and force-pushed

martindurant:
> Rebased to resolve conflicts and force-pushed

Did the same, but you were faster! 
```
To https://github.com/martindurant/intake
 ! [rejected]        unify -> unify (fetch first)
error: failed to push some refs to 'https://github.com/martindurant/intake'
```

danielballan:
Welp, looks like I introduced a syntax error in rebase....

martindurant:
fixed...

danielballan:
Test failure is on one build, GUI related.

martindurant:
You took the button press away from me! :)

PR #491, opened 2020-04-30
Zarr catalogs

### LINKED ISSUE #447: Drivers for zarr hierarchies
I would like to use intake to create a catalog of data stored using zarr. I know that there is already support for zarr via the intake-xarray plugin, however I have zarr data that does not conform to the xarray (CDF) data model, but I would still like to use intake.

I think what I would like to do is simply create a catalog where each entry represents an entire zarr hierarchy. E.g., currently to open a hierarchy we do something like:

```python
import gcsfs
import zarr
storage_path = 'ag1000g-release/phase2.AR1/variation/main/zarr/pass/ag1000g.phase2.ar1.pass'
gcs = gcsfs.GCSFileSystem(token='anon', access='read_only')
store = gcsfs.mapping.GCSMap(storage_path, gcs=gcs, check=False, create=False)
callset = zarr.open_consolidated(store)
callset
```

The resulting `callset` object is an instance of `zarr.hierarchy.Group`.

Instead I'd like to be able to do something like:

```python
import intake
catalog = intake.open_catalog('path/to/catalogs/ag1000g.yml')
callset = catalog.phase2_variation_main_pass.open()
```

...or something like that, and be returned `callset` again as an instance of `zarr.hierarchy.Group` exactly as before.

Any advice on the best approach greatly appreciated.

alimanfoo:
This PR adds an implementation of zarr groups as intake catalogs.

Resolves #447.

### DIFF
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -19,7 +19,7 @@ from .catalog.base import RemoteCatalog
 from .catalog import local
 from .catalog.default import load_combo_catalog
 from .catalog.local import MergedCatalog, EntrypointsCatalog
-from .catalog.zarr import ZarrCatalog
+from .catalog.zarr import ZarrGroupCatalog
 from .container import upload
 from .source import register_driver, registry
 from .source.discovery import autodiscover
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -19,6 +19,7 @@ from .catalog.base import RemoteCatalog
 from .catalog import local
 from .catalog.default import load_combo_catalog
 from .catalog.local import MergedCatalog, EntrypointsCatalog
+from .catalog.zarr import ZarrCatalog
 from .container import upload
 from .source import register_driver, registry
 from .source.discovery import autodiscover
--- a/intake/catalog/zarr.py
+++ b/intake/catalog/zarr.py
@@ -11,13 +11,13 @@ class ZarrGroupCatalog(Catalog):
     partition_access = None
     name = 'zarr_cat'
 
-    def __init__(self, url, storage_options=None, component=None, metadata=None,
+    def __init__(self, urlpath, storage_options=None, component=None, metadata=None,
                  consolidated=False):
         """
 
         Parameters
         ----------
-        url : str
+        urlpath : str
             Location of data file(s), possibly including protocol information
         storage_options : dict, optional
             Passed on to storage backend for remote files
@@ -30,7 +30,7 @@ class ZarrGroupCatalog(Catalog):
         consolidated : bool, optional
             If True, assume Zarr metadata has been consolidated.
         """
-        self._url = url
+        self._urlpath = urlpath
         self._storage_options = storage_options or {}
         self._component = component
         self._consolidated = consolidated
@@ -43,21 +43,21 @@ class ZarrGroupCatalog(Catalog):
         if self._grp is None:
 
             # obtain the zarr root group
-            if isinstance(self._url, zarr.hierarchy.Group):
+            if isinstance(self._urlpath, zarr.hierarchy.Group):
                 # use already-opened group, allows support for nested groups
                 # as catalogs
-                root = self._url
+                root = self._urlpath
 
             else:
 
                 # obtain store
-                if isinstance(self._url, str):
+                if isinstance(self._urlpath, str):
                     # open store from url
                     from fsspec import get_mapper
-                    store = get_mapper(self._url, **self._storage_options)
+                    store = get_mapper(self._urlpath, **self._storage_options)
                 else:
                     # assume store passed directly
-                    store = self._url
+                    store = self._urlpath
 
                 # open root group
                 if self._consolidated:
@@ -82,13 +82,13 @@ class ZarrGroupCatalog(Catalog):
                 entry = LocalCatalogEntry(name=k,
                                           description='',
                                           driver='ndzarr',
-                                          args=dict(url=v),
+                                          args=dict(urlpath=v),
                                           catalog=self)
             else:
                 entry = LocalCatalogEntry(name=k,
                                           description='',
                                           driver='zarr_cat',
-                                          args=dict(url=v))
+                                          args=dict(urlpath=v))
             entries[k] = entry
         self._entries = entries
 
--- a/intake/catalog/zarr.py
+++ b/intake/catalog/zarr.py
@@ -94,6 +94,3 @@ class ZarrGroupCatalog(Catalog):
 
     def to_zarr(self):
         return self._grp
-
-
-register_driver('zarr_cat', ZarrGroupCatalog)
--- a/intake/catalog/zarr.py
+++ b/intake/catalog/zarr.py
@@ -1,0 +1,58 @@
+from .base import Catalog
+from .local import LocalCatalogEntry
+from ..source import register_driver
+
+
+class ZarrCatalog(Catalog):
+    """Catalog as described by a Zarr group."""
+    version = '0.0.1'
+    container = 'catalog'
+    partition_access = None
+    name = 'zarr_cat'
+
+    def __init__(self, url, consolidated=False, **kwargs):
+        self._url = url
+        self._consolidated = consolidated
+        self._grp = None
+        super().__init__(**kwargs)
+
+    def _load(self):
+        import zarr
+
+        if self._grp is None:
+            # obtain the zarr group to wrap as a catalog
+            if isinstance(self._url, zarr.hierarchy.Group):
+                # use already-opened group
+                self._grp = self._url
+            else:
+                # open root group from url
+                from fsspec import get_mapper
+                store = get_mapper(self._url)
+                if self._consolidated:
+                    # use consolidated metadata
+                    self._grp = zarr.open_consolidated(store=store, mode='r')
+                else:
+                    self._grp = zarr.open_group(store=store, mode='r')
+
+        # build catalog entries
+        entries = {}
+        for k, v in self._grp.items():
+            if isinstance(v, zarr.core.Array):
+                entry = LocalCatalogEntry(name=k,
+                                          description='',
+                                          driver='ndzarr',
+                                          args=dict(url=v),
+                                          catalog=self)
+            else:
+                entry = LocalCatalogEntry(name=k,
+                                          description='',
+                                          driver='zarr_cat',
+                                          args=dict(url=v))
+            entries[k] = entry
+        self._entries = entries
+
+    def to_zarr(self):
+        return self._grp
+
+
+register_driver('zarr_cat', ZarrCatalog)
--- a/intake/catalog/zarr.py
+++ b/intake/catalog/zarr.py
@@ -3,36 +3,78 @@ from .local import LocalCatalogEntry
 from ..source import register_driver
 
 
-class ZarrCatalog(Catalog):
-    """Catalog as described by a Zarr group."""
+class ZarrGroupCatalog(Catalog):
+    """A catalog of the members of a Zarr group."""
+
     version = '0.0.1'
     container = 'catalog'
     partition_access = None
     name = 'zarr_cat'
 
-    def __init__(self, url, consolidated=False, **kwargs):
+    def __init__(self, url, storage_options=None, component=None, metadata=None,
+                 consolidated=False):
+        """
+
+        Parameters
+        ----------
+        url : str
+            Location of data file(s), possibly including protocol information
+        storage_options : dict, optional
+            Passed on to storage backend for remote files
+        component : str, optional
+            If None, build a catalog from the root group. If given, build the
+            catalog from the group at this location in the hierarchy.
+        metadata : dict, optional
+            Catalog metadata. If not provided, will be populated from Zarr
+            group attributes.
+        consolidated : bool, optional
+            If True, assume Zarr metadata has been consolidated.
+        """
         self._url = url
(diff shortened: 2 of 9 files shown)


alimanfoo:
This seems to work, I created a local catalog file "ag1000g.yml" with an entry to a zarr hierarchy we have stored in GCS:

```
metadata:
  version: 1
sources:
  phase2_ar1_variation_main_all:
    description: TODO
    driver: zarr_cat
    args:
      url: gs://ag1000g-release/phase2.AR1/variation/main/zarr/all/ag1000g.phase2.ar1
      consolidated: True
```

...then ran:

```
In [1]: import intake                                                         

In [2]: cat = intake.open_catalog('ag1000g.yml')                              

In [3]: callset = cat.phase2_ar1_variation_main_all                           

In [4]: callset                                                               
Out[4]: 
name: phase2_ar1_variation_main_all
container: catalog
plugin: ['zarr_cat']
description: TODO
direct_access: forbid
user_parameters: []
metadata: 
args: 
  url: gs://ag1000g-release/phase2.AR1/variation/main/zarr/all/ag1000g.phase2.ar1
  consolidated: True

In [5]: callset.to_zarr()                                                     
Out[5]: <zarr.hierarchy.Group '/' read-only>

In [6]: callset['2R']                                                         
Out[6]: 
name: 2R
container: catalog
plugin: ['zarr_cat']
description: 
direct_access: True
user_parameters: []
metadata: 
args: 
  url: <zarr.hierarchy.Group '/2R' read-only>

In [7]: callset['2R']['variants']['POS']                                      
Out[7]: 
name: POS
container: ndarray
plugin: ['ndzarr']
description: 
direct_access: True
user_parameters: []
metadata: 
args: 
  url: <zarr.core.Array '/2R/variants/POS' (24767689,) int32 read-only>

In [8]: callset['2R']['variants']['POS'].to_dask()                            
Out[8]: dask.array<from-zarr, shape=(24767689,), dtype=int32, chunksize=(4194304,), chunktype=numpy.ndarray>
```

martindurant:
Well *of course* a zarr group is like a catalog - I had not thought of this before! Indeed, this is also true for the hdf/cdf family too.

Presumably the metadata of each group or array should be taken from the related zarr attrs?

Note that for your example, the following syntax variants ought to work too:
```
callset['2R'].variants.POS  # (attribute access where the parts are valid as python identifiers)
callset['2R', 'variants', 'POS']
callset['2R.variants.POS']
```

intake/catalog/zarr.py
```diff
@@ -0,0 +1,58 @@
...
+                    self._grp = zarr.open_group(store=store, mode='r')
+
+        # build catalog entries
+        entries = {}
+        for k, v in self._grp.items():
+            if isinstance(v, zarr.core.Array):
+                entry = LocalCatalogEntry(name=k,
+                                          description='',
+                                          driver='ndzarr',
+                                          args=dict(url=v),
```
martindurant, line 44: We end up making a new mapper for each child that gets accessed, which may be slow for remote storage. It would be possible to pass on the group/array on the already-instantiated store, particularly for the group case in the next branch.

intake/__init__.py
```diff
@@ -19,6 +19,7 @@
 from .catalog import local
 from .catalog.default import load_combo_catalog
 from .catalog.local import MergedCatalog, EntrypointsCatalog
+from .catalog.zarr import ZarrCatalog
```
martindurant, line 22: Historically, all drivers core to Intake registered themselves at import and are accessible in the top-level. New drivers, though, usually in separate packages, can declare themselves with entrypoints in setup.py/toml, so they don't get imported until needed.

intake/catalog/zarr.py
```diff
@@ -0,0 +1,58 @@
...
+                                          catalog=self)
+            else:
+                entry = LocalCatalogEntry(name=k,
+                                          description='',
+                                          driver='zarr_cat',
+                                          args=dict(url=v))
+            entries[k] = entry
+        self._entries = entries
+
+    def to_zarr(self):
```
martindurant, line 54: We mildly discourage adding to the public API of the driver - it's not obvious to me what this gets you, unless the user wants to start writing to the dataset.

alimanfoo:
> Well _of course_ a zarr group is like a catalog - I had not thought of this before! Indeed, this is also true for the hdf/cdf family too.

Actually, it was [your idea](https://github.com/intake/intake/issues/447#issuecomment-564716187) :-)
 
> Presumably the metadata of each group or array should be taken from the related zarr attrs?

That makes sense to me.

> Note that for your example, the following syntax variants ought to work too:
> 
> ```
> callset['2R'].variants.POS  # (attribute access where the parts are valid as python identifiers)
> callset['2R', 'variants', 'POS']
> callset['2R.variants.POS']
> ```

Good to know, thanks. 

Aside, the h5py and zarr shortcut syntax is to use '/' as separator, e.g., "2R/variants/POS", I wonder if that would make sense for intake too.

martindurant:
> Actually, it was your idea

**So** long ago, in another reality!

> the h5py and zarr shortcut syntax is to use '/' as separator, e.g., "2R/variants/POS"

Certainly. Since you're doing getitem anyway, it might already work? It would take some effort to get ipython tab-completion, though.

alimanfoo:
@martindurant could you suggest a work plan here to complete this PR? Some bullet points for things I should add or change? Obviously needs some tests, but any other pointers appreciated.

martindurant:
@danielballan @jsignell @talebzeghmi @mmccarty @gwbischof - this is the first new driver proposed for the main repo in some time, certainly since the entrypoints work. Given that we already have the zarr array source, it makes sense to me; but we don't really have any policy around this. Any thoughts/concerns?

martindurant:
+ @tacaswell

intake/catalog/zarr.py
alimanfoo: Here the actual child zarr array or group (`v`) is passed through as the value of the `url` argument, which does work and means that the mapper is only used once at the root (I think!)

intake/__init__.py
alimanfoo: OK, I saw some autoregistration code but wasn't sure. Happy to do what you recommend.

intake/catalog/zarr.py
alimanfoo: So the use case is this:

```
import intake                                                         
cat = intake.open_catalog('ag1000g.yml')                              
callset = cat.phase2_ar1_variation_main_all.to_zarr()
```

...i.e., I just want to open the root zarr group and will work via the zarr API from there. 

Or more generally, I want to get the underlying zarr group that a zarr catalog is wrapping, and start working with it directly. So I need some public API for accessing the wrapped zarr group. Happy to use whatever you suggest.

martindurant:
@alimanfoo , seeing no objections please go ahead and fill out tests and docstrings. I am still doubtful about the to_zarr method, but I don't have a better alternative, since this doesn't match to_dask or read methods, and catalogues don't normally implement those anyway.

Please do add a entrypoint for the new driver and include it in the API documentation and list of known divers in plugin-directory.rst

alimanfoo:
Thanks @martindurant.

Re to_zarr() method, an alternative would just be to expose the `_grp` member via a public property, e.g., `group` or `zarr_group`. 

Btw the same requirement arises for the `ZarrArraySource` where a user (me) wants to get access to the underlying zarr array (currently stored in the `_arr` member). Either a `to_zarr()` method or a public property would be useful there too.

alimanfoo:
Hi @martindurant, I've added tests, docstrings and API docs.

I wasn't sure what this meant:

> Please do add a entrypoint for the new driver

Also zarr is not currently listed in requirements.txt or extras_require in setup.py, so to get tests running on travis python 3.7 I've edited the travis config to manually install zarr, but perhaps we should add it somewhere into extras_require - whatever you think best.

martindurant:
I mean an entry like https://github.com/intake/intake/blob/master/setup.py#L54
Then, you don't need to register your driver at all, it would be found whenever a source required it. This is the default way to declare intake drivers in a separate package; in this case, maybe it doesn't make any difference, but it would be good to put the entry in anyway for the sake of consistency.

alimanfoo:
> I mean an entry like https://github.com/intake/intake/blob/master/setup.py#L54

Thanks, done.

One other question, sources like csv and xarray's zarr source use an argument named `urlpath`, but the ZarrArraySource and this new ZarrGroupCatalog class use an argument named `url`.

Would you like to consistify both ZarrArraySource and ZarrGroupCatalog to use `urlpath` instead of `url`? Or keep as `url`?

martindurant:
I suppose a consistent name is useful

alimanfoo:
> I suppose a consistent name is useful

Just wondering if you know of any downstream dependencies on `ZarrArraySource` that might break if we change the argument there? (I think intake-xarray has it's own `ZarrSource`.)

martindurant:
Correct, intake-xarray has its own zarr stuff. You might find the RemoteXarray interesting, since it uses zarr to serialize the metadata, but overrides Dask's handling of reading chunks to fetch each piece from the server.
I am not aware of anyone using ZarrArraySource directly.

alimanfoo:
OK, I changed `url` to `urlpath` in both `ZarrGroupCatalog` and `ZarrArraySource`. Let me know if there's anything else I should do. 

I didn't see a change log, should I make a note of these changes somewhere?

martindurant:
I had been keeping a wiki page on the intake repo as a "news feed", as opposed to changelog, but it's not been touched in ages. Since we're heading towards a pretty big release, I should at least write a blog or maybe start a release notes/changelog page in the docs. No need to do in this PR, though.

alimanfoo:
Awesome, thanks @martindurant!

Btw do you have an ETA for when you plan to cut the next release? We'd like to start using this feature soon, but can install from master if you're planning to wait for a bit.

martindurant:
I'd like to get https://github.com/intake/intake/pull/490 in first, but we'll discuss at tomorrow's developers' meeting whether that's feasible and who has the time to help. That PR needs many updates to the tests, including your new ones here :)

joshmoore:
Thanks, both!  Here is the working catalog for the dataset discussed this evening:

```
metadata:
  version: 1
sources:
  idr0083:
    description: https://forum.image.sc/t/publication-of-idr0083-sars-cov-2-up-close-in-a-human-organoid/37163
    driver: zarr_cat
    args:
      urlpath: https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/9822151.zarr
      consolidated: False
```

```
In [4]: dict(intake.open_catalog('covid.yml').idr0083.to_zarr().attrs)
Out[4]:
{'multiscales': [{'datasets': [{'path': '0'},
    {'path': '1'},
    {'path': '2'},
    {'path': '3'},
...
```

:+1:

----

NB: @alimanfoo : just to save bums like me a moment or two, it might be worth updating your first example (https://github.com/intake/intake/pull/491#issuecomment-622125660) to use `urlpath` or to put the corrected example in the description.

tacaswell:
> Aside, the h5py and zarr shortcut syntax is to use '/' as separator, e.g., "2R/variants/POS", I wonder if that would make sense for intake too.

If intake is going to do implicit access through multiple catalog layers, I think `/` makes a lot more sense than `.`.  

It is a bit late but :+1: from me.  I'm also a bit skeptical of ``to_zarr` as well.  I see the benefit for wanting to get access to the underlying objects, but the way I think about intake is very coupled to abstracting away the underlying storage details.  I am glad someone else seems to be interested in mid-level catalogs interesting entities with their own metadata etc.

alimanfoo:
> NB: @alimanfoo : just to save bums like me a moment or two, it might be worth updating your first example ([#491 (comment)](https://github.com/intake/intake/pull/491#issuecomment-622125660)) to use `urlpath` or to put the corrected example in the description.

Done. Glad to know this work was useful for you too :smile:

alimanfoo:
> If intake is going to do implicit access through multiple catalog layers, I think `/` makes a lot more sense than `.`.

+1

We are very used to working through the zarr/h5py style API where "/" is used to navigate through multiple hierarchy layers.

alimanfoo:
> It is a bit late but  from me. I'm also a bit skeptical of ``to_zarr` as well. I see the benefit for wanting to get access to the underlying objects, but the way I think about intake is very coupled to abstracting away the underlying storage details.

If intake supported "/" as a path separator for traversing catalog layers then there would be less need for us to access the underlying zarr group directly. But I still expect there will be cases where we want to drop down to the zarr API, i.e., return the zarr group and work with it directly.

An interesting question might be, what would you do if intake also supported HDF5 files as catalogs? In that case also I would expect some users would just use intake as a convenient way of opening an HDF5 file, but then would want to get hold of the underlying h5py File or Group object and work with that directly.

martindurant:
@joshmoore , can you place your catalog definition somewhere online so that we can refer to it?

joshmoore:
@martindurant: sure. Now available under:

```
intake.open_catalog("https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/catalogs/idr0083.yml").image_9822151
```

though I'm not sure that that will remain be its permanent home.

martindurant:
That comes through nicely as a zarr_cat, but seems to have no children. Does it show for you? Then probably my current state of intake and/or zarr is broken.

joshmoore:
Ah, that might make sense. EBI's S3 currently does not allow ListObjectsV2. (To be fixed in the next months) So though I can see the items, you can't. Sorry about that.

PR #493, opened 2020-05-04
Fix catalog.save

### LINKED ISSUE #492: .save() functionality in Catalog for custom local catalog class
Found a number of errors in trying to save a catalog that was defined in a custom class. Source here: https://github.com/manfralab/intake-qcodes

Here's the solution I ended up with, including some comments about what errors needed to be fixed. Sorry, this isn't a super clean example...

```
    def serialize(self):
        """
        Produce YAML version of this catalog.
        Note that this is not the same as ``.yaml()``, which produces a YAML
        block referring to this catalog.
        """
        import yaml
        output = {"metadata": self.metadata, "sources": {},
                  "name": self.name}
        for key, entry in self.items():
            # hack to fix serializing the name of the custom catalog
            kw = entry._captured_init_kwargs.copy()
            kw.pop('catalog')

            # 'parameters' should be a dict, is currently being written as a list
            # this hack only works because there are no parameters defined in my catalog
            if isinstance(kw['parameters'], list):
                if not kw['parameters']:
                    kw['parameters'] = {}
                else:
                    raise ValueError('not sure how to serialize this correctly')

            if not kw['cache']:
                kw['cache'] = [] # default value of None raises an error when read back

            output["sources"][key] = kw

        return yaml.dump(output)
```

martindurant:
fixes #492 

Not sure I like this - could use a more general way for JSON/YAML serialisation, in the manner we have ensured for pickle.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -274,8 +274,8 @@ class Catalog(DataSource):
                   "name": self.name}
         for key, entry in self.items():
             kw = entry._captured_init_kwargs.copy()
-            kw.pop('catalog')
-            kw['parameters'] = {k.name: k.__getstate__()['kwargs'] for k in kw['parameters']}
+            kw.pop('catalog', None)
+            kw['parameters'] = {k.name: k.__getstate__()['kwargs'] for k in kw.get('parameters', [])}
             output["sources"][key] = kw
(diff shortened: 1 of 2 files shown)


martindurant:
(leaving this here a short while for alternative suggestions)

PR #494, opened 2020-05-04
POC: open by mimetype

martindurant:
Adds a function to open a URL by guessing the mimetype (or by providing the mimetype).

Deficiencies:
- many of the file types we care about don't have a mimetype, so we probably need our own extension guessing framework too
- ditto for compression
- assumes that the URL is the first and only required parameter to the driver class


martindurant:
For no reason at all, MIME can get complex fast: https://stackoverflow.com/a/42140178/3821154

(this is in case we have MIME as the input, perhaps given by the remote storage directly - there is no way to do this now)

danielballan:
Interesting!

PR #496, opened 2020-05-05
Suppress or fix warnings

tacaswell:
This addresses several things that warn on newer versions of python (invalid escape and `!=` vs `is`), suppresses a warning on import, ~and configures pytest to make sure the warnings don't come back~.

### DIFF
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -27,8 +27,10 @@ from .gui import InstanceMaker
 # Populate list of autodetected drivers (plugins).
 # The built-in ones are included here because intake itself declares them via
 # its own entry_points.
-for name, driver in autodiscover().items():
-    register_driver(name, driver)
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    for name, driver in autodiscover().items():
+        register_driver(name, driver)
 
 logger = logging.getLogger('intake')
 if sys.version_info >= (3, 7):
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -27,10 +27,8 @@ from .gui import InstanceMaker
 # Populate list of autodetected drivers (plugins).
 # The built-in ones are included here because intake itself declares them via
 # its own entry_points.
-with warnings.catch_warnings():
-    warnings.simplefilter("ignore")
-    for name, driver in autodiscover().items():
-        register_driver(name, driver)
+for name, driver in autodiscover().items():
+    register_driver(name, driver)
 
 logger = logging.getLogger('intake')
 if sys.version_info >= (3, 7):
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ jobs:
       install:
         - sudo apt update
         - sudo apt install libsnappy-dev
-        - pip install panel bokeh dask pytest
+        - pip install panel bokeh dask pytest zarr
         - pip install -e .[complete]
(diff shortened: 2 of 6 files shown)


tacaswell:
There is still

```
=============================== warnings summary ===============================
455/home/travis/virtualenv/python3.8.1/lib/python3.8/distutils/__init__.py:4
456  /home/travis/virtualenv/python3.8.1/lib/python3.8/distutils/__init__.py:4: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
457    import imp
458
459-- Docs: https://docs.pytest.org/en/latest/warnings.html
460================= 359 passed, 19 skipped, 1 warning in 58.42s ==================
```

which I have not tracked down how to fix (but I don't get locally).

intake/__init__.py
```diff
@@ -27,8 +27,10 @@
 # Populate list of autodetected drivers (plugins).
 # The built-in ones are included here because intake itself declares them via
 # its own entry_points.
-for name, driver in autodiscover().items():
-    register_driver(name, driver)
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    for name, driver in autodiscover().items():
+        register_driver(name, driver)
```
martindurant, line 33: What drivers were warning here? Emitting this warning is in our control, the idea was to remove this version of autodiscover (as opposed to entryoints) eventually.

tacaswell:
I turned off the fail on warnings to avoid having to track it down and just accepted the printing.

intake/__init__.py
tacaswell: The warning is coming from

https://github.com/intake/intake/blob/76cb2ff2772bf56720fcefcbde5e0e2eb75401a6/intake/source/discovery.py#L55-L60

and is independent of any particular driver.

martindurant: Hm, OK - probably should only warn if we actually find anything

tacaswell: I went with the minimal fix to prevent warning on import in down stream packages.

martindurant: I know, but this basically makes the warning always a no-op; would you mind removing the ignore, and only issue the warnings of the package scan returns anything?

tacaswell: This only skips the warning on import, if the user manually calls `autodiscover()` it will still warn them unless they explicitly pass `do_package_scan=False`.  It seems correct to me to warn the user if something _may_ have been found given their call even if it was not.

martindurant: The purpose of the warning is different: that the user has a package installed which uses the deprecated method to be found by intake. That *does* happen at install time, only; I never expected any user to call the function directly.

tacaswell: I see.  So the remediation is to switch the package to using entry points instead?

martindurant: Yes, this is supposed to get users to switch to a new version of the package containing the driver, and to get the dev of the package to update it to use entrypoints (not in that order!)

martindurant: Would you mind reinstating the warning (i.e., dropping the ignore), but making it trigger only if the package scan turns up something?

martindurant:
Thanks @tacaswell !

tacaswell:
Sorry I did not back to this, other work kept jumping the queue.

PR #498, opened 2020-05-23
Perf defer tokenize

tacaswell:
See the second commit messages for details, but by making this change (and by making sure we opt-out of persistence in our sub-classes) we are able to take the time it takes to init catalogs from ~5-10s to ~5us.

We are creating catalogs "on the fly" from data we already have in memory so all of the data payload is being injected via the arguments so we are just burning a tremendous amount of CPU upfront.

The way this was written we could not figure out any way to over-ride this behavior without compromising using `__getstate__` for pickle.

attn @danielballan @ronpandolfi

### DIFF
--- a/docs/source/persisting.rst
+++ b/docs/source/persisting.rst
@@ -95,7 +95,7 @@ For details on the methods of the persist store, see the API documentation:
 information about the sources they were made from, so that they can be remade
 successfully. This all appears in the source metadata.
 The sources use the "token" of the original
-data source as their key in the store, a value which can be found by ``source._tok``
+data source as their key in the store, a value which can be found by ``dask.base.tokenize(source)``
 for the original source, or can be taken from the metadata of a persisted source.
 
 Note that all of the information about persisted sources is held in a single YAML file in
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -129,10 +129,8 @@ class Catalog(DataSource):
         -------
         Catalog instance
         """
-        from dask.base import tokenize
         cat = cls(**kwargs)
         cat._entries = entries
-        cat._tok = tokenize(kwargs, entries)
         return cat
 
     @property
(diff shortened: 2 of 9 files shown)


martindurant:
It would be reasonable to move the computing of `_tok` to the property, i.e., so it is done on the first access, and not at all if it is never used.

So given @danielballan 's comment, what is the status here?

tacaswell:
Yup, it was easier to fail the tests via exception than check for warnings :facepalm: .

Given that accessing `obj._tok` was in the documentation I erred on the side of not breaking API with no deprecation, if you are not worried about that we can just remove it.

martindurant:
Always prefer the smooth path without depr warnings, which just annoy people. So I would keep the attribute as a computed property and then you can get the best of both.

martindurant:
... plus, the property allows us to cache the result, so that you don't get repeat calls to `tokenize` for the same instance.

danielballan:
Oops, I think @martindurant and @tacaswell are talking past each other.

> plus, the property allows us to cache the result, so that you don't get repeat calls to tokenize for the same instance

The caching will work regardless, via the new `__tok_cache` and `__dask_tokenize__`. The first time `tokenize` is called on an object, the token is computed and the second time the cache is used. The difference is that it's fully internal now; `tokenize` just returns faster after the first time.. This is easier to for downstream users/developers to get right because it uses existing dask machinery and does not introduce a special semi-private attribute (`_tok`) exclusive to intake objects.

Our choice is:

(1) Remove `_tok` and break things.
(2) Leave `_tok` for back-compat with a warning and update the documentation to use `tokenize`.
(3) Leave `_tok` for back-compat without a warning.

@tacaswell implemented (2) in this PR. I don't see any reason to do (1). @martindurant If you have a preference for (3) that's fine by me. Otherwise this is good to merge IMO.

martindurant:
OK, sorry - my head is now thoroughly wrapped around it :)

danielballan:
Surfacing a private comment here for the benefit of anyone following this later:

> To be clear, that PR did deprecate `_tok` and update the docs to remove mention of it. Back-compatibility is maintained with a warning.

PR #501, opened 2020-05-30
FIX: entry property behavior when no entry

tacaswell:
- Catalogs are DataSource sub-classes
 - they now have an entry property that raises NoEntry which is a
   AttributeError sub-class
 - in the Catalog base class there is  a __getattr__ defined which
   gets triggered if attribute access raises AttributeError
 - there is code that tries calling self.entry and catches the NoEntry
   to handle the failure case
 - but __getattr__ catches it first and converts it to a normal
   AttributeError which escapes the try/except in configure_new

This adds an extra special case to the __getattr__ code to raise the
expected attribute and deprecates obj.entry.  It is only used a few
places internally which were replaced with checking
`self._entry is None`.

### DIFF
--- a/intake/source/base.py
+++ b/intake/source/base.py
@@ -270,7 +270,10 @@ class DataSource(DictSerialiseMixin):
     @property
     def gui(self):
         """Source GUI, with parameter selection and plotting"""
-        return self.entry.gui
+        if self._entry is None:
+            raise NoEntry("Source was not made from a catalog entry")
+
+        return self._entry.gui
 
     def close(self):
         """Close open resources corresponding to this data source."""
--- a/intake/source/base.py
+++ b/intake/source/base.py
@@ -11,6 +11,7 @@ from yaml import dump
 
 from .cache import make_caches
 from ..utils import make_path_posix, DictSerialiseMixin, pretty_describe
+import warnings
 
 
 class Schema(dict):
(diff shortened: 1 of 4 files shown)


intake/source/base.py
```diff
@@ -248,19 +250,22 @@ def configure_new(self, **kwargs):
...
             kw = self._captured_init_kwargs.copy()
             kw.update(kwargs)
             return type(self)(*self._captured_init_args, **kw)
 
     __call__ = get = configure_new  # compatibility aliases
 
     def describe(self):
-        return self.entry.describe()
+        if self._entry is None:
+            raise NoEntry("Source was not made from a catalog entry")
```
martindurant, line 266: So is this really the *only* user method of the entry that would get used, aside from making more sources?

tacaswell:
It was also used it the `gui` property which is now fixed.

martindurant:
Knew there must be something else :)
Can you please add test for the gui attribute, then? It should skipif panel is not installed.

tacaswell:
This pretty show-stopping bug (as it is impossible to use catalogs that don't actually have an entry) so I would rather avoid scope creep here.

martindurant:
Done :)
Will merge on green. Thanks for finding this.
All good for release?

PR #502, opened 2020-06-03
MNT: clean up remaining leakage of CatalogEntry to public API

tacaswell:
We found another place Entry objects were leaking out to the user via `cat.items()`.  This PR makes `cat.items()` consistent with item and attribute access by returning `DataSource` objects and deals with the internal fallout.

Added a test.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -202,12 +202,20 @@ class Catalog(DataSource):
         return cat
 
     def filter(self, func):
-        """Create a Catalog of a subset of entries based on a condition
+        """
+        Create a Catalog of a subset of entries based on a condition
 
-        Note that, whatever specific class this is performed on, the return
-        instance is a Catalog. The entries are passed unmodified, so they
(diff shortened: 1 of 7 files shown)


tacaswell:
I'm not sure what is going on with the 3.6 build, I don't see those failures locally.

martindurant:
So the argument is, if cat[k] -> source, then cat.items() -> iter over sources. I get that.
Should we also have .values() too, then, to fill out the dict namespace?

I am interested, what broke for you?

martindurant:
@tacaswell , would you like me to finish this PR off?

tacaswell:
I am not likely to be able to get to this in the next few days, I am :100: :+1: on you taking over this PR if you want.

It was one of our databroker extensions that was doing `cat.items()`, trying to treat the value as a `DateSource` and failing with attribute error because `Entry` no longer does the attribute fallback.

tacaswell:
Thank you for finishing this @martindurant 8

PR #513, opened 2020-07-09
Cache default source

### LINKED ISSUE #512: Performance in loading nested catalogs
I'm seeing some performance regressions in the most recent release of intake (0.6.0), possibly related to #490.

A while back in #333, we had discussed how to avoid fetching catalog data (e.g., with a network request) too many times in the context of a nested catalog. #335 provided a fix, but that no longer seems to work. At present, dotting into nested catalogs seems to recreate the inner catalogs anew every time. I have some contexts where setting up a new catalog is somewhat expensive (tens of seconds), which makes this constant recreation very painful.

A minimal example of this is (adapted from #333):
```python
from time import sleep

import intake
import yaml

# Create a catalog that is slow to initialize
class MyCatalog(intake.catalog.Catalog):
    name = 'mycat'
    def _load(self):
        print('loading', end='...')
        sleep(2)
        print('loaded')

        self._entries = {}

# Write a nested catalog with a mycat entry
inner = MyCatalog(ttl = 200, name='inner')
with open('tmp.yml', 'w') as f:
    f.write(inner.yaml())

# Load the nested catalog, and dot-access the inner catalog several times
outer = intake.open_catalog('tmp.yml')
print("How many times is it loaded?")
for i in range(10):  
    inner = outer.inner
    inner = outer.inner
```

This is probably also related to #410 . If there are some best-practices around this that have recently changed in light of #490, I'd be interested in hearing them as well.

ian-r-rose:
Fixes #512 (at least, the part that doesn't involve lazy catalog loading).


martindurant:
I like "Somebody is a vim user." :)

We may come back to this if we decide that remote sources need this caching too, but since definitely helps and solves your use case, I'll merge. Thanks @ian-r-rose !

Issue #411, opened 2019-09-04
YAMLFilesCatalog not working with remote path


rafa-guedes:
I am trying to organise some intake catalog with some master config pointing to folders containing multiple configs. I'm using the `YAMLFilesCatalog` driver for that. This is working fine when the configs are in my local machine, but not when they are in a bucket.

I have some dummy setup to demonstrate this issue. The two configs (pasted at the bottom here) are in the bucket specified below, this should be able to be reproduced.

**Source *`fromfolder`* defines a sub-catalog in master pointing to a folder via `YAMLFilesCatalog`.**
```python
import intake

# Using local configs
cat = intake.Catalog("path/to/master-catalog.yml")
cat.fromfolder.some_dataset.to_dask() # works

# Using remote configs
cat = intake.Catalog("gs://my-intake/master-catalog.yml")
cat.fromfolder.some_dataset.to_dask() # fails

2019-09-04 17:17:36,031 - intake - INFO - local.py:_load:L764 - Loading "name: child-catalog
container: catalog
plugin: ['yaml_file_cat']
description: YAML file: child-catalog
direct_access: True
user_parameters: []
metadata: 
  catalog_dir: gs://my-intake
args: 
  name: gs://my-intake/folder/*
  ttl: 1
  path: my-intake/folder/child-catalog.yml" as a catalog failed: [Errno 2] No such file or directory: '/home/rguedes/my-intake/folder/child-catalog.yml'
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
```
When using YAMLFilesCatalog, `{{CATALOG_DIR}}` seems to be evaluating to the local directory in my computer instead of the remote path.

This approach works both in my local machine and in the bucket if I point to the child config directly using `YAMLFileCatalog` instead of `YAMLFilesCatalog`.

**Source *`fromfile`* defines a sub-catalog in master pointing to a folder via `YAMLFileCatalog`.**

```python
import intake

# Using local configs
cat = intake.Catalog("path/to/master-catalog.yml")
cat.fromfile.some_dataset.to_dask() # works

# Using remote configs
cat = intake.Catalog("gs://my-intake/master-catalog.yml")
cat.fromfile.some_dataset.to_dask() # works
```

## Master config

**./master-catalog.yml**
```
metadata:
  version: 1

description: Master catalog

sources:

  fromfile:
    description: Sub-catalog from single remote file
    driver: intake.catalog.local.YAMLFileCatalog
    args:
[...]

PR #514, opened 2020-07-13
fix for yamlfilescatalog with remote paths

j-r77:
Minor change; passing the filesystem attribute along when opening catalog files through a YamlFilesCatalog. This solves issue #411 (on my machine at least).

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -737,6 +737,7 @@ class YAMLFilesCatalog(Catalog):
                 '.yaml', '').replace('.yml', '')
             kwargs = self.kwargs.copy()
             kwargs['path'] = f.path
+            kwargs['fs'] = f.fs
             d = make_path_posix(os.path.dirname(f.path))
             if f.path not in self._cats:
                 entry = LocalCatalogEntry(name, "YAML file: %s" % name,


martindurant:
Thank you for your contribution!

Do you think it's possible to write a test case showing the original problem and this solution in action? Does it work if the path is a complete URL with protocol?

j-r77:
I have tested with the following catalog:

```
metadata:
  version: 1
sources:
  testyamlfiles:
    description: ''
    driver: intake.catalog.local.YAMLFilesCatalog
    args:
      path: "adl://devanalyticsstorage/intake/"
      storage_options: ***

  testlocalfiles:
    description: ''
    driver: intake.catalog.local.YAMLFilesCatalog
    args:
      path: "/home/persistent/testcat"

  testlocalfilesrelative:
    description: ''
    driver: intake.catalog.local.YAMLFilesCatalog
    args:
      path: "testcat"
	  
```
Each location contains the catalog1.yml and catalog2.yml from the tests/ folder.
localfiles and localfilesrelative always worked fine, the adl:// one used to drop the protocol and storage, looking instead for a local path:

```
print(list(cat.testyamlfiles()))
2020-07-13 14:10:02,855 - intake - INFO - local.py:_load:L750 - Loading "..." as a catalog failed: [Errno 2] No such file or directory: '/home/persistent/intake/catalog1.yml'
2020-07-13 14:10:02,864 - intake - INFO - local.py:_load:L750 - Loading "..." as a catalog failed: [Errno 2] No such file or directory: '/home/persistent/intake/catalog2.yml'

```
After the fix, this works:
```
print(list(cat.testyamlfiles()))
['ex1', 'ex2', 'ex3', 'ex4']
```
I assumed it would work for other remote storages than adl:// as well, but I don't immediately have a test setup ready for those. If I'm understanding the travis-ci results above correctly, it seems that my fix breaks the intake:// protocol though? Any suggestions on how to test what's going on there?

martindurant:
> it seems that my fix breaks the intake:// protocol

Correct, because localfilesystem, although it serialises ok with pickle, is not set up for msgpack. [This](https://github.com/msgpack/msgpack-python#packingunpacking-of-custom-data-type) kind of thing could be implemented via `fs..__reduce__()[1]`.

martindurant:
Ping @j-r77

Issue #515, opened 2020-07-15
intake_server and http headers with underscore


dylanmcreynolds:
I'm trying to setup intake server instance behind a load balancing firewall (Kubernetes Ingress). Found a very strange issue.

```
In [1]: from intake import RemoteCatalog                                                                                                                                             

In [2]: c = RemoteCatalog('intake://<hostname....>')                                                                                            

In [3]: list(c)                                                                                                                                                                      
Out[3]: ['root']

In [4]: list(c['root'])                                                                                                                                                               
Out[4]: ['root']

In [5]: list(c['root']['root'])                                                                                                                                                        
Out[5]: ['root']
```

After some sniffing of http headers, I found that a request was missing the 'source_id' header which I assume should have told the server what catalog to interrogate. Without it, the server always replies with the root catalog specified in the intake-server's yml file ('root' in the case above.)

This has to do with underscores in headers, which nginx does not pass by default. I tested and a header with 'source-id' would be passed fine. Sniffing around the web, this seems to be some old cgi related issue. I'm working and on a solution in my installation (I do not have access to the the obvious configuration in nginx that would fix this). It might be nice to consider changing the few custom headers that intake client and server pass around so that underscore is not being used.

bjoernenders:
FYI this is solved with
```
nginx.ingress.kubernetes.io/server-snippet: |
  underscores_in_headers on;
```
But this took a very long time to research. Better would be if intake were not using underscores *at all* in their headers. Why not use dashes instead and make every server happy?

martindurant:
Agreed, this can be a simple fix on our end

On July 14, 2020 8:43:40 PM EDT, Bjoern Enders <notifications@github.com> wrote:
>FYI this is solved with
>```
>nginx.ingress.kubernetes.io/server-snippet: |
>  underscores_in_headers on;
>```
>But this took a very long time to research. Better would be if intake
>were not using underscores *at all* in their headers. Why not use
>dashes instead and make every server happy?
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/intake/intake/issues/515#issuecomment-658481861

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

martindurant:
(you may well want to submit the PR if you have the time, you're likely to get to it before me)

dylanmcreynolds:
[...]

PR #516, opened 2020-07-16
Change source_id to source-id

dylanmcreynolds:
Fixes https://github.com/intake/intake/issues/515

I think there's probably an opportunity to make parameter passing more consistent. I updated the documentation just to make it more clear how parameters are passed for each message...sometimes query parameters, sometimes in the body of the message and just in the case of source-id, as an http header. But that would be a greater fix than this, which simply fixes the issue of reverse proxies not passing HTTP headers that contain underscores.

I also tried to write some tests that exercise this, but I was unsure how write a test that loads something in the cache for a future request, which seems to be where source_id is really used.

### DIFF
--- a/docs/source/server.rst
+++ b/docs/source/server.rst
@@ -20,21 +20,21 @@ GET /info
 
 Retrieve information about the data-sets available on this server. The list of data-sets may be
 paginated, in order to avoid excessively long transactions. Notice that the catalog for which a listing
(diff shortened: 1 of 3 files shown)


martindurant:
I will merge this, thank you.
You are right that the server protocol should probably be revisited to make things clearer and simpler.

PR #522, opened 2020-08-12
Defer YAML parsing on cat open until needed

### LINKED ISSUE #506: Make catalogs lazy by default
Currently, the Catalog base does the following on instantiation:
```
        self._entries = self._make_entries_container()  # gives {} by default
        self.force_reload()
```
so that `_load()` is called at that time (and whenever the TTL runs out, which is on by default!). Now, we *can* construct special lazy versions of `_entries`, but it would be nice if the default were to only do any loading when entries are actually accessed in any way.

Related to #505 (where there is no loading of entries in zarrcat's init, only _load, which is the recommended thing to do).

martindurant:
If "name" is not supplied, parse anyway, to get metadata and name.
When called from a catalog, should always have a name.

Fixes #506 (but not with a lazy entries object, which would take more work)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,6 @@ branches:
   only:
   - master
 
-env:
-  global:
-    # ANACONDA_TOKEN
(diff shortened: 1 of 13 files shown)


martindurant:
(some tests will pass doe to including "driver" in `describe()`, which is done to mirror the YAML output for a source)

PR #526, opened 2020-08-24
Lighten imports

martindurant:
With this PR, import times look like
```
  cumtime (ms)    intime (ms)  name
         210.3          104.7  intake
         105              0.6  +source
           5.4            1.2  ++entrypoints
           1.6            1.5  +++zipfile
           2.6            2.6  +++configparser
          19.7            0.7  ++utils
          18.9            0.7  +++yaml
          13.2            0.4  ++++loader
           7.2            7.2  +++++reader
           1.5            1.4  +++++constructor
           2.7            2.7  +++++resolver
           1.8            0.4  ++++dumper
           1.7            0.6  ++++cyaml
           1.1            1.1  +++++_yaml
          79.3            1    ++base
          78.3            0.7  +++cache
          74.7           60.2  ++++fsspec.utils
           1.6            0.8  +++++spec
           1.6            0.3  +++++registry
           1.3            1.3  ++++++distutils.version
          10.5            1.5  +++++mapping
           8.9            0.8  ++++++core
           7.7            0.8  +++++++compression
           3.8            0.6  ++++++++snappy
           2.6            0.6  +++++++++snappy
           2              2    ++++++++++_snappy
           1.3            0.4  ++++++++zstandard
           2.7            2.7  ++++intake.config
```

(I had trouble doing the timing on master, because pyarrow breaks the profiler; but I am satisfied that there is not much out of place in the output above. I did have to remove pyarrow to get the import profile above too, because fsspec optionally imports it. There should probably be an environment variable to prevent that.)

One big problem, causing failures here: `intake.gui` is a panel displaying instance OR a module. This unfortunate situation is totally my fault.


martindurant:
Note: this PR changes some imports!! Some downstream packages may depend on the specific locations, and need to be updated.

martindurant:
OK, so the sub-package has been renamed to `intake.interface`

Merging now

danielballan:
Confirmed that this broke `databroker` https://travis-ci.org/github/bluesky/databroker/jobs/722799708#L659

martindurant:
intake.catalog.base.RemoteCatalog is now intake.catalog.remote.RemoteCatalog ; we can put in the deprecated import in base, as you suggest

PR #528, opened 2020-08-28
Refactor DataSource

martindurant:
Make simplified DataSourceBase, and mixins containing the various optional behaviours which were all within DataSource.

For now, DataSource will depend on all the mixins, so existing behaviour will not change.


martindurant:
cc @tacaswell

martindurant:
(the various mixins will move from the base module, and remaining methods will move to DataSourceBase)

martindurant:
Obviously this needs some docs updates, but I wanted to get opinions at this point.

intake/source/base.py
```diff
@@ -53,7 +52,133 @@ class NoEntry(AttributeError):
...
+    def cache(self, csh):
+        self._cache = csh
+
+    def _get_cache(self, urlpath):
+        if len(self.cache) == 0:
+            return [urlpath]
+        return [c.load(urlpath) for c in self.cache]
+
+
+class GraphicsMixin:
```
tacaswell, line 85: Can we call this `HoloviewsMixin`?

intake/source/base.py
```diff
@@ -62,103 +187,75 @@ class DataSource(DictSerialiseMixin):
 
     This class is not useful in itself, most methods raise NotImplemented.
     """
+
     name = None
     version = None
     container = None
     partition_access = False
-    datashape = None
```
tacaswell, line 69: Why did `datashape` get removed?

martindurant: It's not used anywhere at all, is always None

intake/source/base.py
```diff
@@ -53,7 +52,133 @@ class NoEntry(AttributeError):
...
+        from ..container.persist import PersistStore
+        from dask.base import tokenize
+        if 'original_tok' in self.metadata:
+            raise ValueError('Cannot persist a source taken from the persist '
+                             'store')
+        if ttl is not None and not isinstance(ttl, (int, float)):
+            raise ValueError('Cannot persist using a time to live that is '
+                             f'non-numeric. User-provided ttl was {ttl}')
+        store = PersistStore()
+        out = self._export(store.getdir(self), **kwargs)
```
tacaswell, line 171: Should `_export` be made public?

martindurant: There is a public `export()` function; this is the structure to indicate which might be overridden.

intake/source/base.py
martindurant: Certainly! Or maybe HvPlot, to give the hint that this is plotting/gui-related?

intake/source/base.py
```diff
@@ -53,7 +52,133 @@ class NoEntry(AttributeError):
...
+                                      catdir=self.metadata.get('catalog_dir',
+                                                             None),
+                                      storage_options=self.storage_options)
+        return self._cache
+
+    @cache.setter
+    def cache(self, csh):
+        self._cache = csh
+
+    def _get_cache(self, urlpath):
```
tacaswell, line 79: Can we also make this public?  It might make also make sense to put a no-op version of this on `DataSourceBase` that always returns `[urlpath]` so that consumers don't have to care if there is an actual cache or not.

I am also a bit confused about how the cache actually gets used?  It looks like this is the method that is called in the concrete implementations.  This method then in turn calls `c.load` which defers to the Cache objects and either fetch and store the data, returning a new url (or list of urls?) that point to the cached data.  We do this for all of the caches that are set up, but everyplace this is called we do `self._get_cache(urlpath)[0]` which only looks at the first entry.  I think this means we can spend time / bits caching things we are never going to look at?

tacaswell: On a bit more consideration, would it be better to keep this private, but mix it in at the concrete source level?

martindurant: > would it be better to keep this private, but mix it in at the concrete source level?

In the long run yes, and I want to firmly deprecate this in favour of handling at the fsspec layer anyway. For this PR, I didn't want the actual functionality to move around too much.

intake/source/base.py
```diff
@@ -53,7 +52,133 @@ class NoEntry(AttributeError):
...
+
+    @property
+    def hvplot(self):
+        """
+        Returns a hvPlot object to provide a high-level plotting API.
+        """
+        return self.plot
+
+
+class PersistMixin:
```
tacaswell, line 128: As part of this I think the `s.get_persisted()` logic in `CatalogEntry.__call__` should be delegated to the `get` methods of the concrete implementations.  In addition data sources being able to opt in/out of implementing the persist logic, I think it is critical that the catalogs also be able to opt out of persist being used.

I guess another option is to do this as a mix-in on the `CatalogEntry` as well?

intake/source/base.py
martindurant: Let's leave this as future work. For now, I think we'll update the docs, sayin that new drivers can derive from DataSourceBase any any mixins they like, or DataSource to get them all.

intake/source/base.py
martindurant: Do you think it's reasonable for the catalog to use persisiting, if the mixin is there? It could even check by isinstance specifically.

martindurant: I added it...

intake/source/base.py
```diff
@@ -273,6 +278,18 @@ def _ipython_display_(self):
     def __repr__(self):
         return self.yaml()
 
+    def is_persisted(self):
+        """The base class does not interact with persistence"""
+        return False
+
+    def has_been_ersisted(self):
```
tacaswell, line 285:
```suggestion
    def has_been_persisted(self):
```

martindurant: 👍 
Obviously there should be at least one test driver based on the base class which exercises these methods.

PR #530, opened 2020-09-16
issue 529 fix with test

dwr-psandhu:
I had to modify one of the existing test cases and added a new one with a lone param. The test_utils passes however I am not sure what global effect this change could have. 
@martindurant please review

### DIFF
--- a/intake/source/tests/test_utils.py
+++ b/intake/source/tests/test_utils.py
@@ -18,7 +18,8 @@ from intake.source.utils import (
      'data/LT05_L1TP_042033_*_*_01_T1_sr_band*.tif'),
     ('{year}/{month}/{day}.csv', '*/*/*.csv'),
     ('data/**/*.csv', 'data/**/*.csv'),
-    ('data/{year:4}{month:02}{day:02}.csv', 'data/*.csv'),
+    ('data/{year:4}{month:02}{day:02}.csv', 'data/***.csv'),
+    ('{lone_param}','*')
 ])
 def test_path_to_glob(pattern, expected):
     assert path_to_glob(pattern) == expected
--- a/intake/source/tests/test_utils.py
+++ b/intake/source/tests/test_utils.py
@@ -18,7 +18,7 @@ from intake.source.utils import (
      'data/LT05_L1TP_042033_*_*_01_T1_sr_band*.tif'),
     ('{year}/{month}/{day}.csv', '*/*/*.csv'),
     ('data/**/*.csv', 'data/**/*.csv'),
-    ('data/{year:4}{month:02}{day:02}.csv', 'data/***.csv'),
+    ('data/{year:4}{month:02}{day:02}.csv', 'data/*.csv'),
     ('{lone_param}','*')
 ])
 def test_path_to_glob(pattern, expected):
--- a/intake/source/utils.py
+++ b/intake/source/utils.py
@@ -249,10 +249,10 @@ def path_to_glob(path):
     glob = ''
     prev_literal = None
     prev_field_name = None
-    for literal_text, field_name, format_spec, conversion in fmt.parse(path):
+    for literal_text, field_name, _, _ in fmt.parse(path):
         glob += literal_text
         # condition to avoid repeated * on adjacent fields
-        if field_name and not('' == literal_text and prev_field_name):
+        if field_name and (literal_text or prev_field_name is None):
             glob += '*'
         prev_literal = literal_text
         prev_field_name = field_name
--- a/intake/source/utils.py
+++ b/intake/source/utils.py
@@ -246,14 +246,9 @@ def path_to_glob(path):
         return path
 
     # Get just the real bits of the urlpath
-    literal_texts = [i[0] for i in fmt.parse(path)]
-
-    # Only use a star for first empty string in literal_texts
-    index_of_empty = [i for i, lt in enumerate(literal_texts) if lt == '' and i != 0]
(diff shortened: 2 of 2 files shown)


intake/source/tests/test_utils.py
```diff
@@ -18,7 +18,8 @@
      'data/LT05_L1TP_042033_*_*_01_T1_sr_band*.tif'),
     ('{year}/{month}/{day}.csv', '*/*/*.csv'),
     ('data/**/*.csv', 'data/**/*.csv'),
-    ('data/{year:4}{month:02}{day:02}.csv', 'data/*.csv'),
+    ('data/{year:4}{month:02}{day:02}.csv', 'data/***.csv'),
```
martindurant, line 21: This is definitely not good - the output is not a legal glob string. The asterisks should combine.

intake/source/utils.py
```diff
@@ -246,14 +246,9 @@ def path_to_glob(path):
...
     # Get just the real bits of the urlpath
-    literal_texts = [i[0] for i in fmt.parse(path)]
-
-    # Only use a star for first empty string in literal_texts
-    index_of_empty = [i for i, lt in enumerate(literal_texts) if lt == '' and i != 0]
-    glob = '*'.join([literal_texts[i] for i in range(len(literal_texts)) if i not in index_of_empty])
-
-    return glob
-
+    literal_texts=[ literal_text+ ('*' if field_name else '') for literal_text, field_name, format_spec, conversion in fmt.parse(path)]
```
martindurant, line 249: This is a bit hard to read - perhaps would be better as a long-form for-loop?

martindurant (commented):
@jsignell , this was your thing - do you have any time for a look?

intake/source/utils.py
dwr-psandhu: @martindurant @jsignell I tried to make it simple but stumbled onto a solution. Improvements and suggestion are welcome. See my commit below.

martindurant:
OK, functionality looks good!
Would you mind applying some style to the code, spacer around operators and lines that are not too long? We don't enforce pep8 in this repo, but we do try to keep close to it.

dwr-psandhu:
> Would you mind applying some style to the code, spacer around operators and lines that are not too long? We don't enforce pep8 in this repo, but we do try to keep close to it.
I thought I am using vscode pep8, but I'll do the spacing around operators and lines like you suggest.

intake/source/utils.py
```diff
@@ -245,16 +245,18 @@ def path_to_glob(path):
...
-
-    # Only use a star for first empty string in literal_texts
-    index_of_empty = [i for i, lt in enumerate(literal_texts) if lt == '' and i != 0]
-    glob = '*'.join([literal_texts[i] for i in range(len(literal_texts)) if i not in index_of_empty])
-
+    # calculate glob expression
+    glob = ''
+    prev_literal = None
+    prev_field_name = None
+    for literal_text, field_name, format_spec, conversion in fmt.parse(path):
```
jsignell, line 252: Since you aren't using `format_spec` and `conversion` no need to name them:

```suggestion
    for literal_text, field_name, _, _ in fmt.parse(path):
```

intake/source/utils.py
```diff
@@ -245,16 +245,18 @@ def path_to_glob(path):
...
+        glob += literal_text
+        if field_name and not('' == literal_text and prev_field_name):  # condition to avoid repeated * on adjacent fields
```
jsignell: This is hard to read. It might be easier without the double negative:

```suggestion
        if field_name and (literal_text or prev_field_name is None):  # condition to avoid repeated * on adjacent fields
```

intake/source/utils.py
```diff
@@ -245,16 +245,18 @@ def path_to_glob(path):
...
+            glob += '*'
+        prev_literal = literal_text
```
jsignell, line 257: I can't see where this is used, so maybe get rid of it?

intake/source/utils.py
```diff
@@ -245,16 +245,19 @@ def path_to_glob(path):
...
-    glob = '*'.join([literal_texts[i] for i in range(len(literal_texts)) if i not in index_of_empty])
-
+    # calculate glob expression
+    glob = ''
+    prev_literal = None
+    prev_field_name = None
+    for literal_text, field_name, format_spec, conversion in fmt.parse(path):
+        glob += literal_text
+        # condition to avoid repeated * on adjacent fields
+        if field_name and not('' == literal_text and prev_field_name):
```
jsignell, line 255: This double negative is hard to read. Maybe instead:

```suggestion
        if field_name and (literal_text or prev_field_name is None):
```

intake/source/utils.py
```diff
@@ -245,16 +245,19 @@ def path_to_glob(path):
...
-    # Get just the real bits of the urlpath
-    literal_texts = [i[0] for i in fmt.parse(path)]
-
-    # Only use a star for first empty string in literal_texts
-    index_of_empty = [i for i, lt in enumerate(literal_texts) if lt == '' and i != 0]
-    glob = '*'.join([literal_texts[i] for i in range(len(literal_texts)) if i not in index_of_empty])
-
+    # calculate glob expression
+    glob = ''
+    prev_literal = None
```
jsignell, line 250: See comment below.

intake/source/utils.py
dwr-psandhu: agreed. probably left over from my trying to work out the logic

martindurant:
Thanks for this

jsignell:
Yeah thanks @dwr-psandhu!

PR #538, opened 2020-10-09
Template function to offset datetime parameter

leifdenby:
I have a opendap server where the files are stored with a time-span in the filename which is always one hour apart, for example
`LIDAR_obs_09291300_09291400.nc` 
i.e. the format is 
`LIDAR_obs_{{t_start.strftime("%m%d%H%M")}}_{{t_end.strftime("%m%d%H%M")}}.nc`

It would be nice to have the user just provide `t_start` (which I can then check for validity) and then calculate `t_end` from it (`t_end = t_start + datetime.timedelta(hours=1)`). I appreciate that allowing for custom functions would open a whole can of worms, but maybe adding a specialised jinja template function for offsetting datetimes might be useful. I've created one called `offset_datetime` write
`LIDAR_obs_{{t_start.strftime("%m%d%H%M")}}_{{offset_datetime(t_start, hours=1).strftime("%m%d%H%M")}}.nc`

Would this be something you would consider adding to `intake`?

### DIFF
--- a/intake/catalog/utils.py
+++ b/intake/catalog/utils.py
@@ -12,6 +12,7 @@ import os
 import re
 import shlex
 import subprocess
+import datetime
 import sys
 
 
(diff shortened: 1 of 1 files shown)


martindurant:
I am personally uncertain. I can see the benefit, obviously - and indeed there could be many other utility functions that might be handy. I'm not sure how one would go about specifying the set to include - we would *not* generally want to allow arbitrary code execution during template filling.

Perhaps you could tag one or two of the maintainers on intake-xarray to see if someone else already has a solution for this kind of thing or has thoughts on it.

PR #541, opened 2020-10-23
FIX: defer pulling from entry points and specialer case intake.cat

tacaswell:
If the entrypoints found code that was in turn imported intake.catalog
then, because of the user of `__getattr__` in `intake/__init__.py` the
downstream libraries would fail to import and get see an
AttributeError.

This change defers filling the entry points until the user actually
asks for `intake.cat`, not just importing `intake.catalog`.

### DIFF
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -19,7 +19,6 @@ from .source import register_driver, registry
 imports = {
     "DataSource": "intake.source.base:DataSource",
     'Schema': "intake.source.base:Schema",
-    "cat": "intake.catalog:builtin",
     "load_combo_catalog": "intake.catalog.default:load_combo_catalog",
     "upload": "intake.container:upload",
     "gui": "intake.interface:instance",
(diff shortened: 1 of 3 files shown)


martindurant:
I guess this is another case of "intake don't need to import the world" :)

A simple test would be to import intake, and see that intake.catalog is None (would need to pop intake from sys.modules, etc).

danielballan:
Good suggestion, @martindurant. I went ahead and pushed a commit to implement that.

danielballan:
Very confused by why this broke everything....

martindurant:
Wow, that broke *everything*. Looking now.

danielballan:
Thanks.

martindurant:
OK, all good? I think the short story was "don't break imports" :)

danielballan:
Haha, amen! This looks better. I like isolating the `__getattr__` magic to the submodule in this case.

PR #544, opened 2020-10-30
Add validated entry_type.

### LINKED ISSUE #534: Add API to Catalog declaring type of Entries
Some Catalogs contain a mixture of types, but at the very least we can declare that the type is `object` or `DataSource`. For Catalogs with uniform entries, we can use this to declare that all the entries are, e.g. `CSVDataSource` or `BlueskyRun`. This is useful for GUIs that want to display different UI elements depending on the contents of the catalog.

danielballan:
Closes #534

This adds a new optional parameter to `Catalog`, `entry_type`. Its default value, `object`, accepts entries of any type and is thus backward-compatible. Subclasses or instances may override this default with a more specific value. Consumers of `Catalogs` (thinking mainly of libraries or GUIs, but perhaps the occasional interactive user) can reference `catalog.entry_type`, a new read-only property on `Catalog` instances to know the type of the entries. For example, a GUI can use this to set up a different UI for browsing Catalogs with different types of entries.

- [x] Initial implementation
- [ ] Tests
- [ ] Add analogous API to `RemoteCatalog`

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -40,7 +40,8 @@ class Catalog(DataSource):
 
     def __init__(self, *args, name=None, description=None, metadata=None,
                  ttl=60, getenv=True, getshell=True,
-                 persist_mode='default', storage_options=None):
+                 persist_mode='default', storage_options=None,
+                 entry_type=object):
         """
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


martindurant:
Give me a poke when you want a review here.

PR #548, opened 2020-11-05
Expect entry or source to implement 'container' but not '_container'.

### LINKED ISSUE #545: Regression: internals require API that is not always defined
It seems as if not all Catalogs implement `_container`, and so this change

```diff
-            return e(name=key)
+            if e._container == 'catalog':
+                return e(name=key)
+            return e()
```

in https://github.com/intake/intake/commit/4aa561c86503d609ed73a5ae9c7e9256b6560bb4 creates some issues. Here's a minimal reproducible example using only built-in intake objects.

```py
In [1]: import pandas

In [2]: df = pandas.DataFrame({'a': [1,2,3], 'b': [4,5,6]})

In [3]: df.to_csv('file.csv')

In [4]: from intake.source.csv import CSVSource

In [5]: from intake import Catalog
/home/dallan/Repos/bnl/intake/intake/__init__.py:43: UserWarning: deprecation: intake.Catalog now references the base class intake.catalog.base.Catalog
 If you want to open a generic URL, you should use intake.open_catalog
  warnings.warn('deprecation: intake.Catalog now references the base class intake.catalog.base.Catalog\n '

In [6]: catalog = Catalog.from_dict({'stuff': CSVSource('file.csv')})

In [7]: catalog['stuff']
AttributeError                            Traceback (most recent call last)
<ipython-input-7-ad23fb20a5f6> in <module>
----> 1 catalog['stuff']

~/Repos/bnl/intake/intake/catalog/base.py in __getitem__(self, key)
    386             e._catalog = self
    387             e._pmode = self.pmode
--> 388             if e._container == 'catalog':
    389                 return e(name=key)
    390             return e()

AttributeError: 'CSVSource' object has no attribute '_container'
```

We see the same issue with databroker's Catalogs. I haven't found a workaround yet (chasing that in https://github.com/bluesky/databroker/pull/619). It seems worth fixing upstream anyway.

danielballan:
Closes #545 

This turned out to be an easy fix: just rely on the public API instead of the private one. The intake tests pass without changes. Likewise, the databroker tests pass against this branch with no workarounds. This PR also incidentally enables

```py
Catalog.from_dict({'thing': CSVSource(...)})
```

without wrapping `CSVSource(...)` in an `Entry`, which fits our goal of "Let users never see Entry objects" as @martindurant put it in #490. I added a test to exercise that usage.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -385,7 +385,7 @@ class Catalog(DataSource):
             e = self._entries[key]
             e._catalog = self
             e._pmode = self.pmode
-            if e._container == 'catalog':
+            if e.container == 'catalog':
                 return e(name=key)
             return e()
         if isinstance(key, str) and '.' in key:
--- a/intake/catalog/tests/test_local.py
+++ b/intake/catalog/tests/test_local.py
@@ -664,6 +664,16 @@ def test_filter():
     assert cat2.trial2 == entry2()
 
 
+def test_from_dict_with_data_source():
+    "Check that Catalog.from_dict accepts DataSources not wrapped in Entry."
+    from intake.catalog.base import Catalog
+    fn = os.path.join(tempfile.mkdtemp(), 'mycat.yaml')
+    entry = LocalCatalogEntry(name='trial', description='get this back',
+                              driver='csv', args=dict(urlpath=""))
+    ds = entry()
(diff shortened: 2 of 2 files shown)


martindurant:
At *some point*, we can advertise the alternate use of from_dict

PR #547, opened 2020-11-04
auth=None for base Catalog; remove old usage of intake.Catalog

martindurant:
Fixes #543 
Fixes #542 

@danielballan : how hard would it be to add some bluesky tests to a separate CI build? That seems like a good guard against regressions.

PR #549, opened 2020-11-09
Accept 'auth' parameter in __init__.

danielballan:
In #547, an `auth` attribute was restored, but ``__init__`` still does not
accept an ``auth`` parameter, so the backward-incompatibility was not fully
addressed.


intake/catalog/base.py
```diff
@@ -37,10 +37,9 @@ class Catalog(DataSource):
     # emulate a DataSource
     container = 'catalog'
     name = 'catalog'
-    auth = None
 
     def __init__(self, entries=None, name=None, description=None, metadata=None,
-                 ttl=60, getenv=True, getshell=True,
+                 auth=None, ttl=60, getenv=True, getshell=True,
```
tacaswell, line 42: Is this the position it used to be at?

intake/catalog/base.py
```diff
@@ -58,6 +57,8 @@ def __init__(self, entries=None, name=None, description=None, metadata=None,
         ttl : float, optional
             Lifespan (time to live) of cached modification time. Units are in
             seconds. Defaults to 1.
+        auth : BaseClientAuth or None
```
tacaswell, line 60: This should be moved up to be above ttl?

martindurant:
All remote stuff broke :|

intake/catalog/base.py
danielballan: Yes, I checked before placing it.

intake/catalog/base.py
danielballan: Oops, yes.

danielballan:
Getting closer here...

PR #559, opened 2020-12-16
typo fix

jsta:


martindurant:
@jsignell , are you aware of any upstream changes that would have resulted in the three intake/source/tests/test_csv.py::test_read_pattern* tests starting to fail?

martindurant:
Thank you, @jsta

PR #560, opened 2020-12-16
Fix category ordering for path column

martindurant:
This likely works around a dask bug for include_path_column , that the category columns don't end up in the same order as the file paths (which are sorted by fsspec). I'm not sure is dask is supposed to guarantee the order (cc @jsignell )

Ref #559 - this is causing the failures.


jsignell:
Ah ok. I thought I added category sorting, but maybe I missed something. Thanks for pinging me.

Issue #537, opened 2020-10-02
[Feature Request] Support default environment variables in catalog templating


jhamman:
Suppose I have a catalog that looks like this:

```yaml
sources:
  mysource:
    description: This source needs a prefix environment variable
    args:
      url: "{{ env(MY_PREFIX) }}/data.csv"
```

This requires the `MY_PREFIX` be set by the user for data access to work as designed. Moreover, it results in an unexpected error (something like `FileNotFoundError("'/data.csv' not found")`)  that doesn't tell the user they should have provided an environment variable.

I'm curious if it would be possible to, somewhere in the Catalog, tell intake what to do when  `MY_PREFIX` is not set. In my use case, I would prefer being able to specify a default value but a more informative error would also do.

Something like:

```yaml
env:
  default:
    MY_PREFIX: 'https://example.com'
  # or
  required: [MY_PREFIX, MY_OTHER_PREFIX]

sources:
  ...
```

This syntax would allow intake to a) use default variables when they are not set and/or provide meaningful errors when required variables are not set. Thoughts?

martindurant:
This is definitely doable, after we decide what feels most natural.

For implementation, it would require either the catalog class to copy the variables into any entries that are opened from it, or for the entry to introspect its parent cat when called. Both of those are technically easy, but perhaps feel a little dirty. The CATALOG_DIR variable (based on the path the cat was loaded from, if a file) uses the former mechanism.

PR #562, opened 2020-12-16
Add support for declaring global user parameters

philippjfr:
In some scenarios you are defining multiple sources that require the same parameters. Currently the only way to deal with that is to duplicate the parameters for each source that requires them. This PR additionally allows global parameters, my specific use case was a Snowflake database where a number of parameters are required to declare the connection string (and query):

```yaml
metadata:
  version: 1
parameters:
  warehouse:
    description: "The Snowflake warehouse"
    type: str 
    default: "..."
  role:
    description: "The Snowflake role"
    type: str
    default: "..."
  database:
    description: "The Snowflake database"
    type: str
    default: "..."
  schema:
    description: "The Snowflake schema"
    type: str
    default: "..."
  account:
    description: "The Snowflake account"
    type: str
    default: "..."
sources:
  table1:
    driver: sql
    args:
      uri: 'snowflake://{{env(SNOWFLAKE_USER)}}:{{env(SNOWFLAKE_PASSWORD)}}@{{account}}/{{database}}/{{schema}}?warehouse={{warehouse}}&role={{role}}&echo=False'
      sql_expr: 'SELECT * FROM "..." LIMIT 100'
  table2:
    driver: sql
    args:
      uri: 'snowflake://{{env(SNOWFLAKE_USER)}}:{{env(SNOWFLAKE_PASSWORD)}}@{{account}}/{{database}}/{{schema}}?warehouse={{warehouse}}&role={{role}}&echo=False'
      sql_expr: 'SELECT * FROM "..." LIMIT 100'
```

### ToDo

- [ ] Add tests
- [ ] Ensure this also works with remote catalogues
- [ ] Ensure that source-specific parameter overrides global one

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -418,10 +418,22 @@ class CatalogParser(object):
     def _parse_data_source(self, name, data, global_parameters):
         if data.pop('remote', False):
             from intake.catalog.remote import RemoteCatalogEntry
+            if global_parameters:
+                data = dict(data)
+                source_parameters = data.get('user_parameters', [])
+                data['user_parameters'] = global_parameters + source_parameters
             return RemoteCatalogEntry(getenv=self.getenv,
                                       getshell=self.getshell, **data)
         elif 'cls' in data:
             from intake.utils import remake_instance
+            if global_parameters:
+                data = dict(data)
+                if 'kwargs' in data:
+                    data['kwargs'] = dict(data['kwargs'])
+                else:
+                    data['kwargs'] = {}
+                source_parameters = data['kwargs'].get('parameters', [])
+                data['kwargs']['user_parameters'] = global_parameters + source_parameters
             return remake_instance(data)
         else:
             return self._parse_data_source_local(name, data, global_parameters)
@@ -482,7 +494,7 @@ class CatalogParser(object):
                                  getenv=self.getenv, getshell=self.getshell,
                                  **ds)
 
-    def _parse_data_sources(self, data):
+    def _parse_data_sources(self, data, global_parameters):
         sources = []
 
         if 'sources' not in data:
@@ -496,11 +508,6 @@ class CatalogParser(object):
             self.error("value of key 'sources' must be a dictionary", data,
                        'sources')
             return sources
-
-        if 'parameters' in data:
-            global_parameters = self._parse_user_parameters(data)
-        else:
-            global_parameters = []
 
         for name, source in data['sources'].items():
             if not isinstance(name, str):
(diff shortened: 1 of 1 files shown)


martindurant:
Ref https://github.com/intake/intake/issues/537

philippjfr:
#537 isn't exactly the same since it's about env vars but probably share various similar issues in terms of implementation. In particular your comment about the implementation gives me pause:

>For implementation, it would require either the catalog class to copy the variables into any entries that are opened from it, or for the entry to introspect its parent cat when called. Both of those are technically easy, but perhaps feel a little dirty. The CATALOG_DIR variable (based on the path the cat was loaded from, if a file) uses the former mechanism.

You indicate that the Catalog class would keep the variables and pass them down to the entries as needed. In this particular PR I've simply extracted the global parameters during the yaml parsing which means the Catalog never holds these parameters at all. I can see that it might be desirable for the Catalog to hold the parameters though since that would allow you to create a Catalog with global parameters programmatically rather than limiting it to the yaml parsing. What would you suggest?

philippjfr:
Also have a few questions about the parsing

```python
    def _parse_data_source(self, name, data, global_parameters):
        if data.pop('remote', False):
            from intake.catalog.remote import RemoteCatalogEntry
            return RemoteCatalogEntry(getenv=self.getenv,
                                      getshell=self.getshell, **data)
        elif 'cls' in data:
            from intake.utils import remake_instance
            return remake_instance(data)
        else:
            return self._parse_data_source_local(name, data, global_parameters)
```

Here local data sources are parsed by the `_parse_data_source_local` method, which ends up running `_parse_user_parameters` on `data['parameters']` but in the remote catalog case the `data` is passed directly through to the `RemoteCatalogEntry` which means the parameters are never parsed? Is this intentional? Are they somehow being parsed into `UserParameter` instances elsewhere?

martindurant:
In the remote case, the server parses the user_parameters, and passes them to the client in the serialised version of the entry:
```
$ intake-server ./intake/catalog/tests/catalog1.yml

In [1]: import intake

In [2]: cat = intake.open_catalog("intake://localhost:5000")

In [3]: cat.entry1_part.describe()
Out[3]:
{'name': 'entry1_part',
 'container': 'dataframe',
 'plugin': 'remote',
 'description': 'entry1 part',
 'direct_access': 'allow',
 'metadata': {'foo': 'baz', 'bar': [2, 4, 6]},
 'user_parameters': [{'name': 'part',
   'description': 'part of filename',
   'type': 'str',
   'allowed': ['1', '2'],
   'default': '1'}],
 'args': ('http://localhost:5000/',)}

In [4]: cat.entry1_part(part="2").read()
Out[4]:
       name  score  rank
0    Alice2  100.5     1
1      Bob2   50.3     2
2  Charlie2   25.0     3
3      Eve2   25.0     3
```
The interpretation of the user parameter versus what the user provides can happen locally or on the server (some environment expansions only make sense in the right place, and only some sources are available to be read locally). See intake.catalog.remote.open_remote

martindurant:
This one fell through the cracks. Did you intend to finish it off?

SultanOrazbayev:
Yes, but it might take me another week or so.

martindurant:
No hurry, @SultanOrazbayev , glad to have your attention :)

PR #567, opened 2021-01-27
Allow pathlib.Path in open_catalog

### LINKED ISSUE #314: Pass pathlib to open_catalog
Was trying to pass a `pathlib.Path` to `intake.open_catalog` and got the following error:

```
ValueError                                Traceback (most recent call last)
<ipython-input-22-ceeaa41394e2> in <module>
----> 1 cat = intake.open_catalog(cat_file)

/nas/jlord/conda/envs/intake/lib/python3.7/site-packages/intake/__init__.py in open_catalog(uri, **kwargs)
    149     if driver not in registry:
    150         raise ValueError('Unknown catalog driver (%s), supply one of: %s'
--> 151                          % (driver, list(sorted(registry))))
    152     return registry[driver](uri, **kwargs)
    153 

ValueError: Unknown catalog driver (None), supply one of: ['catalog', 'csv', 'intake_remote', 'ndzarr', 'numpy', 'spark_cat', 'spark_dataframe', 'spark_rdd', 'textfiles', 'yaml_file_cat', 'yaml_files_cat']
```

easy fix is passing `str(pathlib.Path)` but might be nice to be able to pass it directly in the future or giving a clearer error message.

martindurant:
Fixes #314

### DIFF
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -126,6 +126,8 @@ def open_catalog(uri=None, **kwargs):
     intake.open_intake_remote
     """
     driver = kwargs.pop('driver', None)
+    if hasattr(path, "__fspath__"):
+        path = path.__fspath__()
     if driver is None:
         if uri:
             if ((isinstance(uri, str) and "*" in uri)
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -127,7 +127,7 @@ def open_catalog(uri=None, **kwargs):
     intake.open_intake_remote
     """
     driver = kwargs.pop('driver', None)
-    if isinstance(uri, os.Pathlike):
+    if isinstance(uri, os.PathLike):
         uri = os.fspath(uri)
     if driver is None:
         if uri:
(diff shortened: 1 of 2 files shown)


intake/__init__.py
```diff
@@ -126,6 +126,8 @@ def open_catalog(uri=None, **kwargs):
     intake.open_intake_remote
     """
     driver = kwargs.pop('driver', None)
+    if hasattr(uri, "__fspath__"):
+        uri = uri.__fspath__()
```
danielballan, line 130: I think that `os.path(uri)` would be better here. It handles the dispatch. And this reads like `a.__len__()` vs `len(a)`.

danielballan (approved):
This has bitten me a couple times too. Good to see.

I left a minor suggestion.

intake/__init__.py
martindurant: I specifically test for `__fspath__`, though. What is the best way to test for something `Path`-like? Seems best to avoid `isinstance` for, e.g., the possible future fsspec-pathlike compatibility layer.

martindurant: (actually, the simplest conversion from Path to str is `str()`)

Issue #570, opened 2021-02-08
Two new drivers for the plugin directory


AlbertDeFusco:
[intake-sklearn](https://github.com/AlbertDeFusco/intake-sklearn): Load scikit-learn models from Pickle files (`sklearn`)

[intake-metabase](https://github.com/continuumio/intake-metabase): Load Metabase tables as DataFrames (`metabase_table`)

martindurant:
Easiest is to PR to the docs page (plugin-directory.rst)

AlbertDeFusco:
will do!

PR #571, opened 2021-02-08
two new plugins

AlbertDeFusco:
This addresses #570


martindurant:
Do you want to update with the metabase catalog driver?

AlbertDeFusco:
Done! Thanks for the reminder

martindurant:
Thanks!

PR #576, opened 2021-02-18
Mutate entries.

### LINKED ISSUE #575: Mutate rather the overwrite self._entries
An effect of this change is that anything else holding a reference to `self._entries` (say, in a subclass) will have a reference to a stale copy.

https://github.com/intake/intake/commit/e8759199587a7e2ac0e09aa3efcb88593c26caad#diff-600ceff2b8688f2ca92e05b9d68a79b03c7c4dce84c750497ae2aa99b7a2e2ebL793-R794

Would it be OK to revert back to mutating `self._entries`?

danielballan:
Fixes #575

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -792,7 +792,11 @@ class YAMLFilesCatalog(Catalog):
                 entries.update(entry._entries)
             else:
                 entries[entry._name] = entry
-        self._entries = entries
+        # This is thread safe as long as the hash values for entries are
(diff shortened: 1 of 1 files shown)


martindurant:
The build seems to be stalling at test_reload_missing_local_directory , I have no idea why! This only happens on the pip-installed version, where presumably the deps have slightly newer versions.

martindurant:
Oh and plus, a whole load of tests fail first!

PR #598, opened 2021-06-10
Don't let configure_new loose previous kwargs

martindurant:
Ref https://github.com/dr-leo/intake_sdmx/pull/1

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -342,8 +342,8 @@ class Catalog(DataSource):
             # Fall back to __getitem__.
             try:
                 return self[item]  # triggers reload_on_change
-            except KeyError:
-                raise AttributeError(item)
+            except KeyError as e:
+                raise AttributeError(item) from e
         raise AttributeError(item)
 
     def __setitem__(self, key, entry):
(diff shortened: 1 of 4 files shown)


martindurant:
@tacaswell , @danielballan : there are already far too many hidden attributes like this, just to make the API work. The issue here, is that a new data source instance should keep the same kwargs as the one it was made from except for those give by the user. However, if the user passes a changes user-parameter, which would change the value of an actual kwarg, we need to know that that actual kwarg had *not* been passed by the user, but evaluated from default.

Example for the alias test:
```yaml
  alias1:
    driver: alias
    args:
      target: "{{choice}}"
      mapping:
        first: input_data
        second: arr_cache
    parameters:
      choice:
        description: which to alias
        type: str
        default: first
        allowed: ["first", "second"]
```
If first you have 
```python
s = cat.alias1(choice="first")
```
then s will have been passed `target="input_data"`. If you now want to reconfigure with
```python
s2 = s(choice="second")
```
then we need to know that `target` was not among the kwargs explicitly passed to `s`. That is what got implemented here, but it feels bad.

danielballan:
I don’t see a better way of handling this situation.

martindurant:
@dr-leo this is merged

PR #600, opened 2021-07-02
Change YAMLFileCat so that any file-level parameters are inherited by…

blackary:
… child entries, and add tests


blackary:
I am unable to reproduce the test or build failures -- any suggestions?

martindurant:
Fix is in #601, due to dask upstream change

martindurant:
please merge from master

martindurant:
I like what's here.
Two things to check: 
- what happens if we extract a subset of entries via `.search()`?
- what happens if the file is one of many in YAMLFilesCatalog (i.e., returned from `open_catalog(glob_path)`?

It would be good to add documentation, however - I suppose where we talk about the structure of YAML files.

Later, we may well want to move this functionality up to `Catalog`, so that all subclasses allow for cat-global parameters. This might be trickier, since existing implementations take responsibility in creating their entries (but entries keep a reference to the catalog created them).

blackary:
I'll check those things

blackary:
@martindurant I checked & fixed those two cases, and added notes to the documentation about YAML files.

martindurant:
An additional thought: if a catalogue contains references to other catalogues, should the global parameters be inherited?

blackary:
That wouldn't happen with the current implementation (because parameters are only inherited if they're specified in the metadata of the yaml_file_cat file), but it could be added if desired. The nice thing about not inheriting parameters beyond the current file is that it means a catalog file always works in isolation, or with a parent.

martindurant:
OK, maybe that's something to think about in the future. I would say that the ability for catalogs to push some parameters to their children might be interesting. In fact, you perhaps have accidentally enabled it here, should a catalog reference another catalog and include the right `metadata=` argument.

intake/catalog/base.py
```diff
@@ -9,9 +9,11 @@
 import logging
 import re
 import time
+from tkinter import W
```
martindurant, line 12: I didn't notice this before.
This line isn't needed right? tkinter isn't amongst Intake's requirements.

blackary: Not sure how that got in there. Probably an accidental auto-import. I'll remove it.

martindurant: I think I spotted one further down too

martindurant: (oh, you got it)

martindurant:
Thank you, going in

PR #605, opened 2021-07-14
Fix docstring for valid compression values

blackary:
I'm not sure exactly the relationship between the valid compressions according to fsspec, and the compression values in `intake.source.decompress.decomp`, but I do know that `fsspec.open_files("blah", compression="gz")` raises 
```
ValueError: Compression type gz not supported
```

### DIFF
--- a/intake/source/textfiles.py
+++ b/intake/source/textfiles.py
@@ -40,7 +40,7 @@ class TextFilesSource(base.DataSource):
             common
         compression : str or None
             If given, decompress the file with the given codec on load. Can
-            be something like "gz", "bz2", or to try to guess from the filename,
+            be something like "gzip", "bz2", or to try to guess from the filename,
             'infer'
         decoder : function, str or None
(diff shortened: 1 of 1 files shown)


martindurant:
Yep, this should be the same list as in `fsspec.compression.compr`. I'll merge this (thank you), but you could add more detail if you like.

blackary:
Should `intake.source.decompress.decomp` also match `fsspec.compression.compr`?

martindurant:
Those are file extensions, a more appropriate comparison would be with `fsspec.utils.compressions` (which is perhaps in the wrong place). In any case, Intake probably should defer compression to fsspec, rather than have its own. The place where that becomes more tricky, is where the target file contains many files (zip, tar), which with fsspec you would access like "zip://**::s3://bucket/file.zip", meaning "all the files in that target" and makes it tricky to use in conjunction with caching.

PR #609, opened 2021-07-21
convert classes to strings during serialization

### LINKED ISSUE #608: built in yaml_file_cat driver results in yaml ConstructorError for intake_xarray driver
I've got the following `test_catalog.yml` file:
```
metadata:
  version: 1
name: IS2-hdf5-icepyx-intake-catalog
sources:
  is2_local:
    args:
      path_as_pattern: ATL{product:2}-{datetime:%Y%m%d%H%M%S}-Sample.h5
      urlpath: /Users/jessica/computing/icepyx/test_data/ATL06-20181214041627-Sample.h5
      xarray_kwargs:
        engine: h5netcdf
        group: /gt1l/land_ice_segments
    description: test text here
    driver: !!python/name:intake_xarray.netcdf.NetCDFSource ''
    name: is2_local
    parameters: {}
```

When I try to open the catalog using `intake.open_catalog("test_catalog.yml")`, I get a ConstructorError:
```
ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:python/name:intake_xarray.netcdf.NetCDFSource'
  in "<unicode string>", line 13, column 13:
        driver: !!python/name:intake_xarray.netc ... 
```

I've been able to trace the error through intake and yaml to [`intake.utils.yaml_load(stream)`](https://github.com/intake/intake/blob/4cf2a7a914825a4b5d7ade710babd22184a3c267/intake/utils.py#L78). `yaml.safe_load()` is [clearly](https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation) the best method to use, but obviously an issue if we're trying to use other registered drivers.

I'm not sure if this is technically a bug, or if there's some type of driver registration setup/step I'm missing (since I'm able to import and use the driver to read in data). My naive solution would be to somehow add checks and constructors for other drivers (maybe [here](https://github.com/intake/intake/blob/ded664427ea1d31e23e7521ef4d170df2934d741/intake/utils.py#L60)?). I'm an intake novice, but I'd be happy to contribute towards a solution directly to intake if that's the best spot for it, or at least add some notes to the docs if there's another recommended approach.

JessicaS11:
Closes #608 

I tried using `kw['driver'].__name__`, but it output only "NetCDFSource".
Similarly, `kw['driver'].__class__.__name__` and `type(kw['driver'])` output "type" rather than "intake_xarray.netcdf.NetCDFSource". I figured this may be a common issue with other drivers not having registered names and thus went with the non-breaking solution. Happy to change it if there's a better solution.

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -276,8 +276,11 @@ class Catalog(DataSource):
             kw = entry._captured_init_kwargs.copy()
             kw.pop('catalog', None)
             kw['parameters'] = {k.name: k.__getstate__()['kwargs'] for k in kw.get('parameters', [])}
-            if issubclass(kw['driver'], DataSourceBase):
-                kw['driver'] = ".".join([kw['driver'].__module__, kw['driver'].__name__])
+            try:
+                if issubclass(kw['driver'], DataSourceBase):
+                    kw['driver'] = ".".join([kw['driver'].__module__, kw['driver'].__name__])
+            except TypeError:
+                pass # ignore exception for a string input
             output["sources"][key] = kw
         return yaml.dump(output)
 
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -276,6 +276,8 @@ class Catalog(DataSource):
             kw = entry._captured_init_kwargs.copy()
             kw.pop('catalog', None)
             kw['parameters'] = {k.name: k.__getstate__()['kwargs'] for k in kw.get('parameters', [])}
+            if not isinstance(kw['driver'], str):
+                kw['driver'] = str(kw['driver']).split("'")[1]
             output["sources"][key] = kw
         return yaml.dump(output)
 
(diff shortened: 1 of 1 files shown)


intake/catalog/base.py
```diff
@@ -276,6 +276,8 @@ def serialize(self):
             kw = entry._captured_init_kwargs.copy()
             kw.pop('catalog', None)
             kw['parameters'] = {k.name: k.__getstate__()['kwargs'] for k in kw.get('parameters', [])}
+            if not isinstance(kw['driver'], str):
+                kw['driver'] = str(kw['driver']).split("'")[1]
```
martindurant, line 280: Could be `".".join([driver.__module__, driver.__name__])` ? Actually, ":" should also work as the join character I think.

I think that, instead of doing this for all non-string things, I would change the check to
```
if issubclass(kw['driver'], intake.source.base.DataSourceBase)
```

martindurant:
Thanks, @JessicaS11

PR #615, opened 2021-09-07
Add intake_pattern_catalog

tdhopper:


martindurant:
I wonder if you want to specify that this is a catalogue driver, that the various file paths become catalogue entries (as opposed to data, as is the case for path patterns in CSV sources).
Now that I've written that out, it seems maybe too long for this spot!

tdhopper:
@martindurant I had the same thought and had come to the same conclusion

martindurant:
OK, merging this as-is.

PR #618, opened 2021-10-06
Add ability to set persist mode

tylerpotts:
We were working on creating a catalog and were consistently running into this bug: 

```
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/intake/catalog/entry.py in __call__(self, persist, **kwargs)
     76         persist = persist or self._pmode
     77         s = self.get(**kwargs)
---> 78         if persist != 'never' and isinstance(s, PersistMixin) and s.has_been_persisted:
     79             from ..container.persist import store
     80 

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/intake/catalog/base.py in __getattr__(self, item)
    350                 return self[item]  # triggers reload_on_change
    351             except KeyError as e:
--> 352                 raise AttributeError(item) from e
    353         raise AttributeError(item)
```

We wanted persist to be `never`, but there doesn't seem to be a current way to pass this as an argument when creating the catalog on the client side.

This PR adds the ability to pass arguments like so:

```python
catalog = intake.open_catalog("intake://localhost:9001", persist_mode="never")
```

This doesn't change any of the functionality of the codebase.

Kudos to @naatebarber, @ghoersti for the help figuring this out

### DIFF
--- a/intake/tests/test_config.py
+++ b/intake/tests/test_config.py
@@ -12,7 +12,7 @@ import requests
 import intake
 from intake import config
 from intake.util_tests import temp_conf, server
-from intake.remote import RemoteCatalog
+from intake.catalog.remote import RemoteCatalog
 
 @pytest.mark.parametrize('conf', [
     {},


hussainsultan:
cc: @martindurant thanks for looking into it.

ghoersti:
I am unsure if the persisting from remote does work as expected. But we can certainly add tests for this functionality.

martindurant:
> you do not mean server side persistence

Correct.

> we have direct access set to forbid

`Remote*` sources should be persistable too, turning into local parquet, zarr, etc. files. I think that's what you said :)

I believe we agreed on everything.

PR #625, opened 2021-11-02
Add drivers for .json and .jsonl file types

blackary:


martindurant:
I see absolutely nothing wrong with the code here!

It may be worth considering how to draw the boundary between generic objects here and record/dataframe-like json. For example, dask has a [`read_json`](https://docs.dask.org/en/latest/generated/dask.dataframe.read_json.html) method which can split JSONL files, or produce one partition per file if not line-delimited. That can be left for follow-up work.

blackary:
@martindurant Agreed. It seems like it might be worth having different plugins, one which is for dataframe-like json, and one which is for generic json objects. How do you feel about using `json` & `jsonl` for the plugin for generic objects, and then eventually having another plugin `json_tables` or something like that? Or, would you prefer these plugins be called `raw_json`/`raw_jsonl` or something like that, since they're different from most of the plugins which work with tabular data?

martindurant:
I think keeping the current names here is fine. For those that want to be explicit, they can always give the whole package.class form. In case you're wondering, there is a resolution process when two plugins share the same name, but we haven't implemented a "load as dataframe"/"load as generic" set of methods where different implementations are available, something that had been mooted at some point.

intake/source/tests/test_json.py
```diff
@@ -0,0 +1,115 @@
...
+    file_path.write_text("\n".join(json.dumps(row) for row in data))
+    return file_path
+
+
+@pytest.fixture
+def jsonl_file_compressed_paths(tmp_path) -> Dict[Optional[str], str]:
+    data = [{"hello": "world"}, [1,2,3]]
+    file_path = tmp_path / "1.json"
+    compressed_paths = {}
+    for compression in [None, "gzip", "bz2"]:
```
tdhopper, line 56: You could write this as a parameterized fixture instead

intake/source/tests/test_json.py
```diff
@@ -0,0 +1,115 @@
...
+    assert isinstance(out[0], dict)
+    assert out[0]["hello"] == "world"
+
+    assert isinstance(out[1], list)
+    assert out[1] == [1, 2, 3]
+
+
+@pytest.mark.parametrize('compression', [None, 'gzip', 'bz2'])
+def test_compressed_json(json_file_compressed_paths: Dict[Optional[str], str], compression: Optional[str]):
+    path = json_file_compressed_paths[compression]
```
tdhopper, line 89: Wouldn't be required if using a parameterized fixture

intake/source/tests/test_json.py
```diff
@@ -0,0 +1,115 @@
...
+    assert isinstance(out, list)
+
+    assert isinstance(out[0], dict)
+    assert out[0]["hello"] == "world"
+
+    assert isinstance(out[1], list)
+    assert out[1] == [1, 2, 3]
+
+
+@pytest.mark.parametrize('compression', [None, 'gzip', 'bz2'])
```
tdhopper, line 87: Wouldn't be required if using a parameterized fixture

blackary:
@martindurant I added entries in the docs for the new plugins -- let me know if there's anything else you'd like me to add

tdhopper (approved):

intake/source/tests/test_json.py
blackary: Thanks, @tdhopper!

intake/source/tests/test_json.py
```diff
@@ -0,0 +1,70 @@
...
+import pytest
+from fsspec import open_files
+from intake.source.jsonfiles import JSONFileSource, JSONLinesFileSource
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+EXTENSIONS = {
+    None: "",
+    "gzip": ".gz",
+    "bz2": ".bz2",
```
blackary, line 22: @martindurant Is there a mapping somewhere from the compression string to the extension string? I'd love to use that, rather than this hand-written mapping, but couldn't find what I was looking for.

martindurant: `fsspec.utils.compressions` is the mapping {extension: internal name} and `fsspec.compression.compr` is the mapping {internal name: implementation function/class}. The latter will only have entries if the relevant library is installed.

intake/source/jsonfiles.py
```diff
@@ -0,0 +1,168 @@
...
+        self.mode = "rt" if text_mode else "rb"
+        self.encoding = text_encoding
+        self._read = read
+
+        super(JSONFileSource, self).__init__(metadata=metadata)
+
+    def read(self):
+        import fsspec
+
+        urlpath = self._get_cache(self._urlpath)[0]
```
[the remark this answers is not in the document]
blackary: To use that, would you recommend prepending simplecache:: on all urlpaths, if not already there?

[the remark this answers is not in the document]
martindurant: Oh no - this puts the caching decision on the catalog maker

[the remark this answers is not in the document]
blackary: Ah, I see. So, would you prefer that new plugins not use the intake cache at all?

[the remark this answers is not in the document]
martindurant: Calling it as you do in the code here is fine, it will not do anything if no Intake cache is defined. In documentation, you can give examples of fsspec-style caching, if it does everything you want it to.

Issue #628, opened 2021-11-22
Remote catalog pmode 'default' value doesn't allow for double nesting of catalogs.


naatebarber:
@ghoersti

We've created a new catalog container type which allows for multiple levels of nested catalog entries. When there was only one level of nesting, the `persist_mode` attribute could be used to disable persistence to the child-catalogs.

```python
cat = RemoteCatalog(uri, auth=auth, persist_mode="never", ttl=10, page_size=50)

print(f"{type(catalog)=}")
print(f"{catalog.pmode=}")

print(f"{type(catalog.large)=}")
print(f"{catalog.large.pmode=}")

catalog.large.translated_datasets
```

Now that we're nesting entries at > 1 levels deep, we can no longer explicitly set the `persist_mode` attribute when remote catalogs are created. This causes an error to appear when trying to access child catalogs with `direct-access: forbid`

```
type(catalog)=<class 'intake.catalog.remote.RemoteCatalog'>
catalog.pmode='never'
type(catalog.large)=<class 'intake.catalog.remote.RemoteCatalog'>
catalog.large.pmode='default'
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/intake/catalog/base.py in __getattr__(self, item)
    349             try:
--> 350                 return self[item]  # triggers reload_on_change
    351             except KeyError as e:

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/intake/catalog/base.py in __getitem__(self, key)
    422             return out()
--> 423         raise KeyError(key)
    424 

KeyError: 'has_been_persisted'

The above exception was the direct cause of the following exception:

AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_43651/2617722241.py in <module>
      5 print(f"{catalog.large.pmode=}")
      6 
----> 7 catalog.large.translated_datasets

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/intake/catalog/base.py in __getattr__(self, item)
    348             # Fall back to __getitem__.
    349             try:
--> 350                 return self[item]  # triggers reload_on_change
    351             except KeyError as e:
    352                 raise AttributeError(item) from e

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/intake/catalog/base.py in __getitem__(self, key)
    397             e._pmode = self.pmode
[...]

PR #629, opened 2021-11-23
Fixes issue with persistence when RemoteCatalogEntries are Catalogs

naatebarber:
Allows for persist mode to be passed appropriately for remote catalogs.
@ghoersti

Fixes https://github.com/intake/intake/issues/628


martindurant:
Thanks for tracking this down.

PR #630, opened 2021-11-24
Allow user parameters throughout catalog

martindurant:
Currently, catalog-level parameters are established as an argument `user_parameters=` to `Catalog`, or a block in the YAML file:
```
metadata:
  parameters:
    var_name:
      type: str
      description: description
      default: something
```

(requires documentation and tests)

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -293,6 +293,9 @@ resolve to ``"http://server:port/user/"``
 Parameter Definition
 --------------------
 
+Source parameters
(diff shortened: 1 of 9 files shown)


martindurant:
Where a catalog is loaded from multiple files, it will aggregate all the user-parameters from the constituents.
Children of a catalog will inherit the user-parameters of its parent.

martindurant:
cc @jhamman , I think you asked for this, but I don't know where the issue is now.

PR #635, opened 2021-12-22
Add tiled

martindurant:
Closes https://github.com/bluesky/tiled/pull/122

Needs many more tests


martindurant:
@danielballan : following a pretty painful search for the bug in the drivers CLI test, I think we should revisit what that does - autodiscover and autodiscover_all are not actually what get called during intake import , so they are not showing the same information. Also, we can drop package scans altogether at this point.

martindurant:
I'll merge this as a good start, but of course can improve - that way this PR won't get in the way of the other one.

PR #636, opened 2021-12-23
Config / registry

martindurant:
Trying to greatly simplify the config and driver registry systems, in particular not to have separate paths for registering drivers via the CLI versus via code.

@danielballan , this is largely stepping on your previous attempt to retialise these things. We could take this opportunity to remove package scanning altogether.

Ref #635

### DIFF
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -15,7 +15,7 @@ from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
 
-from .source import register_driver, registry
+from .source import registry
 from .catalog.base import Catalog
 
 imports = {
@@ -29,7 +29,7 @@ imports = {
     "register_driver": "intake.source:register_driver",
     "unregister_driver": "intake.source:unregister_driver",
 }
-openers = {}
+openers = set()
 logger = logging.getLogger('intake')
 
 
@@ -41,7 +41,7 @@ def __getattr__(attr):
     """
     gl = globals()
     if attr in openers:
-        driver = openers[attr].load()
+        driver = registry[attr[5:]]  # "open_..."
         gl[attr] = driver
     else:
         if attr in gl:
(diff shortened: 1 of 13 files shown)


martindurant:
(oops, got attached to tiled PR)

danielballan:
This all looks reasonable to me.

intake/__init__.py
```diff
@@ -41,7 +41,7 @@ def __getattr__(attr):
     """
     gl = globals()
     if attr in openers:
-        driver = openers[attr].load()
+        driver = registry[attr[5:]]  # "open_..."
```
danielballan, line 44: Might be worth checking that the first five chars really are `open_` to avoid confusing failure modes.

martindurant:
:)

intake/__init__.py
```diff
@@ -69,21 +69,19 @@ def __dir__(*_, **__):
...
             # primitive name normalization
             func_name = re.sub('[-=~^&|@+]', '_', func_name)
         if func_name.isidentifier():
             # stash name for dir() and later fetch
-            openers[func_name] = plugin
+            openers.add(func_name)
         else:
-            warnings.warn('Invalid Intake plugin name "%s" found.' %
-                          plugin.name)
+            warnings.warn('Invalid Intake plugin name "%s" found.', name)
```
danielballan, line 84:
```suggestion
            warnings.warn('Invalid Intake plugin name "%s" found.', name, stacklevel=2)
```

intake/config.py
```diff
@@ -27,53 +27,77 @@
...
-        fn = cfile()
-    if os.path.isfile(fn):
-        with open(fn) as f:
-            try:
-                conf.update(yaml_load(f))
-            except Exception as e:
-                logger.warning('Failure to load config file "{fn}": {e}'
-                               ''.format(fn=fn, e=e))
+class Config(dict):
+    def __init__(self, filename=None, **kwargs):
```
danielballan, line 41: A pattern I like sometimes is doing the file-based path in a classmethod, as in:

```
Config.from_file(filename, **kwargs)
```

with the `**kwargs` for optional overrides of what's in the file. This way, the parsing is moved out of the single privileged initializer. Also, as this subclasses `dict`, it avoids breaking compatibility with the superclass init.

intake/tests/test_top_level.py
```diff
@@ -45,7 +45,7 @@ def tmp_path_catalog():
 
 
 def test_autoregister_open():
-    assert hasattr(intake, 'open_csv')
+    assert intake.open_csv
```
danielballan, line 48: What is the `assert` mean to verify?

martindurant:
Sorry, missed the comments, I'll add those in too.

intake/tests/test_top_level.py
martindurant: Not sure why I changed this one.
The test is to make sure that the open_csv function exists, given that the csv driver should have been auto registered.

intake/config.py
martindurant: You are right, that would be nicer. There are two complication
- there always is a filename, taken from defaults if not otherwise given, so if `__init__` isn't to handle this, then *every* instantiation would go via `from_file`
- that self.filename also appears in save(). But maybe it shouldn't?

PR #639, opened 2022-01-06
fixing issue where .add does not return updates

eli-pinkus:
`LocalCatalogEntry` `.add` function does not return the updated object when a new path is specified. Docstring says to expect new Catalog returned when a file path is passed.


martindurant:
Would you mind adding a test showing this in action?

eli-pinkus:
Yes, I'll take a look w/ @philippjfr and do so.

Thank you for taking a look.

martindurant:
(ping)

eli-pinkus:
@philippjfr implemented a workaround for the usecase. closing.

PR #644, opened 2022-01-21
BUG: allow None as json compression

raybellwaves:
If you have a json without compression, you would use the default of None.

However, this will currently raise a `ValueError` as None is not in `VALID_COMPRESSIONS`.
https://github.com/intake/intake/blob/master/intake/source/jsonfiles.py#L60

This PR fixes this by also checking for a None value.

cc. @blackary 

Don't think CI failures are related.

### DIFF
--- a/intake/source/jsonfiles.py
+++ b/intake/source/jsonfiles.py
@@ -40,8 +40,8 @@ class JSONFileSource(DataSource):
             common
         compression : str or None
             If given, decompress the file with the given codec on load. Can
-            be something like "zip", "gzip", "bz2", or to try to guess from the
-            filename, 'infer'
+            be something like "zip", "gzip", "bz2", "xz". To try to guess from the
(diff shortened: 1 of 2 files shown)


blackary (approved):
LGTM. Wanna add a test case to [tests/test_json.py](https://github.com/intake/intake/blob/master/intake/source/tests/test_json.py)?

raybellwaves:
> LGTM. Wanna add a test case to [tests/test_json.py](https://github.com/intake/intake/blob/master/intake/source/tests/test_json.py)?

Not sure how to. Had a couple of attempts in my last commits.

Falling back to using `infer` as the default

martindurant:
Merging from master should make it go green again?

raybellwaves:
thanks @blackary for the test help

martindurant:
Perfect, thank you

PR #650, opened 2022-02-08
fixed description issue

kthyng:
When saving an overall catalog, the description would not transfer. Now it does.
Also added a test to check for this.


martindurant:
Perfect, thank you.

PR #655, opened 2022-02-25
Cache enhancement

### LINKED ISSUE #653: Error with cache when running in jupyterlab
More description is given in this [notebook example](https://gist.github.com/Hoxbro/ec14fc9e746c4bc1e24068559d809aba).

Hoxbro:
Fixes #653

### DIFF
--- a/intake/source/cache.py
+++ b/intake/source/cache.py
@@ -155,9 +155,7 @@ class BaseCache(object):
             'cache_download_progress', True)
 
         cache_paths = self._from_metadata(urlpath)
-        if cache_paths is not None:
-            all_file_exists = not any(not os.path.exists(c) for c in cache_paths)
-        if cache_paths is None or not all_file_exists:
+        if cache_paths is None or any(not os.path.exists(c) for c in cache_paths):
             files_in, files_out = self._make_files(urlpath)
             self._load(files_in, files_out, urlpath)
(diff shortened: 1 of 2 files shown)


michaelaye:
Ipywidgets is a missing requirement for tqdm though, I don’t think that check should be in intake.

martindurant:
> Ipywidgets is a missing requirement for tqdm though

It's more subtle than that: tqdm works in the terminal and doesn't need ipywidgets, but when in a notebook, it tries to use it. At least, I think so.

martindurant:
I think this is good for fixing the two issues

Hoxbro:
I agree with @martindurant. 

I see `ipywidgets` as an optional dependency for `tqdm` and can be run without it. When https://github.com/tqdm/tqdm/pull/1218 is merged it will also automatic run the `std.tqdm` version instead of the `notebook.tqdm`.

michaelaye:
Sure but I’m confused why intake isn’t leaving the decision if to use a progress bar or not and which one to tqdm? intake shouldn’t have to manage anything about progress bars.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

martindurant:
I agree with you, @michaelaye , but this workaround can keep up going for now. If it is fixed upstream, we can always remove it.

Issue #654, opened 2022-02-25
template interpolation in parameters is broken


JoranDox:
test.yaml
```
metadata:
  version: 1
sources:
    works:
      description: "this one works"
      driver: intake.catalog.local.YAMLFileCatalog
      args:
        path: "{{ env(HOME) }}/example.csv"

    doesnt:
      description: "this one doesn't work"
      driver: intake.catalog.local.YAMLFileCatalog
      # driver: intakewrapper.MHIntakePluginWrapper
      args:
        path: "{{root}}/example.csv"
      parameters:
        root:
          description: "root path"
          type: str
          default: "{{ env(HOME) }}"
```

notebook.ipynb:
```
import intake
catalog = intake.open_catalog("test.yaml")

catalog.works
>>> shows a catalog entry with a valid path ( path: /home/jovyan/example.csv ), with env(HOME) correctly expanded

catalog.doesnt
>>> shows a catalog entry with an invalid path ( path: '{{ env(HOME) }}/example.csv' ), with env(HOME) literally in there
```

![image](https://user-images.githubusercontent.com/7152733/155711669-1a7ff147-d55e-486d-ba05-c0729e36a87f.png)

I'm pretty sure this used to work, as we've got many older catalogs using that interpolation in parameters, but I can't tell you when it broke, as we haven't been using this recently up until now. Has this been removed? Or changed (in that case, how to do it now)? Or shouldn't we do this for some reason?

I've also found the following (from here https://intake.readthedocs.io/en/latest/catalog.html?#templating)
> The following functions may be available. Since these execute code, the user of a catalog may decide whether they trust those functions or not.

But I have no idea how to trust/untrust these things, is that relevant?

martindurant:
The correct syntax, or at least, the one that works, is:
```yaml
          default: "env(HOME)"
```

I cannot find immediately in the docs where the lack of `{{..}}` specifically for parameter values is mentioned. The UserParameter docstring says
```
    default: type value
        same type as ``type``. It a str, may include special functions
        env, shell, client_env, client_shell.
```
which is not exactly clear.

JoranDox:
Thanks for the clarification Martin! Interesting, do you know what the reasoning is behind not having the `{{...}}` there?

Would you mind if I add a part in [the docs here](https://intake.readthedocs.io/en/latest/catalog.html?#parameter-definition) with something like:

> # Parameter Definition
> 
> ## Source parameters
> 
> ...
> 
> ## Catalog parameters
> 
> ...

[...]

PR #656, opened 2022-02-28
add templating parameters section

JoranDox:
attempts to solve https://github.com/intake/intake/issues/654

### DIFF
--- a/docs/source/catalog.rst
+++ b/docs/source/catalog.rst
@@ -381,6 +381,28 @@ For cases where the catalog is not defined in a YAML spec, the argument ``user_p
 to the constructor takes the same form as ``parameters`` above: a dict of user parameters,
 either as ``UserParameter`` instances or as a dictionary spec for each one.
 
+Templating parameters
+''''''''''''''''''
+
+Template functions can also be used in parameters (see `Templating`_, above), but you can use the available functions directly without the extra `{{...}}`.
+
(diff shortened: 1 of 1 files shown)


docs/source/catalog.rst
```diff
@@ -381,6 +381,28 @@ For cases where the catalog is not defined in a YAML spec, the argument ``user_p
 to the constructor takes the same form as ``parameters`` above: a dict of user parameters,
 either as ``UserParameter`` instances or as a dictionary spec for each one.
 
+Templating parameters
+''''''''''''''''''
+
+Template functions can also be used in parameters (see `Templating`_, above), but you can use the available functions directly without the extra `{{...}}`.
+
+For example, this catalog entry uses the ``env("USER")`` functionality as described to set a default based on the user's home directory.
```
martindurant, line 389: You meant HOME here, right? I don't know offhand whether this works on windows...

martindurant: On reflection, I think it can be left as is, and maybe it does indeed work on windows. We are generally interfacing with fsspec anyway, which casts paths to posix-like, even on windows.

martindurant:
> how did my change in the documentation mess up a test?

It looks like maybe a network timeout for tiled to fetch its example datasets. Perhaps rerunning will fix. Otherwise, I'll have a look to see whether there was a new version of tiled and something changed upstream.

martindurant:
Created https://github.com/bluesky/tiled/issues/178

PR #658, opened 2022-03-28
Keep hold of hvplot calls

martindurant:
@AlbertDeFusco @mattkram 

Allows pushing a plot spec back to a source's metadata.

This is unchanged
```
s = intake... # define a source
s.plot.known_plot()  # existing predefined plot
s.plot.line(args)  # existing general plot
```

This is different
```
s.plot.line(plotname="new", args)
```
as it updates `s`'s metadata directly with the new plotname, which then shows up in the source's YAML representation and can be directly added to a catalog.

### DIFF
--- a/intake/source/base.py
+++ b/intake/source/base.py
@@ -84,6 +84,12 @@ class CacheMixin:
         return [c.load(urlpath) for c in self.cache]
 
 
+try:
+    from hvplot import hvPlot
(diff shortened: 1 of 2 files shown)


martindurant:
cc @philippjfr

philippjfr:
Cool idea, I like it! Should this be documented?

PR #661, opened 2022-04-13
Update references to collections for consistency and clarity

jukent:
There are some name space collisions with "collections" vs "catalogs". Originally a catalog was defined using the term "collection", but now there is a "Collection" class. To reduce confusion, I have replaced references to collections with similar terms (catalog when appropriate, inventory, group, etc). 

This effort has been made in `intake-esm` too, and thus references to `intake-esm` in this documentation that used th term "collection" erroneously were out-of-date.

See discussion on https://github.com/intake/intake-esm/pull/457 with @andersy005


martindurant:
+1
Feels like a thesaurus might have been involved :)

PR #662, opened 2022-05-03
fix typos in docs

thomasastanley:


martindurant:
Thanks for these, @thomasastanley

PR #666, opened 2022-06-10
Add a link to the intake-sqlite driver

zaneselvans:
We created a thin wrapper around the `intake-sql` driver which allows access to remote file-based SQLite databases, by first caching them locally with `fsspec` `simplecache::` and then providing access using `intake-sql` to that locally cached database with a normal `sqlite:///` SQLAlchemy URI.


martindurant:
:+1:

PR #670, opened 2022-06-29
fix: removes leading slash for urljoin

pgierz:
Hosting the intake server behind a reverse proxy at a non-standard address leads to mangling of the actual URLs. I am unsure if this fix covers all cases, but the URLs are now correctly translated for my case.

Here is an extracted basic example:

```python
>>> from urllib.parse import urljoin
>>> base_url = "http://server.org/intake/"
>>> bad_url = urljoin(base_url, "/v1/source")
>>> good_url = urljoin(base_url, "v1/source")
>>> print(f"{bad_url=}, {good_url=}")
bad_url='http://server.org/v1/source', good_url='http://server.org/intake/v1/source'
```

### DIFF
--- a/intake/catalog/remote.py
+++ b/intake/catalog/remote.py
@@ -481,7 +481,7 @@ def open_remote(url, entry, container, user_parameters, description, http_args,
                    name=entry,
                    parameters=user_parameters,
                    available_plugins=list(plugin_registry))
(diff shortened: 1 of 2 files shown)


martindurant:
Makes sense, thank you.

martindurant:
Can you please push an empty commit? For some reason the tests didn't run.

Issue #665, opened 2022-05-27
Does intake need the no_duplicates_constructor for yaml?


JoranDox:
The code here https://github.com/intake/intake/blob/master/intake/utils.py#L26 breaks our yaml intake catalog using merges & overwrites. We found https://github.com/intake/intake/issues/269, but I'm currently wondering if there is any reason to have that at all, as python will gladly just overwrite the duplicate key, and in fact this is used in the merge functionality of the pyyaml reader.

So I'm basically wondering if there's any hidden pitfalls if we'd just do 
```
import intake.catalog.local
intake.catalog.local.yaml_load = yaml.safe_load
```
as workaround for this?

(bonus: the "duplicate" keys are under the metadata part of the intake catalog)

martindurant:
I think I meant to say: please make a PR, and let's see if everything still passes!

PR #671, opened 2022-07-05
Use default yaml safe_load instead of no duplicates constructor #665

j-r77:
https://github.com/intake/intake/issues/665

### DIFF
--- a/intake/utils.py
+++ b/intake/utils.py
@@ -23,28 +23,6 @@ def make_path_posix(path):
     return path.replace('\\', '/').replace('//', '/')
 
 
-def no_duplicates_constructor(loader, node, deep=False):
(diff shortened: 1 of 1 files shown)


martindurant:
Since `yaml_load` is not just pass-through to `safe_load`, we could just remove it and refactor usage and/or set `yaml_load = safe_load`. Can you please check that there is a test that still passes because a duplicate identifier raises an exception here?

martindurant:
One useful facet of `no_duplicate_yaml()` was to interpret python tuples (`!!python/tuple` yaml tag, aka 'tag:yaml.org,2002:python/tuple') as part of safe loading. This is really useful! I don't think there's any way to get this while using the builtin function.

PR #685, opened 2022-09-07
Add CSV output to intake get CLI

### LINKED ISSUE #684: intake get output CSV
The [docs](https://intake.readthedocs.io/en/latest/tools.html#get) say this for `intake get`:

> Given the name of a catalog entry, this subcommand outputs the entire data source to standard output.

But when I run it I see the string representation of the Pandas DataFrame, not the entire dataset as CSV:

```
$ intake get catalogs/bodleian.yaml turkish
                                                  id  ...                                           contents
0  https://iiif.bodleian.ox.ac.uk/iiif/manifest/a...  ...                                                NaN
0  https://iiif.bodleian.ox.ac.uk/iiif/manifest/2...  ...                                                NaN
0  https://iiif.bodleian.ox.ac.uk/iiif/manifest/a...  ...                                                NaN
0  https://iiif.bodleian.ox.ac.uk/iiif/manifest/9...  ...                                                NaN
0  https://iiif.bodleian.ox.ac.uk/iiif/manifest/2...  ...                                                NaN
0  https://iiif.bodleian.ox.ac.uk/iiif/manifest/e...  ...                                                NaN
0  https://iiif.bodleian.ox.ac.uk/iiif/manifest/6...  ...                                                NaN
0  https://iiif.bodleian.ox.ac.uk/iiif/manifest/c...  ...  [Persian ghazals (ff. 1b-16b). Nesimi, Dīvān-i...

[8 rows x 30 columns]
```

Would a PR to implement CSV output to STDOUT or an optional file be a welcome addition?

PS. Thank you for a beautiful and extremely useful data tool!

edsu:
This small change adds the ability to get CSV output from the `intake get` CLI . Previously it was printing out the string representation of a DataFrame. It also adds a `--output` option for writing to a file instead of stdout.

Closes #684

### DIFF
--- a/intake/cli/client/subcommands/get.py
+++ b/intake/cli/client/subcommands/get.py
@@ -16,6 +16,7 @@ log = logging.getLogger(__name__)
 #-----------------------------------------------------------------------------
 
 # Standard library imports
+import sys
 
 # External imports
 
@@ -28,7 +29,7 @@ from intake.cli.util import Subcommand
 #-----------------------------------------------------------------------------
 
 class Get(Subcommand):
-    ''' Get a catalog entry
+    ''' Get a catalog entry as CSV
 
     '''
 
@@ -37,8 +38,14 @@ class Get(Subcommand):
     def initialize(self):
         self.parser.add_argument('uri', metavar='URI', type=str, help='Catalog URI')
         self.parser.add_argument('name', metavar='NAME', type=str, help='Catalog name')
+        self.parser.add_argument('--output', metavar='FILE', type=str, help='Output file')
 
     def invoke(self, args):
         catalog = open_catalog(args.uri)
         with catalog[args.name] as f:
-            print(f.read())
+            df = f.read()
+            if args.output:
+                out = args.output
+            else:
+                out = sys.stdout
+            df.to_csv(out, index=False)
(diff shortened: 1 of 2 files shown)


martindurant:
The documentation of the parameter doesn't really say that the output will be CSV (as opposed to free text or some other format). Could you perhaps be a bit more verbose?

edsu:
Thanks for the feedback. I made a small update to the docs and the help text. Does that help?

intake/cli/client/subcommands/get.py
```diff
@@ -28,7 +29,7 @@
 #-----------------------------------------------------------------------------
 
 class Get(Subcommand):
-    ''' Get a catalog entry
+    ''' Get a catalog entry as CSV
```
martindurant, line 32: Oh I see, so this changes the default case even when not writing to a file. I wonder if the original formatted text version isn't also useful, what do you think?

edsu: I don't think it's useful assuming there is a CSV output, but maybe someone somewhere disagrees? 
 😃

I could refactor to conditionally output csv if you want?

martindurant:
Actually, it occurs to me that this will simply fail if the dataset is not a dataframe type! We should at the very least check, perhaps with hasattr(to_csv).

edsu:
Ok, I put in a guard against `to_csv()` not existing. Due to my own narrow use of Intake I hadn't really considered use cases where the container type isn't a DataFrame! I guess it might make sense to close this PR without merging?

PR #689, opened 2022-10-27
allow for ttl being None

kthyng:
I saw in the docs that TTL can be None to not expire, but if a user selects that there is an error. This small change allows for TTL=None.


kthyng:
Thanks!

PR #690, opened 2022-11-01
fix deprecation for distutils Version classes

raybellwaves:
Was getting

```
  /Users/ray/miniforge3/envs/main/lib/python3.9/site-packages/intake/source/utils.py:11: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    DASK_VERSION = LooseVersion(dask.__version__)
```

Copied the solution in xarray https://github.com/pydata/xarray/pull/6096/files


martindurant:
Wow, we needed that in a lot of places, thank you!

PR #691, opened 2022-11-04
Move msgpack_numpy import so tests can pass if not installed

jim22k:
When running tests, I didn't have `msgpack_numpy` installed and tests failed.

This PR moves one import to avoid the error and skips a test when `msgpack_numpy`
is not installed.

It also adds a note to one test that I ran into on a new Mac. This is not a bug in the test,
but took me a while to figure out the root cause (Mac was stealing the default port for
use by AirPlay Receiver).

### DIFF
--- a/intake/cli/server/tests/test_server.py
+++ b/intake/cli/server/tests/test_server.py
@@ -257,6 +257,9 @@ def multi_server(tmpdir):
     shutil.rmtree(tmpdir)
 
 
+# If this tests fails on Mac running Monterey or newer,
+# Look at this solution to Apple using port 5000:
+# https://developer.apple.com/forums/thread/682332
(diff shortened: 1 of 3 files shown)


intake/cli/server/tests/test_server.py
```diff
@@ -257,6 +257,9 @@ def multi_server(tmpdir):
     shutil.rmtree(tmpdir)
 
 
+# If this tests fails on Mac running Monterey or newer,
+# Look at this solution to Apple using port 5000:
+# https://developer.apple.com/forums/thread/682332
 def test_flatten_flag(multi_server):
```
martindurant, line 263: I suppose in the long run we ought to have a fixture picking random ports and retrying if the port is not available. Now now, though.

PR #693, opened 2022-11-07
Fix bug in GUI when customizing a new plot

jim22k:
For the dataframe container, attempting to "customize" a new plot resulted in a KeyError inside `dfviz`. This fixes the issue.

Additionally, this PR allows the GUI to be opened for the non-builtin dataset prior to calling `intake.gui`.
This will now work:
```
import intake
ds = intake.open_catalog(...)
intake.interface.gui.GUI([ds])
```

### DIFF
--- a/intake/catalog/tests/test_catalog_save.py
+++ b/intake/catalog/tests/test_catalog_save.py
@@ -1,7 +1,7 @@
 """
 Test saving catalogs.
 """
-
+import os
 import intake
 from intake.catalog import Catalog
 from intake.catalog.local import LocalCatalogEntry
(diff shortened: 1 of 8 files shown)


intake/catalog/tests/test_catalog_save.py
```diff
@@ -10,6 +10,8 @@
 def test_catalog_description():
```
martindurant, line 10: Was this test failing? It'd be good to make a test that would have caught the missing key in `__init__`.

jim22k: It wasn't failing, but was leaving the created test catalog in a non-temporary directory.

intake/catalog/tests/test_catalog_save.py
```diff
@@ -10,6 +10,8 @@
...
     """Make sure the description comes through the save."""
 
+    NAME = 'desc_test.yaml'
```
martindurant, line 13: OK, got it - so this should be a tmpdir/tmpfile location

jim22k: Changed this to use tmpdir and removed the explicit delete.

PR #695, opened 2022-11-14
Editable plots!

jim22k:
Create, rename, edit, delete plots -- all from the GUI.

If the catalog is a YAMLFileCatalog, changes will be saved.

Also, the default plot creation library for dataframes is now hvplot.explorer rather than dfviz.

### No plot selected -- changed "Customize" to "Create"
No plot selected

### Creation uses hvplot.explorer
Create new plot

### When viewing a plot, new options are available
Edit existing plot

### Dialog for renaming a plot
Rename plot

### DIFF
--- a/intake/catalog/entry.py
+++ b/intake/catalog/entry.py
@@ -74,7 +74,7 @@ class CatalogEntry(DictSerialiseMixin):
                 'always', 'never', 'default']:
             raise ValueError('Persist value (%s) not understood' % persist)
         persist = persist or self._pmode
-        s = self.get(**kwargs)  # Should this live in an `else` below the `if`?
+        s = self.get(**kwargs)
         if persist != 'never' and isinstance(s, PersistMixin) and s.has_been_persisted:
             from ..container.persist import store
 
--- a/intake/catalog/entry.py
+++ b/intake/catalog/entry.py
@@ -74,7 +74,7 @@ class CatalogEntry(DictSerialiseMixin):
                 'always', 'never', 'default']:
             raise ValueError('Persist value (%s) not understood' % persist)
         persist = persist or self._pmode
-        s = self.get(**kwargs)
+        s = self.get(**kwargs)  # Should this live in an `else` below the `if`?
         if persist != 'never' and isinstance(s, PersistMixin) and s.has_been_persisted:
             from ..container.persist import store
 
--- a/intake/interface/source/defined_plots.py
+++ b/intake/interface/source/defined_plots.py
@@ -13,6 +13,12 @@ except ImportError:
     dfviz = False
 
 try:
+    import hvplot
+    assert LooseVersion(hvplot.__version__) >= LooseVersion("0.8.1")
+except ImportError:
+    hvplot = False
+
+try:
     import xrviz
     from xrviz.dashboard import Dashboard as XRViz
     assert LooseVersion(xrviz.__version__) >= LooseVersion("0.1.1")
@@ -21,7 +27,7 @@ except ImportError:
 
 import panel as pn
 from ..base import BaseView
-from ...catalog.local import LocalCatalogEntry
+from ...catalog.local import LocalCatalogEntry, YAMLFileCatalog
 
 
 class Plots(BaseView):
@@ -32,13 +38,12 @@ class Plots(BaseView):
     ----------
     source: intake catalog entry, or list of same
         source to describe in this object
+    edit_callback: callback to alert that plot has been edited
 
     Attributes
     ----------
     has_plots: bool
         whether the source has plots defined
-    instructions_contents: str
-        instructions to put on the plot selector label
     options: list
         plots options defined on the source
     selected: str
@@ -54,23 +59,37 @@ class Plots(BaseView):
     """
     select = None
 
-    def __init__(self, source=None, **kwargs):
-        self.custom = pn.widgets.Button(name='Customize...', width_policy='min', align='center')
+    def __init__(self, source=None, edit_callback=None, **kwargs):
+        self.custom = pn.widgets.Button(name='Create', width=70, align='center')
         self.source = source
+        self.callback_on_edit = edit_callback
         self.panel = pn.Column(name='Plot', width_policy='max', margin=0)
+        self._behavior_callback = None
         super().__init__(**kwargs)
 
     def setup(self):
         self.instructions = pn.pane.Markdown(
-            self.instructions_contents, align='center', width_policy='max')
+            "**Select Plot**", align='center', width=70)
         self.select = pn.widgets.Select(options=self.options, height=30,
                                         align='center', min_width=200)
+        self.edit_options = pn.widgets.Select(options=['  Edit', '  Clone', '  Rename', '  Delete'],
+                                              align='center',
+                                              width=25, margin=(5, -10))
         self.pane = pn.pane.HoloViews(self._plot_object(self.selected),
                                       name="Plot")
 
+        self.interact_label = pn.pane.Markdown('Name', align='center', width_policy='max', min_width=100)
+        self.interact_name = pn.widgets.TextInput(placeholder='Name of new plot...')
+        self.interact_cancel = pn.widgets.Button(name="Cancel", width=100)
+        self.interact_save = pn.widgets.Button(name="Save", button_type='primary', width=100)
+
         self.watchers = [
-            self.select.param.watch(self.callback, ['options', 'value']),
-            self.custom.param.watch(self.interact, ['clicks'])
+            self.select.param.watch(self.plot_selected, ['value']),
+            self.custom.param.watch(self.interact, ['clicks']),
+            self.edit_options.param.watch(self.interact, ['value']),
+            self.interact_name.param.watch(self.name_changed, ['value_input']),
+            self.interact_cancel.param.watch(self.cancel, ['clicks']),
+            self.interact_save.param.watch(self.interact_action, ['clicks']),
         ]
         self.out = pn.Row(self.pane, name="Plot")
 
@@ -78,14 +97,24 @@ class Plots(BaseView):
             pn.Row(
                 self.instructions,
                 self.select,
-                self.custom
+                self.custom,
+                self.edit_options,
+            ),
+            pn.Row(
+                self.interact_label,
+                self.interact_name,
+                self.interact_cancel,
+                self.interact_save,
             ),
             self.out,
         ]
+        # Set initial visibility
+        self.children[0][-1].visible = False
+        self.children[1].visible = False
 
     @BaseView.source.setter
     def source(self, source):
-        """When the source gets updated, update the the options in
+        """When the source gets updated, update the options in
         the selector"""
         if source and isinstance(source, list):
             source = source[0]
@@ -94,7 +123,9 @@ class Plots(BaseView):
         BaseView.source.fset(self, source)
         if self.select:
             self.select.options = self.options
-        if source and dfviz and source.container == 'dataframe':
+        if source and hvplot and source.container == 'dataframe':
+            self.custom.disabled = False
+        elif source and dfviz and source.container == 'dataframe':
             self.custom.disabled = False
         elif source and xrviz and source.container in ['xarray', 'ndarray', 'numpy']:
             self.custom.disabled = False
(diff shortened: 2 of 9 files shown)


jim22k:
This definitely needs tests added for the new functionality.

philippjfr:
Really neat! Let me know if you ran into any issues or have suggestions for improvements.

kthyng:
Looks awesome!! And looking forward to trying it out.

intake/catalog/entry.py
```diff
@@ -74,7 +74,7 @@ def __call__(self, persist=None, **kwargs):
                 'always', 'never', 'default']:
             raise ValueError('Persist value (%s) not understood' % persist)
         persist = persist or self._pmode
-        s = self.get(**kwargs)
+        s = self.get(**kwargs)  # Should this live in an `else` below the `if`?
```
martindurant, line 77: We need an instance of `s`, with all its parameters, to know if a persisted version exists.

intake/interface/source/defined_plots.py
```diff
@@ -12,6 +12,12 @@
 except ImportError:
     dfviz = False
 
+try:
+    import hvplot
+    assert LooseVersion(hvplot.__version__) >= LooseVersion("0.8.1")
```
martindurant, line 17: `LooseVersion` is deprecated, no? Suggest `packaging.Version`.

Is there an argument for requiring hvplot?

intake/interface/source/defined_plots.py
```diff
@@ -54,38 +59,62 @@ class Plots(BaseView):
...
         self.source = source
+        self.callback_on_edit = edit_callback
         self.panel = pn.Column(name='Plot', width_policy='max', margin=0)
+        self._behavior_callback = None
         super().__init__(**kwargs)
 
     def setup(self):
         self.instructions = pn.pane.Markdown(
-            self.instructions_contents, align='center', width_policy='max')
+            "**Select Plot**", align='center', width=70)
```
martindurant, line 72: Will need to test this on a range of browser sizes if using any absolute numbers for width.

intake/interface/source/defined_plots.py
```diff
@@ -54,38 +59,62 @@ class Plots(BaseView):
...
         self.select = pn.widgets.Select(options=self.options, height=30,
                                         align='center', min_width=200)
+        self.edit_options = pn.widgets.Select(options=['  Edit', '  Clone', '  Rename', '  Delete'],
```
martindurant, line 75: Why the spaces in the labels?

intake/interface/source/defined_plots.py
```diff
@@ -54,38 +59,62 @@ class Plots(BaseView):
...
+            ),
+            pn.Row(
+                self.interact_label,
+                self.interact_name,
+                self.interact_cancel,
+                self.interact_save,
             ),
             self.out,
         ]
+        # Set initial visibility
```
martindurant, line 111: Would appreciate a comment saying items these are

intake/interface/source/defined_plots.py
```diff
@@ -94,7 +123,9 @@ def source(self, source):
         BaseView.source.fset(self, source)
         if self.select:
             self.select.options = self.options
-        if source and dfviz and source.container == 'dataframe':
+        if source and hvplot and source.container == 'dataframe':
```
martindurant, line 126: Have you tried xarrays/arrays? Should ask upstream if that is supported (yet).

intake/interface/source/defined_plots.py
```diff
@@ -94,7 +123,9 @@ def source(self, source):
...
+            self.custom.disabled = False
+        elif source and dfviz and source.container == 'dataframe':
```
martindurant, line 128: dfviz is not well maintained and should probably be retired. This is OK for now, but following the ext release, we will probably retire dfviz so that it instructs users to get latest hvplot instead.

intake/interface/source/defined_plots.py
```diff
@@ -128,15 +152,99 @@ def selected(self, selected):
...
                 self.pane.object = self._plot_object(event.new)
-            if event.name == 'options':
-                self.instructions.object = self.instructions_contents
+            self.custom.name = "Create" if str(self.select.value) == 'None' else "Edit"
+            self.edit_options.visible = self.custom.name == "Edit"
+
+    def name_changed(self, *events):
+        for event in events:
+            if event.name == "value_input":
+                # Empty name not allowed and name cannot already be in use
```
martindurant, line 165: We don't have alerts, bit I wonder if we can inform the user why they cannot save? It could just be carefully covered in docs.

intake/interface/source/defined_plots.py
```diff
@@ -128,15 +152,99 @@ def selected(self, selected):
...
+                if len(event.new) == 0 or event.new in self.options:
+                    self.interact_save.disabled = True
+                else:
+                    self.interact_save.disabled = False
+
+    def cancel(self, _):
+        self.pane.object = self._plot_object(self.selected)
+        self.out[0] = self.pane
+        self.children[0].visible = True
```
martindurant, line 174: Is it worth keeping these children as named attributes, so its more clear which visibilities are set?

intake/interface/source/defined_plots.py
```diff
@@ -128,15 +152,99 @@ def selected(self, selected):
...
+        self.children[1].visible = False
+        self._behavior_callback = None
+
+    def interact_action(self, _):
```
martindurant, line 178: This is the saving of the plot into the catalog? Then the catalog is written to the original location, since that's what default `add()` does.

So this is not a way to "save as...". Do you anticipate enabling that workflow in the GUI? How about taking a catalog which is not YAMLFile, and saving it as YAML (which you can usually do)? 

I think the full set of possible behaviours and how we can fit them in is the most critical thing in this PR. May well be worth asking a design colleague to take a look.

intake/interface/source/defined_plots.py
```diff
@@ -128,15 +152,99 @@ def selected(self, selected):
...
+        # Stop catalog from autoreloading
+        self.source.cat.ttl = None
+        # Perform action
+        self._behavior_callback()
+        # Save catalog (if YAMLFileCatalog)
+        if isinstance(self.source.cat, YAMLFileCatalog):
+            self.source.cat.add(self.source)
+        # Update description
+        self.callback_on_edit()
```
martindurant, line 187: I don't understand how this is supposed to be used. Surely at least `self` should be passed to the func, so that it can find out whatever information it needs.

intake/interface/source/defined_plots.py
```diff
@@ -128,15 +152,99 @@ def selected(self, selected):
...
+        self.out[0] = self.pane
+        self.children[0].visible = True
+        self.children[1].visible = False
+        self._behavior_callback = None
+
+    def interact_action(self, _):
+        # Stop catalog from autoreloading
+        self.source.cat.ttl = None
+        # Perform action
+        self._behavior_callback()
```
martindurant, line 182: We're relying on this having a value, but if an exception happens during an action, it may still be None.

intake/interface/source/defined_plots.py
```diff
@@ -166,6 +279,49 @@ def _plot_object(self, selected):
             if plot_method:
                 return plot_method()
 
+    def _create(self):
+        plot_name = self.interact_name.value
+        # Add plot metadata to both DataSource and CatalogEntry
+        self.source.metadata.setdefault('plots', {})[plot_name] = self._viz.settings()
```
martindurant, line 285: Sorry that you need to do this in two places.

martindurant (commented):
On the first pass, this looks excellent. I have some questions and minor comments, but I thought it may be worth talking with a front end design person about the various different operations we might want to enable in the GUI. Optional.

intake/interface/source/defined_plots.py
philippjfr: We're working on xarray support so it's not supported yet.

intake/interface/source/defined_plots.py
jim22k: Changing to `packaging.version.Version`.

jim22k: I don't know if requiring hvplot makes sense for intake as a whole, but I could see requiring hvplot when someone accesses `intake.gui`. I'll add that so that the GUI experience is better for users.

intake/interface/source/defined_plots.py
jim22k: I'm creating a dropdown selector for additional options, but I want it rendered to look like a down arrow next to the "Edit" button. By adding spaces at the beginning and forcing `width=25`, it gets me what I want. It is a bit of a hack, but it works without requiring changes to panel.

I'll add a comment in the code because it isn't obvious why the spaces exist.

intake/interface/source/defined_plots.py
jim22k: I added comments anywhere I reference `self.children[0]` or `self.children[1]`. I agree those are confusing without a name or comment.

intake/interface/source/defined_plots.py
jim22k: If we make hvplot required, then removing dfviz should be okay.

intake/interface/source/defined_plots.py
jim22k: I'm adding an alert, so we can give useful feedback to the user.
Alert screenshot

intake/interface/source/defined_plots.py
jim22k: Yes, actually, that will make things clearer. I will remove the copious comments and make these named attributes so the code is self-describing.

intake/interface/source/defined_plots.py
jim22k: Good catch. I'll check for `not None` before attempting to call.

intake/interface/source/gui.py
```diff
@@ -154,6 +155,10 @@ def on_click_pars_widget(self, event):
         else:
             self.description.panel.remove(self.pars_editor.panel)
 
+    def on_plot_edited(self):
+        # Manually triggered from plot widget when plot is saved
+        self.description.source = self.sources
```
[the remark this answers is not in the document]
jim22k: Yes, essentially a "redraw" of the widget showing the YAML. I decided to revamp this to look more like a normal widget "watch" rather than passing the callback when creating the Plot.

intake/interface/source/defined_plots.py
jim22k: This is now a proper Event class with access to the Plot and various attributes describing which action was taken.

intake/interface/source/defined_plots.py
jim22k: The correct workflow feels like it should happen at the Sources level. If a source is a local YAML file, then plots become editable and it will save over the existing catalog. If a source is not editable, the plot editing options would be grayed out.

A new functionality needs to be added to a Source to make a copy and/or make a linked wrapper. But this is part of the bigger need for sharing catalogs, nesting catalogs, pointing to other catalogs, etc from the GUI. I think that can wait for a future PR as the scope is quite large.

intake/interface/source/defined_plots.py
jim22k: I tested it at various zoom levels and browsers (Chrome, Firefox, Brave) and it looked correct on all those.
In the worst case, the text will simply wrap and display on two lines, so I think this is a minor concern.

jim22k:
Tests have been added. This should be ready for another review.

intake/interface/source/defined_plots.py
```diff
@@ -1,27 +1,29 @@
 #-----------------------------------------------------------------------------
-# Copyright (c) 2012 - 2019, Anaconda, Inc. and Intake contributors
+# Copyright (c) 2012 - 2022, Anaconda, Inc. and Intake contributors
```
martindurant, line 2: It's fine to change this, but I am told by people who know legal things much better than I that it's not necessary

martindurant:
Is it too late to make a PyData lightning talk out of this?

jim22k:
> Is it too late to make a PyData lightning talk out of this?

Too late, I think. My understanding is that lightning talks at PyData Global were all scheduled weeks ago. No option for last-minute talks.

martindurant:
I think your work here warrants demos wherever they can be published.

PR #710, opened 2023-02-21
Housekeeping

blakerosenthal:
Add pre-commit, black, ruff, and isort. Mostly stuck with defaults except for `line-length = 165`.


martindurant:
+1

Issue #714, opened 2023-02-28
Revising catalog-level GUI


AlbertDeFusco:
I get weird things when I try to open the GUI on a catalog-of-catalogs.

```yaml
# catalogs.yml
sources:
    weather:
        driver: yaml_file_cat
        args:
            path: https://gist.githubusercontent.com/AlbertDeFusco/60a3d2031f89f1fed1d396bb367fe168/raw/401b29419608abd51985f45a81308bc3f4534410/weather.yaml
```

then in the notebook the gui doesn't appear to work.

<img width="705" alt="Screenshot 2023-02-27 at 23 03 45" src="https://user-images.githubusercontent.com/43654/221759070-3d655bb0-cb89-4c0a-9efc-c1affdb02e47.png">

If instead I swap out [`CatalogGUI`](https://github.com/intake/intake/blob/9d9eebd602fde69ed36bb2e0976d2c27564c05f9/intake/catalog/base.py#L484-L492) for the `interface.gui.GUI`

```python
    @property
    def gui(self):
        if not hasattr(self, "_gui"):
            from ..interface import do_import
            from ..interface.gui import GUI
            do_import()
            self._gui = GUI(self)
        else:
            self._gui.visible = True
        return self._gui
```

I get a much better experience with nested catalogs

<img width="870" alt="Screenshot 2023-02-27 at 23 07 28" src="https://user-images.githubusercontent.com/43654/221759476-ff6f6e13-14ad-44e9-8afc-c1303bc06a42.png">

I am interested in making a PR if you think this is reasonable. Also, is this a good use of `do_import()` or should something else be done to ensure that the panel extension is loaded?

martindurant:
Yes, I agree that that catalog and source-level GUIs are probably not fit for purpose and in practice rarely used/poorly documented. Having a single entry point would simplify our life and also allow for the code to be refactored. I don't know the reason that your display looks so weird, though!

We could probably do with a general brainstorming of user expected behaviour around these visual components, especially if they are to be integrated with jlab extensions or online services.

> should something else be done 

There is a `intake.output_notebook` function. do_import adds a little CSS which is probably not needed...

PR #716, opened 2023-03-01
create GUI instance rather than CatalogGUI

AlbertDeFusco:
This addresses #714 . The CatalogGUI instance does not work well if there are nested catalogs. Here's an example catalog.yaml file

```yaml
name: my_local_catalog
sources:
    df:
      driver: csv
      args:
        urlpath: https://bit.ly/autompg-csv
    weather:
        driver: yaml_file_cat
        args:
            path: https://gist.githubusercontent.com/AlbertDeFusco/60a3d2031f89f1fed1d396bb367fe168/raw/401b29419608abd51985f45a81308bc3f4534410/weather.yaml
```

I've included a short screen capture to demonstrate what this change enables. You'll notice `df` in the list of sources in the root of the catalog and the nested `weather` catalog.

ScreenFlow

### DIFF
--- a/intake/catalog/base.py
+++ b/intake/catalog/base.py
@@ -484,8 +484,9 @@ class Catalog(DataSource):
     @property
     def gui(self):
         if not hasattr(self, "_gui"):
-            from ..interface.gui import GUI
             from ..interface import output_notebook
+            from ..interface.gui import GUI
+
             output_notebook()
             self._gui = GUI(self)
         else:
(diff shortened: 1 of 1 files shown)


sophiamyang:
Hi @AlbertDeFusco, @martindurant mentioned that the GUI should look differently for intake 0.6.8? I don't think I see the updated GUI yet : )
image

Issue #717, opened 2023-03-06
Enable hotplugging into Intake registry?


blakerosenthal:
I noticed that `intake.register_driver` adds a new entry to the registry but [make_open_functions](https://github.com/intake/intake/blob/1e26dc1bb9e6beedb683761cbc5ec49a4b6a2b30/intake/__init__.py#L75) is only called on import so my new drivers don't get their `intake.open_*` functions until I reinstall or call `make_open_functions` manually. Would it make sense to build the open function whenever a new driver is registered?

martindurant:
Yes, that makes sense.

PR #719, opened 2023-03-07
Autodiscover 'open_' func when adding to driver registry

blakerosenthal:
Move `make_open_functions()` into DriverSources class as `openers()` method. This will add/remove all the relevant `open_*` functions to intake whenever the registry is updated, not just when intake is first imported.

I needed to rearrange `intake.__getattr__` in order to avoid some circular imports.

Closes https://github.com/intake/intake/issues/717

### DIFF
--- a/intake/source/discovery.py
+++ b/intake/source/discovery.py
@@ -83,14 +83,12 @@ class DriverSouces:
             self._scanned = _package_scan()
         return self._scanned
 
-    @property
     def disabled(self):
         if self._disabled is None:
             self._disabled = [k for k, v in self.conf.get("drivers", {}).items() if v is False]
 
         return self._disabled
 
-    @property
     def registered(self):
         # priority order (decreasing): runtime, config, entrypoints, package scan
         if self._registered is None:
@@ -105,9 +103,8 @@ class DriverSouces:
 
         return self._registered
 
-    @property
     def enabled_plugins(self):
-        return {k: v for k, v in self.registered.items() if k not in self.disabled}
+        return {k: v for k, v in self.registered().items() if k not in self.disabled()}
 
     def register_driver(self, name, value, clobber=False, do_enable=False):
         """Add runtime driver definition to list of registered drivers (drivers in global scope with corresponding ``intake.open_*`` function)
@@ -124,20 +121,20 @@ class DriverSouces:
             If True, unset the disabled flag for this driver
         """
         name = _normalize(name)
-        if name in self.registered and not clobber:
+        if name in self.registered() and not clobber:
             raise ValueError(f"Driver {name} already enabled")
-        if name in self.disabled:
+        if name in self.disabled():
             if do_enable:
-                self.disabled.remove(name)
+                self.enable(name, value)
             else:
                 logger.warning(f"Adding driver {name}, but it is disabled")
 
-        self.registered[name] = value
+        self.registered()[name] = value
 
     def unregister_driver(self, name):
         """Remove runtime registered driver"""
         name = _normalize(name)
-        self.registered.pop(name)
+        self.registered().pop(name)
 
     def enable(self, name, driver=None):
         """
@@ -153,7 +150,16 @@ class DriverSouces:
             Dotted object name, as in ``'intake_xarray.xzarr.ZarrSource'``.
             If None, simply remove driver disable flag, if it is found
         """
-        self.register_driver(name, driver, clobber=True, do_enable=True)
+        config = self.conf
+        if "drivers" not in config:
+            config["drivers"] = {}
+        if driver:
+            config["drivers"][name] = driver
+        elif config["drivers"].get(name) is False:
+            del config["drivers"][name]
+        if name in self.disabled():
+            self.disabled().remove(name)
+        config.save()
 
     def disable(self, name):
         """Disable a driver by name.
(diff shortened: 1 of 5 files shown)


intake/source/discovery.py
```diff
@@ -97,6 +98,22 @@ def enabled_plugins(self):
...
+        """From the current state of ``registry``, create open_* functions"""
+
+        openers = set()
+        for name in self.enabled_plugins():
+            func_name = "open_" + name
+            if not func_name.isidentifier():
+                # primitive name normalization
+                func_name = re.sub("[-=~^&|@+]", "_", func_name)
+            if func_name.isidentifier():
+                # stash name for dir() and later fetch
```
martindurant, line 111: Can you check the overhead from running this function? We are not potentially running it many times, like every time the user does `intake.<tab>`.

blakerosenthal: That's a good catch. I didn't realize `intake.__getattr__` is called so many times (~18) on a single `intake.<tab>`. I'll move this to a cache!

martindurant: 18 sounds excessive! Maybe would be fewer in ipython and even more in raw python console. Still, if 18x the running time is still <1ms (should be), I wouldn't change anything.

martindurant:
That ended up more of a refactor than I had thought :)
Please give me a little time to check through, but first glance looks OK.

blakerosenthal:
> That ended up more of a refactor than I had thought :) Please give me a little time to check through, but first glance looks OK.

Me too. I wanted to `drivers.registered` to reflect the state of all enabled drivers at runtime but realized it only holds plugins registered _at runtime_. Whenever we needed the current state of all the plugins we ran a full scan (entrypoints, config) via `drivers.enabled_plugins()`, but this is also what I needed for the openers. The rearrangement just reflects the cleanup work associated with making `enabled_plugins` a cached instance property rather than an instance method.

intake/source/discovery.py
```diff
@@ -84,21 +83,34 @@ def scanned(self):
             self._scanned = _package_scan()
         return self._scanned
 
-    def enabled_plugins(self):
+    @property
```
martindurant, line 86: Not that I mind, but was there a particular reason for turning these into properties? It makes the diff that bit longer.

intake/source/discovery.py
```diff
@@ -139,14 +153,7 @@ def enable(self, name, driver=None):
...
             If None, simply remove driver disable flag, if it is found
         """
-        config = self.conf
-        if "drivers" not in config:
-            config["drivers"] = {}
-        if driver:
-            config["drivers"][name] = driver
-        elif config["drivers"].get(name) is False:
-            del config["drivers"][name]
-        config.save()
```
martindurant, line 149: enable/disable previously saved the altered config, so that te driver in question remained in the new state for the next session. That means the only way to do this now is via the CLI commands (which edit the config directly). Should save() be at least an option? I don't really see a downside to saving every time: if an ephemeral driver is not found next time, you can't load it whether it's in the enabled list or not.

intake/source/discovery.py
blakerosenthal: Oh good question. I assumed you wouldn't be able to perform direct operations on plain methods (`self.disabled().append(plugin)` for example) but I didn't test that assumption!

martindurant: Yes, you totally can, since it gives you a mutable container (list?). With properties, it would be more appropriate to have a setter method, I suppose, but far too verbose here.

blakerosenthal: Nice, TIL! I'll turn them back into methods :)

intake/source/discovery.py
blakerosenthal: That makes sense. Currently the config is the only record of disabled plugins, so if the user decides to disable a runtime plugin or a discovered plugin then the change is reflected in the config. I was trying to make it so disabling a runtime plugin didn't affect the next session. Would it make sense to only save() if the plugin came from the config, or is the existing behavior preferable?

martindurant: I think if a runtime plugin is registered and then explicitly disabled, it's OK to save that in the config - if the config is then loaded in a new session without the plugin, it won't show up any way. Re-registering it will call enable again automatically?

intake/source/discovery.py
```diff
@@ -84,21 +83,31 @@ def scanned(self):
             self._scanned = _package_scan()
         return self._scanned
 
-    def enabled_plugins(self):
+    def disabled(self):
+        if self._disabled is None:
+            self._disabled = [k for k, v in self.conf.get("drivers", {}).items() if v is False]
+
+        return self._disabled
```
martindurant, line 90: Maybe it should be a set to prevent us from accidentally adding the same name twice?

intake/source/discovery.py
```diff
@@ -158,11 +171,16 @@ def disable(self, name):
         name : string
             As in ``'zarr'``
         """
+        name = _normalize(name)
         config = self.conf
         if "drivers" not in config:
             config["drivers"] = {}
         config["drivers"][name] = False
         config.save()
+        self.disabled().append(name)
```
martindurant, line 180: if disabled gives a set, this would be add()

intake/source/discovery.py
martindurant: A test of this would be to disable a given name twice and then enable it once - is is still in the disabled list?

martindurant (commented):
One very minor point

intake/source/discovery.py
blakerosenthal: This should definitely be a set

PR #726, opened 2023-04-03
DOC: update plugin directory link

raybellwaves:


intake/catalog/local.py
```diff
@@ -246,7 +246,7 @@ def _create_open_args(self, user_parameters):
             raise ValueError(
                 "No plugins loaded for this entry: %s\n"
                 "A listing of installable plugins can be found "
-                "at https://intake.readthedocs.io/en/latest/plugin"
+                "at https://intake.readthedocs.io/en/latest/plugin-directory.html"
```
martindurant, line 249: It's on the next line, no?

raybellwaves: you're right. sorry for the noise

PR #729, opened 2023-04-17
DataFrame pipeline transforms

AlbertDeFusco:
I've put together a new transform class `DataFramePipeline` that accepts a list of steps, where each step is a member method of the previous step.

For example the following catalog has a CSV source to ready the data and a DataFramePipeline source to perform a groupby and return selected columns.

```yaml
sources:
  auto:
    driver: csv
    args:
      urlpath: https://bit.ly/auto-mpg-csv
  avg_mpg_hp_by_origin:
    driver: intake.source.derived.DataFramePipeline
    args:
      targets:
        - auto
      steps:
        - method: groupby
          kwargs:
            by: origin
        - method: cols
          kwargs:
            columns:
              - hp
              - mpg
        - method: mean
```

Here's an example of reading from the catalog

```python
In [1]: import intake

In [2]: cat = intake.open_catalog('pipeline.yaml')

In [3]: cat.avg_mpg_hp_by_origin.to_dask().compute()
Out[3]: 
                hp        mpg
origin                       
Asia     79.835443  30.450633
Europe   80.558824  27.602941
US      119.048980  20.033469
```

### DIFF
--- a/intake/source/derived.py
+++ b/intake/source/derived.py
@@ -308,18 +308,6 @@ class DataFramePipeline(DataFrameTransform):
     container = "dataframe"
     required_params = ["steps"]
 
-    class DataFrameStep(object):
-        def __init__(self, method):
-            self.method = method
-
-        def __call__(self, df, **kwargs):
-            if callable(self.method):
-                pass
-            elif self.method == "cols":
-                return df.__getitem__(kwargs["columns"])
-            else:
-                return getattr(df, self.method)(**kwargs)
-
     def __init__(self, steps, **kwargs):
         kwargs.update(transform=self.pipeline, steps=steps, transform_kwargs={})
         super().__init__(**kwargs)
@@ -333,6 +321,10 @@ class DataFramePipeline(DataFrameTransform):
                 df = method(df, **kwargs)
             elif method == "cols":
                 df = df.__getitem__(kwargs["columns"])
+            elif method in ("apply", "transform"):
+                kwargs_func = kwargs.pop("func")
+                func = kwargs_func if callable(kwargs_func) else import_name(kwargs_func)
+                df = df.apply(func, **kwargs)
             else:
                 try:
                     func = getattr(df, method)
--- a/intake/source/derived.py
+++ b/intake/source/derived.py
@@ -268,3 +268,76 @@ class Columns(DataFrameTransform):
 
     def pick_columns(self, df):
         return df[self._params["columns"]]
+
+
+class DataFramePipeline(DataFrameTransform):
+    """Apply a sequence of transformations over a DataFrame
+
+    The sequence of steps is defined as a list-of-dicts under
+    the key "steps". Each step requires a "method" key and
+    can optionally take "kwargs".
+
+    A special method called 'cols' enables column selection
+    and takes the kwarg 'columns'.
+
+    If a method cannot be found as an attribute of the output
+    of the previous step, the pipeline will attempt to import
+    the method and apply it.
+
+    Here's an example of performing a groupby, selecting two
+    columns and computing the mean.
+
+    by_origin:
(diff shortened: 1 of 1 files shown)


intake/source/derived.py
```diff
@@ -268,3 +268,68 @@ def __init__(self, columns, **kwargs):
...
+    def __init__(self, steps, **kwargs):
+        kwargs.update(transform=self.pipeline, steps=steps, transform_kwargs={})
+        super().__init__(**kwargs)
+
+    def pipeline(self, df):
+        """Apply pipeline steps"""
+        for step in self._params["steps"]:
+            method = step["method"]
+            kwargs = step.get("kwargs", {})
+            if callable(method):
```
martindurant, line 320: It would never be a callable unless constructed dynamically in a session, right? Or did you think to use `import_name` here?

intake/source/derived.py
```diff
@@ -268,3 +268,68 @@ def __init__(self, columns, **kwargs):
...
+                df = method(df, **kwargs)
+            elif method == "cols":
```
martindurant, line 322: loc/iloc also take special syntax

intake/source/derived.py
```diff
@@ -268,3 +268,68 @@ def __init__(self, columns, **kwargs):
...
+        steps:
+          - method: groupby
+            kwargs:
+              by: origin
+          - method: cols
+            kwargs:
+              columns:
+                - hp
+                - mpg
+          - method: mean
```
martindurant, line 304: Is it OK if the output of the source is not a dataframe? I think yes.

martindurant (commented):
Great addition, thank you. Do you intend to add any tests?

AlbertDeFusco:
Here's some more examples with accessors, assign, and join

```yaml
sources:
  auto:
    driver: csv
    args:
      urlpath: https://bit.ly/auto-mpg-csv

  avg_mpg_hp_by_origin:
    driver: intake.source.derived.DataFramePipeline
    args:
      targets:
        - auto
      steps:
        - method: groupby
          kwargs:
            by: origin
        - method: cols
          kwargs:
            columns:
              - hp
              - mpg
        - method: mean

  year:
    driver: intake.source.derived.DataFramePipeline
    args:
      targets:
        - auto
      steps:
        - method: cols
          kwargs:
            columns: yr
        - method: __add__
          kwargs:
            other: 1900

  make_and_model:
    driver: intake.source.derived.DataFramePipeline
    args:
      targets:
        - auto
      steps:
        - method: cols
          kwargs:
            columns: name
        - method: str.split
          kwargs:
            expand: true
            n: 1
        - method: rename
          kwargs:
            columns:
              0: 'make'
              1: 'model'

  auto_cleaned:
    driver: intake.source.derived.DataFramePipeline
    args:
      targets:
        - auto
        - year
        - make_and_model
      steps:
        - method: join
          kwargs:
            other: make_and_model
        - method: assign
          kwargs:
            yr: year
        - method: drop
          kwargs:
            columns: name
```

intake/source/derived.py
AlbertDeFusco: Sure!

intake/source/derived.py
martindurant: > dask is not always happy with the general use of loc unless you have divisions on the index

Yes, this is exactly right

intake/source/derived.py
AlbertDeFusco: I think that's the case. I've not practiced making a pipeline dynamically, I've found trying to do that somewhat difficult. I'll give it a try again.

Later on I have a fallback to import_name if the method doesn't look like a member method or special case.

AlbertDeFusco:
Great. I've been working on the docs, but I'll hold off for a second PR.

martindurant:
Broken by pyarrow 12, please bear with me

Issue #730, opened 2023-04-24
 intake.open_thredds_cat shows error


tarankalra1:
i have installed intake via pip (version 0.6.5). I am getting an error while opening a sample xml file. 

<img width="1163" alt="Screen Shot 2023-04-24 at 3 51 47 PM" src="https://user-images.githubusercontent.com/10886837/234101596-1160ada2-d661-4648-abdd-e83ccd9240df.png">

martindurant:
Do you have intake-thredds installed?

tarankalra1:
that was quick. thank you and that helped.

martindurant:
@blakerosenthal : perhaps an attribute lookup of the form `intake_*` should, if unsuccessful, direct users to the intake plugin registry?

blakerosenthal:
Whoops, sorry I missed this notification. I think this is a great idea. Intake could do better about displaying missing libraries in general.

PR #731, opened 2023-04-28
Update missing driver error message

blakerosenthal:
Adds link to plugin directory if user tries to call `intake.open_*` or `intake.open_catalog(url, driver=...)` with an unregistered driver

Closes https://github.com/intake/intake/issues/730

### DIFF
--- a/intake/__init__.py
+++ b/intake/__init__.py
@@ -61,9 +61,18 @@ def __getattr__(attr):
         gl[attr] = mod
         return mod
 
-    if attr[:5] == "open_" and attr[5:] in registry.drivers.enabled_plugins():
-        driver = registry[attr[5:]]  # "open_..."
-        return driver
+    if attr[:5] == "open_":
+        if attr[5:] in registry.drivers.enabled_plugins():
+            driver = registry[attr[5:]]  # "open_..."
+            return driver
+        else:
+            registered_methods = [f"open_{driver}" for driver in registry.drivers.enabled_plugins()]
+            raise AttributeError(
+                f"Unknown open method '{attr}'. "
+                "Do you need to install a new driver from the plugin directory? "
+                "https://intake.readthedocs.io/en/latest/plugin-directory.html\n"
+                f"Registered opener methods: {registered_methods}"
+            )
 
     raise AttributeError(attr)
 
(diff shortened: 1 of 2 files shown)


intake/__init__.py
```diff
@@ -145,5 +154,10 @@ def open_catalog(uri=None, **kwargs):
     if "_file" not in driver:
         kwargs.pop("fs", None)
     if driver not in registry:
-        raise ValueError("Unknown catalog driver (%s), supply one of: %s" % (driver, list(sorted(registry))))
+        raise ValueError(
+            f"Unknown catalog driver '{driver}'. "
+            "Do you need to install a new driver from the plugin directory? "
+            "https://intake.readthedocs.io/en/latest/plugin-directory.html\n"
+            f"Current registry: {list(sorted(registry))}"
```
martindurant, line 161: We could in theory filter this list (which might be long) for only those that derive from `Catalog`. That would mean actually importing them, so I am slightly against it.

blakerosenthal: That might be less confusing, I agree. But yeah we'd possibly need to import a bunch of stuff just to display an error. Alternatively we don't have to show the list at all and could say something like `"Use list(registry) to see currently loaded drivers"`.

Getting into over-engineering territory here, but the `registry` could also have dynamic a `.catalog_drivers` store.

martindurant: > Getting into over-engineering territory here, but the registry could also have dynamic a .catalog_drivers store.

Maybe if someone asks for it! Most catalog drivers have "cat" in the name anyway, so on reflection, I think we should leave it as it is.

intake/__init__.py
```diff
@@ -61,9 +61,18 @@ def __getattr__(attr):
         gl[attr] = mod
         return mod
 
-    if attr[:5] == "open_" and attr[5:] in registry.drivers.enabled_plugins():
-        driver = registry[attr[5:]]  # "open_..."
-        return driver
+    if attr[:5] == "open_":
+        if attr[5:] in registry.drivers.enabled_plugins():
+            driver = registry[attr[5:]]  # "open_..."
```
martindurant, line 66: FYI: the other place this kind of thing is currently raised is when accessing a data source from a catalog with unknown driver (actually, there can be a list of drivers, so this implies none of them were reachable). 
```
            raise ValueError(
                "No plugins loaded for this entry: %s\n"
                "A listing of installable plugins can be found "
                "at https://intake.readthedocs.io/en/latest/plugin"
                "-directory.html ." % self._driver
            )
```
in intake.catalog.local.LocalCatalogEntry._create_open_args

PR #734, opened 2023-05-11
Dedaskify CSV

blakerosenthal:
Attempt at removing Dask as default reader for CSVSource.

Tries to be smart about checking whether Dask DF `source._dask_df` has been created for the source -- if so, call `source._dask_df.compute()` as before. Otherwise, use pandas.

_get_partition(i) treats each input file as a partition if using the pandas backend, otherwise uses `dask.get_partition(i)`. This could potentially lead to inconsistent reads, for example if user calls `source._get_partition(i)`, invoking the pandas backend (and reading `source._files[i]`), then later calls `source.to_dask()`: `source._get_partition(i)` will likely return a different result. Is this a reasonable problem to expect?

Also I suspect that caching pandas dataframes into `source._pandas_dfs` is not a good idea for memory reasons. Should I leave it as-is, make it optional, or remove it?

### DIFF
--- a/intake/source/csv.py
+++ b/intake/source/csv.py
@@ -150,16 +150,18 @@ class CSVSource(base.DataSource, base.PatternMixin):
             df_part = pd.read_csv(url_part, storage_options=self._storage_options, **self._csv_kwargs)
 
         else:
-            drop_path_column = "include_path_column" not in self._csv_kwargs
+            include_path_column = "include_path_column" in self._csv_kwargs
             path_column = self._path_column()
 
-            df_part = pd.read_csv(url_part, storage_options=self._storage_options, **self._csv_kwargs)
+            csv_kwargs = self._csv_kwargs
+            csv_kwargs.pop("include_path_column")
+            df_part = pd.read_csv(url_part, storage_options=self._storage_options, **csv_kwargs)
 
             # add the new columns to the dataframe
             self._set_pattern_columns(path_column)
 
-            if drop_path_column:
-                df_part = df_part.drop([path_column], axis=1)
+            if include_path_column:
+                df_part[path_column] = url_part
 
         self._pandas_dfs[i] = df_part
         return df_part
--- a/intake/source/csv.py
+++ b/intake/source/csv.py
@@ -5,6 +5,8 @@
 # The full license is in the LICENSE file, distributed with this software.
 # -----------------------------------------------------------------------------
 
+from fsspec.core import get_fs_token_paths
+
 from . import base
 from .utils import reverse_formats, unique_string
 
@@ -121,8 +123,6 @@ class CSVSource(base.DataSource, base.PatternMixin):
         if self.pattern is None and not glob_in_path:
             self._files = urlpath
         else:
-            from fsspec.core import get_fs_token_paths
-
             self._files = get_fs_token_paths(urlpath)[2]
 
         return sorted(self._files)
--- a/intake/source/csv.py
+++ b/intake/source/csv.py
@@ -51,7 +51,8 @@ class CSVSource(base.DataSource, base.PatternMixin):
         self.urlpath = urlpath
         self._storage_options = storage_options
         self._csv_kwargs = csv_kwargs or {}
-        self._dataframe = None
+        self._dask_df = None
+        self._pandas_dfs = None
 
         super(CSVSource, self).__init__(metadata=metadata)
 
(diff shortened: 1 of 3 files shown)


martindurant:
> Also I suspect that caching pandas dataframes into source._pandas_dfs is not a good idea for memory reasons. Should I leave it as-is, make it optional, or remove it?

Probably should stash just enough to be able to say something about the source, i.e., df[:0] or df.dtypes or something. It has never been structly codified what discover() should return, so it's pretty free.

intake/source/csv.py
```diff
@@ -49,23 +49,25 @@ def __init__(self, urlpath, csv_kwargs=None, metadata=None, storage_options=None
...
+        self._storage_options = storage_options or {}
         self._csv_kwargs = csv_kwargs or {}
-        self._dataframe = None
+        self._dask_df = None
+        self._files = None
+        self._schema = None
 
         super(CSVSource, self).__init__(metadata=metadata)
 
     def _set_pattern_columns(self, path_column):
```
blakerosenthal, line 60: @martindurant Does this need to happen for pandas too?

intake/source/csv.py
```diff
@@ -81,46 +83,114 @@ def _path_column(self):
...
+            urlpath = [urlpath]
+
+        glob_in_path = any("*" in path for path in urlpath)
+        if self.pattern is None and not glob_in_path:
+            self._files = urlpath
+        else:
+            import fsspec
+            from fsspec.core import split_protocol
+
+            protocol, _ = split_protocol(urlpath)
```
martindurant, line 126: url_to_filesystem or get_fs_token_paths should cover these three lines

intake/source/csv.py
martindurant: Interesting question. Pandas doesn't have this functionality now, and doesn't load multiple files in one go anyway. Some might in principle be relying on this dask behaviour even when only ever using read(), but it seems unlikely. Let's see how tricky it is to add.

intake/source/csv.py
```diff
@@ -81,46 +83,114 @@ def _path_column(self):
...
+
+        include_path_column = "include_path_column" in self._csv_kwargs
+        path_column = self._path_column()
+
+        csv_kwargs = self._csv_kwargs
+        csv_kwargs.pop("include_path_column")
+        df_part = pd.read_csv(url_part, storage_options=self._storage_options, **csv_kwargs)
+
+        if include_path_column:
+            df_part[path_column] = url_part
```
martindurant, line 178: Would be nice if this can be categorical dtype

martindurant (commented):
Ona first quick scan, everything looks good as we discussed earlier. Posting this now to answer your question.

intake/source/csv.py
blakerosenthal: Not too tricky, assuming I got it right!

intake/source/csv.py
```diff
@@ -81,46 +84,115 @@ def _path_column(self):
...
+        if not isinstance(urlpath, list):
+            urlpath = [urlpath]
 
-        if self._dataframe is None:
-            self._open_dataset(urlpath)
+        glob_in_path = any("*" in path for path in urlpath)
+        if self.pattern is None and not glob_in_path:
+            self._files = urlpath
+        else:
+            from fsspec.core import get_fs_token_paths
```
martindurant, line 124: This can be a top level import, we definitely depend on fsspec.

intake/source/csv.py
```diff
@@ -81,46 +84,115 @@ def _path_column(self):
...
 
     def _get_partition(self, i):
-        self._get_schema()
-        return self._dataframe.get_partition(i).compute()
+        if self._dask_df is not None:
+            # FIX: side effects from having different partition
+            # definitions (dask vs. files)?
+            return self._dask_df.get_partition(i).compute()
+
+        url_part = self.files()[i]
```
martindurant, line 159: +1

intake/source/csv.py
```diff
@@ -81,46 +84,115 @@ def _path_column(self):
...
+        df_part[path_column] = pd.Series(url_part, index=df_part.index, dtype=CategoricalDtype(self.files()))
+        df_part = self._set_pattern_columns(df_part, path_column)
+        if drop_path_column:
+            df_part = df_part.drop([path_column], axis=1)
+
+        return df_part
 
     def read(self):
+        if self._dask_df is not None:
+            return self._dask_df.compute()
```
martindurant, line 186: We may revisit this - there is an argument to not use dask here even if to_dask was previously called.

intake/source/csv.py
blakerosenthal: as in `url_part = self.files()[i+1]`? Dask dataframe [is 0-indexed](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.get_partition.html)...

intake/source/csv.py
blakerosenthal: Yeah this feels a little squirrelly to me. I think the _registered backend_ idea would resolve this by prescribing a reader.

intake/source/csv.py
```diff
@@ -81,46 +84,115 @@ def _path_column(self):
...
+            self._dask_df = self._dask_df.drop([path_column], axis=1)
+
+    def files(self):
+        if self._files is not None:
+            return self._files
 
-    def _get_schema(self):
         urlpath = self._get_cache(self._urlpath)[0]
+        if not isinstance(urlpath, list):
+            urlpath = [urlpath]
```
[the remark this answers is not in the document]
blakerosenthal: Just double checked -- I think we're fully covered by
```
intake/source/tests/test_csv.py
  test_read_pandas
  test_read_list
  test_read_pattern_pandas
  test_read_pattern_with_path_as_pattern_str
  test_read_partition
```

intake/source/csv.py
martindurant: No, "+1" was meant to express agreement (like 👍 ), since this is a specific thing we talked about.

PR #737, opened 2023-05-17
Intake 2.0

martindurant:
This PR got very long and complex, so I wrote up a description in https://github.com/martindurant/intake/blob/reader/README_refactor.md


martindurant:
Some open questions:
- how to handle entrypoints plugins in this system (they would need to be imported before we know what third party imports they need, or even what datatypes they handle)
- is a catalog really a datatype in this sense? This is a standing problem in the existing mode, since Catalog.read() makes no sense.
- how does the server fit in? 
- what about transforms? I guess they declare the instance type they accept or a class function that can decide if it's compatible or not.
- each existing Source maps to a datatype/reader pair, but we would need a shim to cover to_dask()
- can we, then, be backward compatible?
- some third party libraries (xarray, anything imaging) has its own datatype guessing routines we could use

blakerosenthal:
My biggest question is whether or not we're comfortable breaking the API with a v1.0.

martindurant:
@jsignell , @danielballan : I thought you might be interested in an attempted reworking like this.

--

@AlbertDeFusco @mattkram : any thoughts on how this complicates integration? As stated, backward compatibility is possible but tricky.

AlbertDeFusco:
I'll give it some thought to backwards compatibility. If there are new entrypoint names for v1 intake plugins then we can possibly have plugin packages register both formats for as long as the maintainer wishes to support `intake < 1`.

I do like the new datatypes. Would a general APIService base class be useful? Maybe something like this where the requests/httpx session management is handled in the reader? And with transformations you can in a second catalog entry convert from dict -> dataframe or subclass the APIRequest to do other processing after the request is made in a plugin.

```python
@dataclass
class APIService(Service):
    url: str
    method: str
    headers: str | dict
    params: str | dict
    json: str | dict
    verify: bool

class APIRequest(BaseReader):
    imports = {"requests"}  # <-- or a variant that uses httpx
    implements = {datatypes.APIService}
    output_instance = "dict"
    concat_func = # handle paging through concatenation?
    func = "requests:request"
```

martindurant:
> Would a general APIService base class be useful?

Certainly I can imagine something just like that. Every API is different, of course, but they also have much in common. The likes of datasette have shown how useful they can be.

The list of dataclasses I have so far is not meant to be exhaustive, of course. It doesn't even cover all of existing intake; but they are very easy to add. For example, "code" would be an easy type to add, and a python-code reader would produce imported python objects.

I should also add @philippjfr here - in this model, intake would pass a concrete data type (pandas, dask...) and arguments, so hvplot would not have to special-case for intake. We haven't even got into how to actually describe data sources in catalogs yet. Does lumen have any optinion on any of this?

AlbertDeFusco:
I have a question about the converter proposal: `convert(self.data, "dask.dataframe:DataFrame", **kw)`. Is `self.data` the input specification or a data container?

I see in some cases conversion to_dask should build the dataframe from the source files rather than ask to convert an in-memory object like a Pandas dataframe. Am I thinking about this correctly?

martindurant:
> Is self.data the input specification or a data container?

As written, the converted are intending to work on the outputs of a reader's read() method, so instances like pandas.DataFrame (or whatever). 

You are proposing an alternative, wherein we have a reader with a defined a datatype instance (should rename Base to something better, I realise), and we want to use a different reader.

```
new_reader = ReaderClass(data=original_reader.data)
```

This will generally work, except of course you lose the reader-specific arguments, some of which may also apply to the new reader. While e.g., pd.read_csv->dd.read_csv overlap a lot, I don't think we can generally map which arguments can be transferred. That means, that if we have a data instance we wish to make loadable with multiple readers needing specific arguments, we will need to list the arguments for **each** possible reader.

AlbertDeFusco:
👍  Got it, thanks. Would this data instance specification, that could include reader-specific flags, be part of the Catalog YAML specification?

philippjfr:
>I should also add @philippjfr here - in this model, intake would pass a concrete data type (pandas, dask...) and arguments, so hvplot would not have to special-case for intake. We haven't even got into how to actually describe data sources in catalogs yet. Does lumen have any opinion on any of this?

I'm not sure I have sufficient context to really answer that question. Right now Lumen really is designed around pandas/dask DataFrames and that's the only data structures that ever get passed around (each transform has to declare if it supports dask or not). I did have plans to support xarray as well but I've not found anyone to pay for that yet 😄 . 

In terms of describing the data sources, Lumen has no real opinions but it could definitely tap into a better description of the data. Otherwise it basically has to perform fairly expensive queries to get the metadata it needs from the data (you can of course disable this), but effectively it does this:

1. Query the data for all columns and their types
2. Depending on the type query the data for it's range (numerical / dates) or all possible values (string, categorical types)

Using this information it can then automatically construct (widget) filters to query subsets of the data. Still not sure this addresses your question.

martindurant:
Yes, I would say so. Exactly how it would look, I'm not quite sure yet. Do we want the user to have _choices_ at load time:
- "I can load this data with these readers, of which these have specific configuration"
- or specify that a data source is a data-type/reader combination (like the current system) and we have a bunch of helpers in how to make these

martindurant:
Since I have started to write up what exists here ( https://github.com/martindurant/intake/blob/reader/README_refactor.md ), I am now inviting intake-affiliated people that might now know about this effort to have a look and let me know their opinions. As a reminder, this is a design showing all the things that are intended, and is not yet a complete (or even majority complete) replacement for existing Intake.

cc
@raybellwaves @rabernat @blakerosenthal @AlbertDeFusco @jsignell @rsignell-usgs @andersy005 @aaronspring @scottyhq @d70-t @TomAugspurger @danielballan @kthyng

raybellwaves:
This is exciting @martindurant. Nothing much to add from me it all reads well. I like the idea of additional readers for the same dataset. For example when working with grib files I can use intake-xarry (~`xr.open_dataset(fs.open(s3_uri)), engine="cfgrib")` but there are some grib file that aren't cf compliant and I end up using fs.get() and then pygrib. In addition, I would make use of the different readers depending on the compute I was working on. If i'm working on my local machine I may want to read a parquet using dask whereas if i'm on a big EC2 machine I want just to just read it with pandas

kthyng:
@martindurant Thanks for tagging me, and sorry it has taken me awhile to get here. This feels like a much lower level comment than much of the description in your overhaul, but one thing I continually wish for is an easy way to add catalogs together such that their sources are added together. Subtraction or deletion would also be useful. It's hard to imagine the other changes you're describing to comment on them but they do sound pretty impactful.

martindurant:
Yes, the new Catalog implements add and inplace add (+=). I'm not sure how subtract would work....

kthyng:
Excellent. I guess I just mean that being able to remove a source from a Catalog would probably also be handy, not really subtracting two catalogs.

I do use the hvplot plotting capability built into intake currently FWIW. I wouldn't mind it being more flexible though.

martindurant:
> I do use the hvplot plotting capability built into intake currently FWIW. I wouldn't mind it being more flexible though.

Intake 2 has a `.hvplot` registered attribute, which can be encoded in a catalog. It is no longer a magic entry in a source's metadata, but a data product in its own right. You could, in theory, have intermediate plots and more refined versions or even compose them. There is also a matplotlib output (for pandas only right now) which can optionally be piped to PNG as an example.

kthyng:
cool!

pbranson:
I have been looking a bit over the code here - it seems like it might be sufficiently well advanced to start trying out for netcdf, zarr, kerchunk and pandas?

martindurant:
>  it seems like it might be sufficiently well advanced to start trying out for netcdf, zarr, kerchunk and pandas?

It can certainly load all those things I think, and you can put them into YAML catalogs. Do you mean I should busy myself with writing more documentation?

pbranson:
> It can certainly load all those things I think, and you can put them into YAML catalogs. Do you mean I should busy myself with writing more documentation?

Probably just add some links to the new README_refactor.md for your gists would be sufficient for now:
[What the rewrite is for (github.com)](https://gist.github.com/martindurant/0ab2fd6a13a34e93c1995afa2e2cc59b)
[Intake 2 examples (github.com)](https://gist.github.com/martindurant/d5d7b521d0e5fec8adfc5652b8f3242c)

pbranson:
It seems like the merging of the readers and converters may have caused an issue with reading a pipeline:

```
import numpy as np
import xarray as xr
from intake.readers.entry import Catalog
from intake.readers import readers

coord_map = {'time':np.arange(24),'lon':np.arange(0,360,10),'lat':np.arange(0,360,10)}
vars_map = {'u':('time','lon','lat'),
            'v':('time','lon','lat'),
            'w':('time','lon','lat')}
ds = xr.Dataset()
ds = ds.assign_coords(coord_map)
for k,v in vars_map.items():
    ds[k]=(v,np.random.random(tuple([len(ds[d]) for d in v])))
ds.to_netcdf('test.nc')
cat = Catalog()
nc_file = readers.datatypes.HDF5('test.nc')
cat['nc_reader'] = readers.XArrayDatasetReader(nc_file)
cat['subset'] = cat['nc_reader']['u','v']

print(cat)
print(cat.nc_reader.read())

subset = cat.subset.read()
print(subset)
```

Outputs:
```
Catalog
 named datasets: ['nc_reader', 'subset']
<xarray.Dataset>
Dimensions:  (time: 24, lon: 36, lat: 36)
Coordinates:
  * time     (time) int64 0 1 2 3 4 5 6 7 8 9 ... 14 15 16 17 18 19 20 21 22 23
  * lon      (lon) int64 0 10 20 30 40 50 60 70 ... 290 300 310 320 330 340 350
  * lat      (lat) int64 0 10 20 30 40 50 60 70 ... 290 300 310 320 330 340 350
Data variables:
    u        (time, lon, lat) float64 ...
    v        (time, lon, lat) float64 ...
    w        (time, lon, lat) float64 ...
PipelineReader: 

  0: intake.readers.readers:XArrayDatasetReader, () {} => xarray:DataSet
  1: GetItem, (('u', 'v'),) {} => xarray:DataSet
```

Couldn't work out how to realise a pipeline as the dataset.

martindurant:
I think this is fixed in the commit I just pushed. 

BUT, the line should be 
```
cat['subset'] = cat['nc_reader'][['u','v']]
```
(extra square brackets) ,right?

raybellwaves:
As someone who was tagged early on in this PR bravo @martindurant bravo!

danielballan:
Same!

kthyng:
I'm excited to try it out! Way to go!

PR #742, opened 2023-05-20
Allow override of arguments for source from an EntryPoint

### LINKED ISSUE #740: Unable to use user parameters with sources in data package
I seem to be unable to specify user parameters on data sources in a data package.

As a reproducible example, I've created a very simple intake data package [here](https://github.com/dougiesquire/intake_data_package) using an example YAML catalog [taken from the intake docs](https://intake.readthedocs.io/en/latest/catalog.html#templating-parameters):

```yaml
sources:
  variabledefault:
    description: "This entry leads to an example csv file in the user's home directory by default, but the user can pass root='somepath' to override that."
    driver: csv
    args:
      urlpath: "{{root}}/example.csv"
    parameters:
      root:
        description: "root path"
        type: str
        default: "env(HOME)"
```

I register the `variabledefault` source as an entry in `intake.cat` called `mydata`.

The following then works as expected and templates the default value for `root`:

```python
>>> import intake
>>> intake.cat.mydata
sources:
  variabledefault:
    args:
      urlpath: /home/example.csv
    description: This entry leads to an example csv file in the user's home directory
      by default, but the user can pass root='somepath' to override that.
    driver: intake.source.csv.CSVSource
    metadata:
      catalog_dir: /home/intake_data_package/package/
```

However, trying to pass a new `root` as a user parameter returns a `TypeError`:

```python
>>> intake.cat.mydata(root="foo")
  File "<stdin>", line 1, in <module>
  File "/.../intake/source/base.py", line 395, in configure_new
    obj = self._entry(**kw)
          ^^^^^^^^^^^^^^^^^
  File "/.../intake/catalog/entry.py", line 79, in __call__
    s = self.get(**kwargs)
        ^^^^^^^^^^^^^^^^^^
TypeError: EntrypointEntry.get() got an unexpected keyword argument 'root'
```

Is this something that could be made to work? Apologies in advance if this is user error.

martindurant:
Fixes #740 

@dougiesquire : this fixes your situation in as far as you can override any argument for the source. However, you cannot have user parameters, since those are attached to entries, not sources. The thing you point the entry point at is a source and not an entry (which only exist in other catalogs). 

I don't actually know how you have defined the source, so perhaps it would be worth looking at your code to figure that out.

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -818,9 +818,14 @@ class EntrypointEntry(CatalogEntry):
             "container": self._container,
         }
 
-    def get(self):
+    def get(self, **kwargs):
         """Instantiate the DataSource for the given parameters"""
-        return self._entrypoint.load()
+        source = self._entrypoint.load()
+        if kwargs:
+            kw = source._captured_init_kwargs.copy()
+            kw.update(kwargs)
+            source = type(source)(**kw)
+        return source
 
 
 class EntrypointsCatalog(Catalog):
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -818,14 +818,14 @@ class EntrypointEntry(CatalogEntry):
             "container": self._container,
         }
 
-    def get(self, **kwargs):
+    def __call__(self, **kwargs):
         """Instantiate the DataSource for the given parameters"""
         source = self._entrypoint.load()
         if kwargs:
-            kw = source._captured_init_kwargs.copy()
-            kw.update(kwargs)
-            source = type(source)(**kw)
+            source = source.configure_new(**kwargs)
         return source
+
+    get = __call__
 
 
 class EntrypointsCatalog(Catalog):


martindurant:
Actually, I think I've figured it out. Can you please test with this?

dougiesquire:
Yes, that fixed my minimal example [here](https://github.com/dougiesquire/intake_data_package) which registers a `CSVSource`. But I can't get things working when the source inherits from the `Catalog` class. I'm almost certain this is now user error, but here's another stripped back example:

Suppose I have a nested catalog:
```yaml
metadata:
  parameters:
    ver:
      description: "Catalog version"
      type: str
      default: "latest"

sources:
  cat0:
    args:
      path: "{{CATALOG_DIR}}/../source/{{ver}}/cat0.yaml"
    description: "An intake catalog"
    driver: intake.catalog.local.YAMLFileCatalog
    metadata:
      version: "{{ver}}"
```

I'm hoping to be able to register `cat0` in `intake.cat` to provide the following functionality:

- When a user calls `intake.cat.cat0` they get the `"latest"` version
- Users can get specific versions using something like `intake.cat.cat0(ver="another")`

An example of how I've tried to set this up is [here](https://github.com/dougiesquire/intake_data_package/tree/catalog_example). When I try something like `intake.cat.cat0(ver="another")`, I still get back the default (`"latest"`) version.

Note that if simply replace the driver in the YAML above with `intake.source.csv.CSVSource` then the path is templated correctly (with the changes in this PR).

martindurant:
@dougiesquire , can you put the parameters into the source definition rather than the global metadata?

dougiesquire:
That works - thanks! Although

```python
intake.cat.cat0(ver="another")
```

only works if the default parameter is valid. For example, if I change to `default: "doesnotexist"`, then `intake.cat.cat0(ver="another")` fails with:

```
FileNotFoundError: [Errno 2] No such file or directory: '/mypath/../source/doesnotexist/cat0.yaml'
```

Perhaps this is expected/desired behaviour?

martindurant:
Indeed that is expected, but perhaps not desired. When you do `cat.cat0`, the instance is created before the chance to make the amendment in the arguments. In the case of Catalog, this means eagerly loading the file. It could, in theory, be rewritten to only load when an access is made, but this would need some effort.

martindurant:
Merging this now as a useful change. We can follow up on my last point, above, if necessary.

PR #743, opened 2023-05-23
Fix samba share access on Windows

### LINKED ISSUE #739: BUG: cannot .read() from Windows samba share
I have a problem with reading files from shared Windows samba storage.
I can read the catalog `.yml` file and all the metadata, but when I launch `.read()` it replaces the first slash with `c:` and cannot read the content of the file.

Minimal reproducible example:
```python
from pathlib import Path

import intake

DATA_PATH = Path(r"\\path\to\windows\samba\share")
cat = intake.open_catalog(DATA_PATH / "cat.yml")
cat.test_numpy.read()
```

Content of the `cat.yml`
```yaml
description: Samba share test
sources:
  test_numpy:
    description: Test numpy
    driver: numpy
    args:
      path: "{{ CATALOG_DIR }}/numpy.npy"
```

When I launch `print(cat.test_numpy)` everything is ok (except of the number of slashes in the beginning and before `numpy.npy`):
```yaml
sources:
  test_numpy:
    args:
      path: /path/to/windows/samba/share//numpy.npy
    description: Test numpy
    driver: intake.source.npy.NPySource
    metadata:
      catalog_dir: /path/to/windows/samba/share/
```
But it crashes on `cat.test_numpy.read()`:
```
FileNotFoundError: [Errno 2] No such file or directory: '[c:/path/to/windows/samba/sahre/numpy.npy](file:///C:/path/to/windows/samba/share/numpy.npy)'
```
As far as I understand, it replaces the first slash in the path `//path/to/windows/samba/share/numpy.npy` with `C:`, and therefore cannot access the file.

OS: Windows 10
intake: 0.6.8 and commit e362825 from git
intake-xarray: 0.4.1
fsspec: 2023.5.0
python: 3.10.11

kadykov:
Fixes #739, by the solution proposed by @martindurant


martindurant:
```
    def test_get_dir():
>       assert get_dir("file:///path/catalog.yml") == "file:///path"
E       AssertionError: assert 'file:///path/' == 'file:///path'
E         - file:///path
E         + file:///path/
E         ?             +
```

I would change the test in this case - we expect dir paths to terminate with "/" in intake usage (may revisit this later); there are a couple of instances in this test that need correcting. 
Previously, we had different behaviour depending on whether a local path started with file:// or not, which is problematic.

kadykov:
The main problem is not in stripping of the last `/`, but that the private method `_parent()` of the `ffsspec.implementations.local.LocalFileSystem` does not handle `file://` paths correctly.

```python
from intake.catalog.local import split_protocol, get_filesystem_class

path = "file:///path/catalog.yml"

protocol, _ = split_protocol(path)
out = get_filesystem_class(protocol)._parent(path)
print(out)
```
This code above on Windows returns `c:/path` and it breaks samba share access.
I don't think that using private methods from 3rd party packages is a good idea. Can we replace it with something else like this?
```python
from pathlib import PurePosixPath

str(PurePosixPath(path).parent)
```

martindurant:
Maybe that should be 
```
fs = get_filesystem_class(protocol)
out = fs._parent(fs._strip_protocol(path))
```

There is no problem calling "_" methods, the projects are closely tied. It just means not "end user" method. You could propose at fsspec that methods be elevated to public, maybe.

kadykov:
I think that we have to decrease coupling between these projects to simplify further development, so I implemented the same functionality using `pathlib.PurePath`.
I also modified the tests for `get_dir()` so that it expects `/` in the end of a directory, and threats paths like `example/` as relative.

intake/catalog/tests/test_local.py
```diff
@@ -121,19 +121,19 @@ def test_use_source_plugin_from_config(catalog1):
 
 
 def test_get_dir():
-    assert get_dir("file:///path/catalog.yml") == "file:///path"
-    assert get_dir("https://example.com/catalog.yml") == "https://example.com"
+    assert get_dir("file:///path/catalog.yml") == "file:/path/"
```
martindurant, line 124: The loss of "//" is problematic here

kadykov: I reverted the previous version with keeping `://`

intake/catalog/local.py
```diff
@@ -503,17 +503,8 @@ def register_plugin_module(mod):
...
-        if "://" not in out:
-            # some FSs strip this, some do not
-            out = protocol + "://" + out
-        return out
-    path = make_path_posix(os.path.join(os.getcwd(), os.path.dirname(path)))
-    if path[-1] != "/":
-        path += "/"
-    return path
+    """Get the parent directory of the path"""
+    return f"{PurePath(path).parent.as_posix().replace(':/', '://')}/"
```
martindurant, line 507: This now fails for Windows paths

kadykov: Sorry, I don't get it. On my Windows laptop, all the tests from `test_local.py` are completed successfully except for `def test_default_expansions()`, but this test was failing before.

If you are worrying about converting the path to POSIX, the `get_filesystem_class(protocol)._parent(path)` does it also by calling `make_path_posix(path)`:
```python
    @classmethod
    def _parent(cls, path):
        path = cls._strip_protocol(path).rstrip("/")
        if "/" in path:
            return path.rsplit("/", 1)[0]
        else:
            return cls.root_marker

    @classmethod
    def _strip_protocol(cls, path):
        path = stringify_path(path)
        if path.startswith("file://"):
            path = path[7:]
        elif path.startswith("file:"):
            path = path[5:]
        return make_path_posix(path).rstrip("/") or cls.root_marker
```

intake/catalog/tests/test_local.py
```diff
@@ -121,19 +121,19 @@ def test_use_source_plugin_from_config(catalog1):
 
 
 def test_get_dir():
-    assert get_dir("file:///path/catalog.yml") == "file:///path"
-    assert get_dir("https://example.com/catalog.yml") == "https://example.com"
+    assert get_dir("file:///path/catalog.yml") == "file://path/"
```
martindurant, line 124: I am still confused by `.replace(':/', '://')`, and its apparent inverse which happened here.

Does as_posix() silently change "file://"->"file:/" ? What then happens to paths that genuinely have colons, like "x:/path/..."?

PR #748, opened 2023-05-31
Gui monolith

martindurant:


martindurant:
@blakerosenthal , @AlbertDeFusco : you might be interested in playing around with this, the far simpler rewrite of the GUI code. For the moment it still requires `pn.extension("codeeditor")` explicitly in a notebook; I'm no sure where to go with that, since Intake cannot know what other extensions may be required later.

martindurant:
Screen Shot 2023-06-02 at 15 53 57

martindurant:
^ in classic notebook

Issue #750, opened 2023-06-08
BUG: Using wildcards in open_csv in intake 0.7.0 when reading from S3 causes failure


arvindh-manian:
```python
import intake
intake.open_csv("s3://test-for-intake/intake_test.csv").read()
```

This gives the expected output 
```
   col1  col2
0  val1  val2
```

But using a wildcard
```python
import intake
intake.open_csv("s3://test-for-intake/*.csv").read()
```

Causes this error:
```
Traceback (most recent call last):
File "<stdin>", line 1, in <module>
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/intake/source/csv.py", line 190, in read
    self._get_schema()
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/intake/source/csv.py", line 142, in _get_schema
    df = self._get_partition(0)
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/intake/source/csv.py", line 160, in _get_partition
    return self._read_pandas(url_part, i)
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/intake/source/csv.py", line 166, in _read_pandas
    return pd.read_csv(url_part, storage_options=self._storage_options, **self._csv_kwargs)
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 912, in read_csv
    return _read(filepath_or_buffer, kwds)
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 577, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 1407, in __init__
    self._engine = self._make_engine(f, self.engine)
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 1661, in _make_engine
    self.handles = get_handle(
  File "/Users/arvindh/.pyenv/versions/3.9.7/lib/python3.9/site-packages/pandas/io/common.py", line 859, in get_handle
    handle = open(
FileNotFoundError: [Errno 2] No such file or directory: 'test-for-intake/intake_test.csv'
```

[...]

PR #751, opened 2023-06-08
fix fsspec protocol in CSV

blakerosenthal:
Fixes https://github.com/intake/intake/issues/750

In splitting out `_files` in https://github.com/intake/intake/pull/734 I inadvertently broke non-local CSV sources. Turns out `fsspec.core.get_fs_token_paths` does not preserve protocol prefixes (presumably because it also returns a FileSystem obj). Intake doesn't have any s3 tests either, so it might be worth getting some mocks to work.

@martindurant This could possibly also be an issue [in the refactor](https://github.com/intake/intake/blob/c3a6f47a802af3db655b49d98aadb9319fb589ac/intake/readers/datatypes.py)

### DIFF
--- a/intake/source/csv.py
+++ b/intake/source/csv.py
@@ -5,7 +5,7 @@
 # The full license is in the LICENSE file, distributed with this software.
 # -----------------------------------------------------------------------------
 
-from fsspec.core import get_fs_token_paths
+from fsspec.core import get_fs_token_paths, split_protocol
 
 from . import base
 from .utils import reverse_formats, unique_string
@@ -123,7 +123,10 @@ class CSVSource(base.DataSource, base.PatternMixin):
         if self.pattern is None and not glob_in_path:
             self._files = urlpath
         else:
-            self._files = get_fs_token_paths(urlpath)[2]
+            protocol = split_protocol(urlpath[0])[0]
+            self._files = get_fs_token_paths(urlpath, storage_options=self._storage_options)[2]
+            if protocol:
+                self._files = [f"{protocol}://{fn}" for fn in self._files]
 
         return sorted(self._files)
 
--- a/intake/source/csv.py
+++ b/intake/source/csv.py
@@ -124,11 +124,12 @@ class CSVSource(base.DataSource, base.PatternMixin):
             self._files = urlpath
         else:
             protocol = split_protocol(urlpath[0])[0]
-            self._files = get_fs_token_paths(urlpath, storage_options=self._storage_options)[2]
+            fs, _, paths = get_fs_token_paths(urlpath, storage_options=self._storage_options)
+            self._files = sorted(paths)
             if protocol:
-                self._files = [f"{protocol}://{fn}" for fn in self._files]
+                self._files = [fs.unstrip_protocol(fn) for fn in self._files]
 
-        return sorted(self._files)
+        return self._files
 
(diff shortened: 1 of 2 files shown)


blakerosenthal:
The mamba GH action seems to be running into this issue
https://github.com/mamba-org/mamba/issues/2363

Not sure if it's worth temporarily switching to miniconda, or to just wait for the fix

blakerosenthal:
[This commit](https://github.com/intake/intake/pull/751/commits/3f5189995f9824b3094d2422fc62e9c604d8a092) is still relevant because the old Action was giving deprecation warnings.

intake/source/csv.py
```diff
@@ -123,7 +123,10 @@ def files(self):
         if self.pattern is None and not glob_in_path:
             self._files = urlpath
         else:
-            self._files = get_fs_token_paths(urlpath)[2]
+            protocol = split_protocol(urlpath[0])[0]
+            self._files = get_fs_token_paths(urlpath, storage_options=self._storage_options)[2]
+            if protocol:
+                self._files = [f"{protocol}://{fn}" for fn in self._files]
```
martindurant, line 129: Unfortunately, not all implementations strip the protocol the same way, so you'll need `fs.unstrip_protocol`.

Other options would be:
- add `protocol=` to the storage_options, if it's not already there
- to store the fs we infer
- to use fs.open_files and store those OpenFiles
I don't immediately know how much further knock-on changes either of those would cause.

martindurant:
> The old Action was giving deprecation warnings.

I've seen this and need to fix in a number of places, although I don't see a suggestion that the older action is actually problematic (yet)

intake/source/csv.py
martindurant: I think I prefer bullet 1.

blakerosenthal: Bullet 1 definitely looks best to me too. Unfortunately this wasn't working for me:

```python
In [4]: from fsspec.core import get_fs_token_paths

In [5]: url = "s3://test-for-intake/*.csv"

In [6]: so = {"s3": {"anon": True}}

In [7]: get_fs_token_paths(url, storage_options=so)
Out[7]:
(<s3fs.core.S3FileSystem at 0x151378f40>,
 '081ddbc04ab2fb22c401dcebc75ec0bc',
 ['test-for-intake/intake_test.csv'])

In [8]: so["protocol"] = "s3"

In [9]: so
Out[9]: {'s3': {'anon': True}, 'protocol': 's3'}

In [10]: get_fs_token_paths(url, storage_options=so)
Out[10]:
(<s3fs.core.S3FileSystem at 0x151378f40>,
 '081ddbc04ab2fb22c401dcebc75ec0bc',
 ['test-for-intake/intake_test.csv'])

In [11]: get_fs_token_paths(url, storage_options=so, protocol="s3")
Out[11]:
(<s3fs.core.S3FileSystem at 0x151378f40>,
 '081ddbc04ab2fb22c401dcebc75ec0bc',
 ['test-for-intake/intake_test.csv'])
```

martindurant: Ah right. So those paths are correct, but pandas won't even ask fsspec to handle them unless there's a protocol:// in place :|
Back to unstrip, then.

blakerosenthal:
Looks like mamba is happy again

Issue #755, opened 2023-08-30
Reading CSV sources that require `skipfooter` crashes due to `nrows` being set in CSVSource._get_schema


korsbakken:
In line 145 of intake/source/csv.py, in the method `CSVSource._get_schema`, the `nrows` parameter to `pandas.read_csv` gets set to 10, presumably to read only 10 rows to infer the schema in constant time, as per the driver guidelines. However, `nrows` is incompatible with the `skipfooter` parameter, and this results in a crash when calling `.read()` on a data source that sets `skipfooter` in `csv_args`, e.g., to read a CSV file that breaks the strict column format a CSV file is supposed to have by putting metadata or comments as footnotes at the bottom of the file (unfortunately a very common practice).

The key line is here:
https://github.com/intake/intake/blob/81b1567a2030adfb22b856b4f63cefe35de68983/intake/source/csv.py#L145

To reproduce, create a catalog with a csv source entry that has `skipfooter` under `args/csv_args`, e.g.:
```
my_csv_source:
  driver: csv
  args:
    urlpath: "{{CATALOG_DIR}}/csv_file_with_footnotes.csv"
    csv_kwargs:
      skipfooter: 2
```

When calling `.read()` on the resulting `CSVSource` object, this will result in pandas throwing `ValueError: 'skipfooter' not supported with 'nrows'` when `pandas.read_csv` gets called during the call to `CSVSource._get_schema`.

I'm not sure what the best way is to get around this. Maybe have `CSVSource._get_schema` check whether `skipfooter` is contained in `self._csv_kwargs`, temporarily back it up and unset it if it is, and then put it back before `._get_schema` returns? Maybe also take some precautions to avoid the footers messing up the schema inference, such as setting `nrows` to `10+skipfooter` and then remove `skipfooter` rows from the bottom of the DataFrame that gets returned by `pandas.read_csv`?

martindurant:
Quick question: do you ever actually check the schema before reading the data?

Side question: do you ever read this data with dask, or only pandas?

korsbakken:
The error occurs when I call `.read()` on the data source object. The same error happens if I call `.discover()` on the same object. Is that what you meant by "check the schema"? As stated previoiusly, the error is generated in `CSVSource._get_schema`.

I only read the data with pandas. It's way too small for dask to be useful.

martindurant:
> Is that what you meant by "check the schema"?

[...]

PR #756, opened 2023-08-30
Fix crash in CSVSource._get_schema when `skipfooter` is set in csv_kwargs

korsbakken:
Fixes issue #755


intake/source/csv.py
```diff
@@ -142,12 +142,23 @@ def _get_schema(self):
             return self._schema
 
         nrows = self._csv_kwargs.get("nrows")
-        self._csv_kwargs["nrows"] = 10
+        skipfooter = self._csv_kwargs.get("skipfooter")
```
martindurant, line 145: How about
```
nrows = self._csv_kwargs.pop("nrows")
skipfooter = self._csv_kwargs.pop("skipfooter")
df = self._get_partition(0)
if nrows:
    self._csv_kwargs["nrows"] = nrows
if skipfooter:
    self._csv_kwargs["skipfooter"] = skipfooter
```

korsbakken: That wouldn't work in this case, on two counts:

1. `dict.pop(key)` fails and reaises a `KeyError` if `key` is not present in the dict. We don't know that `self._csv_kwargs` contains either `"nrows"` or `"skipfooter"` to begin with. To avoid a possible crash, we have to either use `dict.get` as in the PR code, or test first with, e.g., `if "skipfooter" in selv._csv_kwargs` first, which seems more awkward. Is there any particular reason you want to use `dict.pop` here?

2. If `"skipfooter"` is in `self._csv_kwargs`, I think we should add that number of rows to the 10 rows to be read, like I have done in line 148 of the PR code, and then remove the same number of rows from the bottom of the returned DataFrame, like I do in line 160. If `skipfooter` is set, it means that there are rows at the bottom of the CSV file that should not be used for inferring the schema. We should therefore remove any rows that might be part of the footer, which could happen if the CSV is small enough that the footers start less than 10 rows from the top. But we probably also want to make sure that we get 10 rows to infer the schema from, if the size of the CSV allows for that. We achieve that if we read `10+skipfooter` rows, and then remove the bottom `skipfooter` number of rows afterwards. It means a marginal efficiency loss, but should be negligible if `skipfooter` is not ridiculously high.

martindurant:
You convinced me that your code is fine. Would you mind making a test including skipfooter= ?

korsbakken:
Hi, and sorry for the long response times, a lot going on here right now.

Yes I can certainly add a test for the skipfooter functionality. What would be the best way of doing so? I'm thinking the following, but let me know if there's way that fits better with the existing test regime:

1. Add a sample CSV file with footer rows in `intake/source/tests/`.
2. Add the CSV file name to the dict returned by `test_csv.data_filenames()`
3. Add a pytest fixture that gets the new file name from `data_filenames()`, and returns `CSVSource(filename, csv_kwargs={'skipfooter': n})`, where  `n` is the number of footer rows.
4. Add a test function in `test_csv.py` that uses the new `CSVSource` fixture, and asserts that the `.discover()` method returns the expected schema (and doesn't crash), and that the `.read()` method returns the expected data, without the footer rows.

martindurant:
Honestly any of those would be perfectly fine with me, whichever you prefer.

korsbakken:
Sorry if I wasn't clear, I meant that as a list of steps, not options. I think we need all of them for a functioning test, or at least we need some series of steps that makes a CSV file with footers available, and then step 4. But anyway, I will get cracking later today.

korsbakken:
I have now added test CSVs in intake/source/tests/, and test code in intake/source/tests/test_csv.py. The new tests both pass. They test that both the content and the inferred schema with footers are equal to what you get from an equivalent CSV file without footers.

I also had to tweak the code in `intake.source.csv._get_schema()` from the original PR. It turned out that the implementation there still included the footers in the inferred dtypes if the CSV file was less than `10+skipfooters` rows. The modified code avoids that issue. To do so, it has to read in the CSV file an extra time if it detects that the DataFrame has less than `10+skipfooter` rows. That might mean a slight performance hit, but should not be noticeable given that it only happens for very small CSV files.

I also changed to using `csv_kwargs.pop` instead of first `.get` and then `del`, btw. I liked that pattern better myself after looking at it.

martindurant:
I am not certain, but I suspect that the test failures are due to the extra files in the test csv dictionary.

korsbakken:
Looks like there are a bunch of test failures related to the filenames I added in `test_csv.data_filenames`. I'm not sure I quite understand how the filenames there are parsed. It works fine for the test function `test_skipfooter` that I added, but other tests apparently use `data_filenames` in a way that fails with the additions that I made.

Could you enlighten me on how I can add new files to `test_csv.data_filenames` in a way that doesn't break any existing tests?

martindurant:
> The only partial exception is /catalog/tests/test_remote_integration.py, which for some reason hangs on my computer and had to be skipped, don't know why.

Seems ok here, so I'm not worried about it.

PR #757, opened 2023-09-05
Precise 'int' type in 'zarr' test

kadykov:
This PR fixes falling 'zarr' test

### DIFF
--- a/intake/source/tests/test_npy.py
+++ b/intake/source/tests/test_npy.py
@@ -83,7 +83,7 @@ def test_zarr_minimal():
     assert s.container == "ndarray"
     assert s.read().tolist() == [73, 98, 46, 38, 20, 12, 31, 8, 89, 72]
     assert s.npartitions == 1
-    assert s.dtype == "int"
+    assert s.dtype == np.int64
     assert s.shape == (10,)
     assert (s.read_partition((0,)) == s.read()).all()
 


martindurant:
This is due to a new version of numpy?
```
assert np.array([1]).dtype == "int"
```

I am fine to merge this, but it confuses me!

kadykov:
I have this output for this test:
I don't know. I agree, it is a bit weird.
```python
intake/source/tests/test_npy.py::test_zarr_minimal failed: def test_zarr_minimal():
        pytest.importorskip("zarr")
        cat = intake.open_catalog(posixpath.join(here, "sources.yaml"))
        s = cat.zarr1()
        assert s.container == "ndarray"
        assert s.read().tolist() == [73, 98, 46, 38, 20, 12, 31, 8, 89, 72]
        assert s.npartitions == 1
>       assert s.dtype == "int"
E       AssertionError: assert dtype('int64') == 'int'
E        +  where dtype('int64') = sources:\n  zarr1:\n    args:\n      urlpath: c:/Users/kadykov/Documents/prog/git_repos/intake/intake/source/tests//data....ZarrArraySource\n    metadata:\n      catalog_dir: c:/Users/kadykov/Documents/prog/git_repos/intake/intake/source/tests/\n.dtype

intake\source\tests\test_npy.py:86: AssertionError
```
OS Windows 10
python 3.8.17
numpy 1.24.4
pytest 7.4.1

martindurant:
Perhaps "int64" would have worked, or 's.dtype.kind == "i"`... Anyway, this is fine now.

kadykov:
Yes, both of these options also work

PR #761, opened 2023-09-05
Add CI tests on Windows

### LINKED ISSUE #760: Add tests on Windows to CI github actions
Probably, it would be better to create another job in the main CI GitHub action that runs the same tests on `windows-latest`

kadykov:
The aim of this PR is to improve automatic test coverage of the CI actions by launching the same tests on Linux and Windows

Fixes #760


martindurant:
This requires #759 to pass?

kadykov:
Yes, but I'm afraid it is not sufficient.
In any case, I think it is better to launch the tests on both platforms.

martindurant:
I only see a couple of path-related failures, and I presume your other PR fixes at least one of them

martindurant:
OK, just the one failure. Will you have the chance to track this down?

kadykov:
Yes, I have to check once agian #743

kadykov:
As far as I understand, using `postgresql` fixture from `pytest-postgresql` [is not fully supported on Windows](https://github.com/ClearcodeHQ/pytest-postgresql/issues/303), so I skipped these tests on Windows.

martindurant:
Seems like we should also posixiify the `sqlite_with_data` path in the duckDB test. Not sure if the reader should do that.

kadykov:
I agree, that we should not posixify the path in the DuckDB reader.

The problem was not with the posixification, but with the protocol extraction, that we also should not do there.

intake/readers/readers.py
```diff
@@ -360,7 +360,7 @@ def _duck(self, data):
             # https://duckdb.org/docs/extensions/
             if conn.startswith("sqlite:"):
                 duckdb.connect(":default:").execute("INSTALL sqlite;LOAD sqlite;")
-                conn = re.sub("^sqlite3?://", "", conn)
+                conn = re.sub("^sqlite3?(:///|://|:/)", "", conn)
```
martindurant, line 363:
```suggestion
                conn = re.sub("^sqlite3?:[/]+", "", conn)
```
perhaps?

.github/workflows/main.yaml
```diff
@@ -32,3 +32,29 @@ jobs:
...
+
+      - name: pip-install
+        shell: bash -l {0}
+        run: |
+          pip install . --no-deps
+
+      - name: Run Tests
+        shell: bash -l {0}
+        run: |
+          pytest -v intake/readers -k "not skip_on_windows"
```
martindurant, line 60: Why not actually make them skip with a skipif?

martindurant (approved):
A couple of small suggestions.

intake/readers/readers.py
kadykov: Accordin to the [URI syntax](https://datatracker.ietf.org/doc/html/rfc3986#section-3), only the following combinations are possible:

1. `sqlite://file:/path/to/database.db` with an authority component `file:`.
2. `sqlite:///path/to/database.db` with an empty authority component.
3. `sqlite:/absolute/path/to/database.db` without an authority component, the `/` after the `:` symbol points to the root of the file system.
4. `sqlite:relative/path/to/database.db` without an authority component, the absence of the `/` after the `:` symbol means that the path is relative to the current folder.

From [the documentation](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3):

> If a URI contains an authority component, then the path component must either be empty or begin with a slash ("/") character.

Therefore, a path with more than three slashes after the `:` symbol is not allowed.
Please correct me if I'm wrong.

I would prefer a code that breaks [as soon as possible](https://en.wikipedia.org/wiki/Fail-fast_system), instead of a code that silently corrects an invalid input and makes the debugging harder.

kadykov: Nice suggestion! It seems more readable to me.

.github/workflows/main.yaml
kadykov: Done!

kadykov:
I combined Windows and Linux test jobs with a `matrix`.
Now this PR is ready to be merged.

PR #763, opened 2023-09-06
Fix get_dir() behavior on Windows

kadykov:
This PR fixes `get_dir()` behavior on the Windows platform.

It is supposed to replace #743, and should be compatible with #761.

Function `get_dir()` is rewritten according to the tests in `test_local.test_get_dir()`, but the expected behavior is quite weird.
It looks like the real problem is somewhere else, but we are trying to patch it here.

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -502,18 +502,15 @@ def register_plugin_module(mod):
             register_driver(k, v, clobber=True)
 
 
-def get_dir(path):
-    if "://" in path:
-        protocol, _ = split_protocol(path)
-        out = get_filesystem_class(protocol)._parent(path)
-        if "://" not in out:
-            # some FSs strip this, some do not
-            out = protocol + "://" + out
-        return out
-    path = make_path_posix(os.path.join(os.getcwd(), os.path.dirname(path)))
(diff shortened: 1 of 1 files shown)


intake/catalog/local.py
```diff
@@ -502,18 +502,15 @@ def register_plugin_module(mod):
...
-    return path
+def get_dir(path: str) -> str:
+    """Get the absolute path of the directory of the file"""
+    protocol, pure_path = split_protocol(path)
+    dirname = os.path.dirname(pure_path)
+    if protocol:
+        # leave the path without '/' in the end
+        return f"{protocol}://{dirname}"
+    # get absolute path, convert to posix and add '/' in the end
+    return f"{os.path.abspath(dirname).replace(os.sep, r'/')}/"
```
martindurant, line 513: Will this not produce the OS-dependent version of the path?
I wonder whether `fsspec.core.url_to_fs` might be a more appropriate function for this job.

kadykov: I think it would be quite challenging due to [its strange behavior](https://github.com/fsspec/filesystem_spec/discussions/1356).
I have prepared another solution #768

Issue #764, opened 2023-09-06
Fix inconsistent `urlpath` links to directories in tests and docs


kadykov:
In the documentation and in test files you can find different ways to link to the subdirectory.
Like this:
```yaml
      urlpath: '{{ CATALOG_DIR }}/cache_data/states.csv'
```
and like that:
```yaml
      urlpath: '{{ CATALOG_DIR }}cache_data/states.csv'
```
What do you think if we keep only one way, for instance the first one with implicit `/`?
Probably, it will be safer and easier to maintain...

martindurant:
Yes, I think the first of these feels more right.

kadykov:
Fixed by #765

PR #765, opened 2023-09-06
Ensure "/" between CATALOG_DIR and file or subfolder

kadykov:
This PR makes links to files and subfolders in catalog files consistent, according to #764


kadykov:
@martindurant Do you have any comments on this pull request?

martindurant:
Yes, I think this is the right way around.

PR #766, opened 2023-09-07
Allow any signed integer

kadykov:
This PR is a minor correction to PR #757


martindurant:
I would have been fine either way ;)

PR #772, opened 2023-10-09
Fix get_dir() using fsspec

kadykov:
This is another solution for #739, and it works better than #768.
However, it breaks catalogs created in memory, like in `test_fsspec_integration`

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -11,7 +11,7 @@ import os
 import warnings
 
 import entrypoints
-from fsspec import get_filesystem_class, open_files
+from fsspec import open_files
 from fsspec.core import split_protocol
 
 from .. import __version__
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -502,18 +502,11 @@ def register_plugin_module(mod):
             register_driver(k, v, clobber=True)
 
 
-def get_dir(path):
-    if "://" in path:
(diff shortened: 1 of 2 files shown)


martindurant:
> it breaks catalogs created in memory

Can you be more specific on what breaks?

martindurant:
> Is it the expected behaviour of _parent()?

Yes, I think it is: fsspec filesystems return file strings within their own understanding of the storage backend, so normally without the protocol attached. `fs.unstrip_protocol` exists to make the full FS-independent path. 

The following code in `fsspec.generic.GenericFileSystem` might be a good way to provide a concrete space for this kind of thing.
```diff
--- a/fsspec/generic.py
+++ b/fsspec/generic.py
@@ -171,6 +171,10 @@ class GenericFileSystem(AsyncFileSystem):
         self.method = default_method
         super().__init__(**kwargs)

+    def _parent(self, path):
+        fs = _resolve_fs(path, self.method)
+        return fs.unstrip_protocol(fs._parent(path))
+
```

martindurant:
Or perhaps like this? I feel like we've written this a few times and now come to anything satisfactory:
```
def abs_parent(path):
    fs, _ = fsspec.core.url_to_fs(path)
    return fs.unstrip_protocol(fs._parent(path))
```

kadykov:
It seems that the solution with `UPath` #773 works better, therefore it is preferable to use it instead.

PR #773, opened 2023-10-25
Fix get_dir() with UPath

kadykov:
This PR is another try to solve #739

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -11,7 +11,7 @@ import os
 import warnings
 
 import entrypoints
-from fsspec import get_filesystem_class, open_files
+from fsspec import open_files
 from fsspec.core import split_protocol
 
 from .. import __version__
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -502,18 +502,11 @@ def register_plugin_module(mod):
             register_driver(k, v, clobber=True)
 
 
-def get_dir(path):
-    if "://" in path:
-        protocol, _ = split_protocol(path)
-        out = get_filesystem_class(protocol)._parent(path)
-        if "://" not in out:
-            # some FSs strip this, some do not
-            out = protocol + "://" + out
-        return out
-    path = make_path_posix(os.path.join(os.getcwd(), os.path.dirname(path)))
-    if path[-1] != "/":
-        path += "/"
-    return path
+def get_dir(path: str) -> str:
+    """Get the directory path of the file"""
+    protocol, pure_path = split_protocol(path)
+    dirname = os.path.dirname(pure_path)
+    return f"{protocol}://{dirname}" if protocol else dirname
 
 
 class YAMLFileCatalog(Catalog):
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -12,7 +12,7 @@ import warnings
 
 import entrypoints
 from fsspec import open_files
-from fsspec.core import split_protocol, url_to_fs
+from upath import UPath
 
 from .. import __version__
 from ..source import get_plugin_class, register_driver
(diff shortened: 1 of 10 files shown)


kadykov:
As suggested in [universal_pathlib](https://github.com/fsspec/universal_pathlib/issues/161) thread, it seems that it is better to avoid `memory://` "double-slash" notation.
@martindurant What do you think about this solution?

martindurant:
>  it is better to avoid memory:// "double-slash" notation

Absolutely not, this is authoritative notation for all kinds of URLs in all kinds of places _except_ local files (on windows).

kadykov:
> > it is better to avoid memory:// "double-slash" notation
> 
> Absolutely not, this is authoritative notation for all kinds of URLs in all kinds of places _except_ local files (on windows).

The path like `memory:/file.csv` and `memory:///file.csv` work well, but `memory://file.csv` is [not yet](https://github.com/fsspec/universal_pathlib/pull/152) compatible with `UPath`.

@martindurant Just to avoid any misunderstanding, I propose to avoid using "double-slash" notation only for links to locations in *memory*. For the other storage types, we will continue to use the existing notations.

martindurant:
> avoid using "double-slash" notation only for links to locations in memory.

It seems dangerous to special-case this - other URLs tool very similar! Every backend has its peculiarities (e.g., it's unlikely in S3 to want the parent of a bucket; and in HTTP, you usually don't have hierarchical parents at all; for local, posix is very different from windows).

kadykov:
> > avoid using "double-slash" notation only for links to locations in memory.
> 
> It seems dangerous to special-case this - other URLs tool very similar! Every backend has its peculiarities (e.g., it's unlikely in S3 to want the parent of a bucket; and in HTTP, you usually don't have hierarchical parents at all; for local, posix is very different from windows).

@martindurant Sorry, I'm not sure that I follow what you are trying to say.

The solution with `UPath` seems to work quite well:

```python
def get_dir(path: str) -> str:
    return str(UPath(path).parent)
```

The only exception is when we try to reconstruct `urlpath` from the `{{CATALOG_DIR}}/file.ext` entry in a catalog located in memory:

```python
import intake
import fsspec

mem = fsspec.filesystem("memory")
with mem.open("cat_memory.yaml", "wt") as f:
    f.write(
            """
sources:
  implicit:
    driver: csv
    description: o
    args:
      urlpath: "{{CATALOG_DIR}}/file_memory.csv"
"""
    )
with mem.open("file_memory.csv", "wt") as f:
    f.write("a,b\n0,1")
cat = intake.open_catalog("memory://cat_memory.yaml")  # double-slash notation
cat.implicit.read()  # Does not work

cat = intake.open_catalog("memory:///cat_memory.yaml")  # empty authority URI
cat.implicit.read()  # Ok
```

Therefore, in this PR, I propose to temporary modify the `test_fsspec_integration()` test to use `memory:///cat.yml` instead of `memory://cat.yml`, that's it. After [this issue](https://github.com/fsspec/universal_pathlib/issues/161) [is fixed](https://github.com/fsspec/universal_pathlib/pull/152) we could extend the test with `memory://cat.yml`.

No changes to S3, HTTP, file or anything else.

It seems that it is the final piece in the puzzle of the Windows support, and we will be able to fix #739, then merge #761 and fix #760.

ap--:
I would currently not recommend using `UPath` to add support for retrieving the parent path for arbitrary fsspec filesystems.

The current implementation relies a lot on cpython's pathlib way of splitting path parts and `universal_pathlib` can't give guarantees that `UPath(...).parent` will do the correct thing for fsspec filesystems that are not tested in the universal_pathlib test suite.

This will change in future versions, when this will rely more on the individual AbstractFilesystem subclasses, and will most likely be available once cpython's `pathlib.BasePath` classes for implementing virtual filesystems will become available (probably python==3.13)

kadykov:
> I would currently not recommend using `UPath` to add support for retrieving the parent path for arbitrary fsspec filesystems.

@ap-- Unfortunately, we don't have any better solution so far.

I have checked the [UPath patch](https://github.com/fsspec/universal_pathlib/pull/162) and it works well with `intake` and `get_dir()`:
```python
def get_dir(path):
    return str(UPath(path).parent)
```

When it will be in upstream, the modifications of the memory paths like `memory://` in tests will not be necessary.

PR #779, opened 2024-02-05
Remove cache and test for py38

Hoxbro:
Currently, intake 2.0 supports Python 3.8, and `functools.cache` was first added in Python 3.9. 

https://github.com/intake/intake/blob/6426267e4f01ed55b79c15ad3e514b2176442720/pyproject.toml#L25

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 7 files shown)


martindurant:
Thank you for putting in the effort here. The obvious alternative would have been to drop 3.8, as you can see (some elements of) pandas already has. As it is, we probably have to skip the failing test.

martindurant:
> the already released version

Fair enough, we can drop in the next one after

Issue #781, opened 2024-02-05
Add missing intake/_version.py


kadykov:
As it is required in 
https://github.com/intake/intake/blob/6426267e4f01ed55b79c15ad3e514b2176442720/intake/__init__.py#L10

martindurant:
Isn't this always made by `pip`? Or are you saying there should be a fallback file for the case that the package is being imported directly from the repo?

kadykov:
Yes, `pip` + `setuptools_scm` create the file, but the idea is to be able to use the package directly from the repository without installation.
Solved by #782, so we can close it.

PR #786, opened 2024-02-08
Add migration guide

martindurant:
For devs who have written v1 plugins. Still fairly terse.


intake/_version.py
djhoese, line 1: The top comment on this file says not to commit it :wink:

martindurant: It is in .giignore, and I also ran git update-index --assume-unchanged , but some base version of the file needs to be there

djhoese: I've used setuptools_scm for quite a while and unless something has changed, this file should never need to be in revision controlled. The only requirement is that users `pip install .` or `pip install -e .` before trying to import or use the package.

martindurant: So I was misled in https://github.com/intake/intake/issues/781 ?

djhoese: It depends what you think about that last comment. If you want the package to be importable without having it installed then you need some fallback (more on that later). I think in the modern python world of Cython extensions and entry points this isn't a reasonable expectation, even though this package doesn't (from what I can tell) have any of these.

As for a fallback, I think it would be better to have a `try/except ImportError` in `__init__` with a fallback version of something like "0.0.0" rather than adding a file to revision control that says not to add it to revision control. Especially since this file gets modified every time a contributor installs the package. That means that every PR has the potential for an (unnecessary) update to this file.

CC @kadykov

PR #791, opened 2024-02-27
readd entrypoints for old style sources in this package

martindurant:
Fixes https://github.com/intake/intake-xarray/issues/140

@forestbat , if you wouldn't mind testing this. I am unsure whether I want to pollute the top-level namespace with these deprecated open* names. Do you think it would be reasonable to hide them behind a config option, or require calling a function to see them?

### DIFF
--- a/intake/catalog/local.py
+++ b/intake/catalog/local.py
@@ -394,6 +394,10 @@ class CatalogParser(object):
                 )
                 continue
 
+            elif "module" in plugin_source:
+                import intake
+
+                intake.import_name(plugin_source["module"])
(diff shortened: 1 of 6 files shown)


forestbat:
> Fixes [intake/intake-xarray#140](https://github.com/intake/intake-xarray/issues/140)
> 
> @forestbat , if you wouldn't mind testing this. I am unsure whether I want to pollute the top-level namespace with these deprecated open* names. Do you think it would be reasonable to hide them behind a config option, or require calling a function to see them?

I think the later is better, because function can be prompted by IDE, however config file can't.

Issue #802, opened 2024-04-29
`reader_from_call` needs fixing. 


pt247:
The simple example in [docs](https://intake.readthedocs.io/en/latest/walkthrough2.html#simple-example) is broken.

Steps to reproduce:
1. Following gives no errors
```
import pandas as pd
df = pd.read_csv(url)
```

but this gives error:
```
In [25]: reader = intake.reader_from_call('df = pd.read_csv(url)')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[25], line 1
----> 1 reader = intake.reader_from_call('df = pd.read_csv(url)')

File ~/work/mcg_prototype/.venv/lib/python3.12/site-packages/intake/readers/readers.py:1359, in reader_from_call(func, join_lines, *args, **kwargs)
   1357     func = func.replace("\n", "")
   1358 frame = inspect.currentframe().f_back
-> 1359 match = re.match("^(.*=)?(.*?)[(](.*)[)]")
   1360 if match:
   1361     groups = match.groups()

TypeError: match() missing 1 required positional argument: 'string'

In [26]: 

```

Issue: 

https://github.com/intake/intake/commit/9cfa7a82feda20816672994adcc1bb92f628adf7#r141464783

kthyng:
Ah, I was about to say this! Thanks.

PR #803, opened 2024-04-29
fix re.match function call in reader_from_call

pt247:
fix re.match function call in reader_from_call

fixes: #802


martindurant:
Of course, the real problem is, that we have no test (so slips like this can get into the code). Would you mind turning what you had in your issue into a test?

pt247:
Added a test. Is that enough for this PR?

martindurant:
Perfect, thank you.

PR #804, opened 2024-05-06
Update class names in pandas/xarray converters

nshankar:
I noticed a couple typos leading to errors like
```
AttributeError: module 'xarray' has no attribute 'DataSet'. Did you mean: 'Dataset'?
```

Simple script to recreate the error:
```
import intake
import xarray as xr 

url = "gs://weatherbench2/datasets/era5/1959-2022-6h-1440x721.zarr"
# xr.open_dataset(url, engine="zarr", consolidated=True)

reader = intake.reader_from_call(
    'xr.open_dataset(url, engine="zarr", consolidated=True)'
)
```

### DIFF
--- a/intake/readers/convert.py
+++ b/intake/readers/convert.py
@@ -108,12 +108,12 @@ class PandasToGeopandas(BaseConverter):
 
 
 class XarrayToPandas(BaseConverter):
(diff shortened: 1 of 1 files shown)


martindurant:
Thank you. I wonder if there's a convenient way to test the existence of all of the string-defined classes in with the consistency checks.

Issue #809, opened 2024-05-23
A path starting with "http" from xpublish interpreted incorrectly


kthyng:
Hi! I am trying to update the intake catalog presentation of the xpublish server from [xpublish-intake](https://github.com/xpublish-community/xpublish-intake) to v2. It uses the zarr to get into the intake catalog. I have a constructed urlpath from xpublish to put into my catalog like `http://0.0.0.0:9005/datasets/air/zarr` and it is being interpreted incorrectly in this code:

https://github.com/intake/intake/blob/2abbfad05bd9e9b372f1937895834ab226785806/intake/readers/readers.py#L1021-L1024

such that it will not be read in correctly. I think some additional logic needs to be put into place to separate out this sort of situation so that it does not assume a DAP endpoint, but I'm not sure how to differentiate properly. Thanks for any help!

cc @abkfenris

PR #810, opened 2024-05-23
differentiate http reader for zarr via xpublish-intake

kthyng:
This prevents urlpaths like `http://0.0.0.0:9005/datasets/air/zarr` that are being served by `xpublish-intake` from being served by the DAP end point and instead being served by the input zarr engine information.

@abkfenris Figured out the necessary code change.

Fixes https://github.com/intake/intake/issues/809

### DIFF
--- a/intake/readers/readers.py
+++ b/intake/readers/readers.py
@@ -1021,7 +1021,7 @@ class XArrayDatasetReader(FileReader):
             if (
                 isinstance(data, datatypes.FileData)
                 and data.url.startswith("http")
-                and not datatypes.Zarr
+                and not isinstance(data, datatypes.Zarr)
             ):
                 # special case, because xarray would assume a DAP endpoint
                 f = fsspec.open(data.url, **(data.storage_options or {})).open()
--- a/intake/readers/readers.py
+++ b/intake/readers/readers.py
@@ -1019,8 +1019,8 @@ class XArrayDatasetReader(FileReader):
         else:
             # TODO: recognise fsspec URLs, and optionally use open_local for engines tha need it
             if (
-                isinstance(data, datatypes.FileData) 
-                and data.url.startswith("http") 
+                isinstance(data, datatypes.FileData)
+                and data.url.startswith("http")
                 and not datatypes.Zarr
             ):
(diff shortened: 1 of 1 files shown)


intake/readers/readers.py
```diff
@@ -1018,7 +1018,11 @@ def _read(self, data, **kw):
             return open_mfdataset(data.url, **kw)
         else:
             # TODO: recognise fsspec URLs, and optionally use open_local for engines tha need it
-            if isinstance(data, datatypes.FileData) and data.url.startswith("http"):
+            if (
+                isinstance(data, datatypes.FileData)
+                and data.url.startswith("http")
+                and not datatypes.Zarr
```
martindurant, line 1024: I think:
```suggestion
                and not isinstance(data, datatypes.Zarr)
```

`not datatypes.Zarr` is always `False`, so you would never hit the DAP endpoint.

kthyng: Sorry about that, I should have stopped to make a test. I am working on that now.

kthyng: Well, it is hard to make a test without sample data.

kthyng: I made your change and have the start of a test but don't have any good zarr vs opendap example data to use to differentiate between the two http cases to use. Do you have any suggestions?

martindurant: A local server perhaps? You can then write some data to a temporary directory.

martindurant:
If you can't easily think of a way to test, I'll merge this in anyway, since we know it works for you. The whole of Intake 2 needs extensive testing...

kthyng:
I know the zarr/xpublish side works. Do you have an opendap intake catalog handy to make sure that still works?

martindurant:
Perhaps from https://github.com/intake/intake-xarray/blob/master/intake_xarray/tests/test_network.py ?

kthyng:
Yes, ok I have verified both types of inputs. I think it is good to go. ✅

intake/readers/readers.py
abkfenris: My sneaky attempt to move more folks away from OpenDAP has been foiled.

PR #815, opened 2024-05-31
added alias test

kthyng:


martindurant:
This test is useful, and we can further figure out when the aliases get updated - so long as we agree that those are indeed the names that are exposed via `list`.

PR #817, opened 2024-05-31
small change to alias test

kthyng:
wanted to add a reader modification to add a hash to the mix and make sure that hash does not make it into the aliases nor list(cat).


kthyng:
@martindurant Sorry I should have thought of this in the first place.

PR #821, opened 2024-06-05
Raise when trying to store a dynamic function

martindurant:
cc @kthyng

### DIFF
--- a/intake/readers/tests/test_errors.py
+++ b/intake/readers/tests/test_errors.py
@@ -1,0 +1,18 @@
+import pytest
+
+import intake
+
+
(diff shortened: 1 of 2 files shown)


kthyng:
I tested the code from this PR with a catalog I am working with that uses a `parse_dates` function in a catalog. Should this pass the test? It did for me.

martindurant:
It depends on what that function is. Normal functions defined in modules should be fine. Is your function the one that failed to import last time?

kthyng:
Yes, or similar. I now have utils functions defined (outside the catalogs) to account for this issue.

PR #831, opened 2024-06-24
Rename the zarr source back to ndzarr to avoid conflict with intake-x…

rafa-guedes:
The new `zarr` alias for `intake.source.zarr.ZarrArraySource` conflicts with the [intake-xarray one](https://github.com/intake/intake-xarray/blob/master/setup.py#L29). This pull request changes it back to `ndzarr` as defined in intake<2.


rafa-guedes:
Closes https://github.com/intake/intake-xarray/issues/144

PR #833, opened 2024-06-26
revise LlamaServerReader

AlbertDeFusco:
* safetensors requires a transformation step, support removed for now
* transformers is not required to run llama.cpp
* extend check_imports to search for llama.cpp executable
* set default port and host to avoid parsing from output
* add a health check

### DIFF
--- a/intake/readers/convert.py
+++ b/intake/readers/convert.py
@@ -5,11 +5,13 @@ By convention, functions here do not change the data, just how it is held.
 from __future__ import annotations
 
 import re
+from urllib.parse import urljoin
 from functools import lru_cache
 from itertools import chain
 
 from intake import import_name, conf
-from intake.readers import BaseData, BaseReader, readers
+from intake.readers.datatypes import OpenAIService
+from intake.readers import BaseData, BaseReader, readers, LlamaServerReader, OpenAIReader
 from intake.readers.utils import all_to_one, subclasses, safe_dict
 
 
@@ -412,6 +414,31 @@ class DataFrameToMetadata(BaseConverter):
             out["shape"] = [x.count(), len(x.columns)]
             out["schema"] = safe_dict(x.schema)
         return safe_dict(out)
+
+
+class GGUFToLlamaCPPService(BaseConverter):
+    instances = {"intake.readers.datatypes:GGUF": "intake.readers.datatypes:LlamaCPPService"}
+
+    def run(self, x, **kwargs):
+        return LlamaServerReader(x).read(**kwargs)
+
+
+class LLamaCPPServiceToOpenAIService(BaseConverter):
+    instances = {
+        "intake.readers.datatypes:LlamaCPPService": "intake.readers.datatypes:OpenAIService"
+    }
+
+    def run(self, x, options=None):
+        url = urljoin(x.url, "/v1")
+        service = OpenAIService(url=url, key="none", options=options)
+        return service
+
+
+class OpenAIServiceToOpenAIClient(BaseConverter):
+    instance = {"intake.readers.datatypes:OpenAIService": "openai:OpenAI"}
+
+    def run(self, x):
+        return OpenAIReader(x).read()
 
 
 def convert_class(data, out_type: str):
--- a/intake/readers/convert.py
+++ b/intake/readers/convert.py
@@ -435,7 +435,7 @@ class LLamaCPPServiceToOpenAIService(BaseConverter):
 
 
 class OpenAIServiceToOpenAIClient(BaseConverter):
-    instance = {"intake.readers.datatypes:OpenAIService": "openai:OpenAI"}
+    instances = {"intake.readers.datatypes:OpenAIService": "openai:OpenAI"}
 
     def run(self, x):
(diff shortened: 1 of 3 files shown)


martindurant:
Thanks for the changes. Are you intending to use this in something interesting?

intake/readers/readers.py
```diff
@@ -514,16 +514,39 @@ class LlamaServerReader(BaseReader):
...
-    imports = {"transformers"}
+    implements = {datatypes.GGUF}
+    imports = {"requests"}
+
+    @classmethod
+    def _find_executable(cls):
+        import shutil
+
+        # executables were renamed in https://github.com/ggerganov/llama.cpp/pull/7809
+        path = shutil.which("llama-server")
```
martindurant, line 525: I'm glad to see it's not "server" anymore!

AlbertDeFusco:
I like the idea of Intake being able to launch services and I like I can create a single catalog for data and models that my project depends on.

martindurant:
Sorry, I didn't notice that this PR became un-draft. It is now on my list.

AlbertDeFusco:
Do you require tests? I think I can come up with some appropriate mocks.

intake/readers/readers.py
```diff
@@ -513,36 +513,117 @@ def _read(self, name, **kw):
...
 
-    output_instance = "intake.readers.datatypes:ChatService"
-    implements = {datatypes.GGUF, datatypes.SafeTensors}
-    imports = {"transformers"}
+    output_instance = "intake.readers.datatypes:LlamaCPPService"
+    implements = {datatypes.GGUF}
+    imports = {"requests"}
+
+    @classmethod
+    def _find_executable(cls):
```
martindurant, line 521: Should we `@cache` this?

intake/readers/readers.py
```diff
@@ -513,36 +513,117 @@ def _read(self, name, **kw):
...
+        storage_options = {} if data.storage_options is None else data.storage_options
+        cache_location = os.path.join(user_data_dir(), "llama.cpp")
+        options = {protocol: storage_options, "simplecache": {"cache_storage": cache_location}}
+        fs, path = fsspec.url_to_fs(f"simplecache::{data.url}", **options)
+
+        cached_fn = fs._check_file(path)
+        if cached_fn:
+            return cached_fn
+
+        with fs.fs.open(path) as f:
```
martindurant, line 563: TODO: fsspec caching filesystems should allow the setting of a callback for when they download. I believe this was suggested in the fsspec repo.

martindurant: Was there a reason to use `open/read` rather than `get()`

intake/readers/readers.py
```diff
@@ -513,36 +513,117 @@ def _read(self, name, **kw):
...
+            finally:
+                cb.close()
+                cached.close()
 
-    def _read(self, data, log_file="out.log", **kwargs):
-        # TODO: list common options, like PORT
+            return cached_fn
+
+    def _read(self, data, log_file="llama-cpp.log", **kwargs):
+        startup_timeout = kwargs.pop("startup_timeout", 60)
```
martindurant, line 583: List these kwargs in the signature, so the defaults show up in documentation?

martindurant (commented):
(Posting now, since I noticed changes were pushed)

intake/readers/readers.py
AlbertDeFusco: What I've found is that while cachingfs methods supported a callback it was only activated when reading the file from the cache and not forwarded to the target fs while populating the cache. I'll get some time to make some PRs.

I also noticed some oddities with .get() where it would deadlock that I need to follow through on.

intake/readers/readers.py
AlbertDeFusco: done

intake/readers/readers.py
AlbertDeFusco: done

intake/readers/convert.py
```diff
@@ -414,6 +416,31 @@ def run(self, x, *args, **kwargs):
         return safe_dict(out)
 
 
+class GGUFToLlamaCPPService(BaseConverter):
+    instances = {"intake.readers.datatypes:GGUF": "intake.readers.datatypes:LlamaCPPService"}
+
+    def run(self, x, **kwargs):
+        return LlamaServerReader(x).read(**kwargs)
```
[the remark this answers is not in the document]
AlbertDeFusco: I will give it some more thought.

intake/readers/readers.py
martindurant: I thought I had replied to this: probably fsspec is trying to list directories to find matches for the "?" in the URL, which looks like a glob wildcard. This can be suppressed by passing `get([path])`.

AlbertDeFusco: `.get([path], [output_path], callback=...)` does avoid the deadlock, but the callback is on the list of files rather than a progress bar per file. I'll go with `.get_file()` for this case.

AlbertDeFusco: At least for callbacks that implement `.branch()`

PR #838, opened 2024-07-11
double dash llama-server kwargs

AlbertDeFusco:


martindurant:
There are both "-" (for common ones) and "--" arguments - is there any way to tell the difference?

AlbertDeFusco:
I don't have a way to distinguish them. I will add a note in the docs that long-form names are supported as kwargs. I'll check the llama-server help one more time to see if some are single-dash only.

martindurant:
We could maybe maintain an explicit list of short forms to check for, without promising to match all possible future single dash args.

AlbertDeFusco:
That's a good plan.

martindurant:
I'll merge this anyway now, so that you can make a demo :)

PR #857, opened 2024-12-20
Fix handling of environment variables of type string

### LINKED ISSUE #855: Unbound local error when setting cache path with environment variables
Hi there,
I've been experiencing an error when providing intake with a local cache path using environment variables.
I'm using python 3.12.8 and intake 2.0.7

## Steps to reproduce

- set an environment variable (using bash on linux)

```bash
export INTAKE_CACHE="icache"
```
- start a python console and import intake

I would expect intake to store files in a folder called `icache` in the current working directory. However, I get this error:
>   File "/home/jroettenbacher/miniforge3/envs/update-requirements/lib/python3.12/site-packages/intake/config.py", line 162, in load_env
    self[k2] = val
               ^^^
UnboundLocalError: cannot access local variable 'val' where it is not associated with a value

As far as I can tell the problem arises as `ast.literal_eval()`, which loads the environment variables, does not like string input. Except it wants everything to be quoted even strings. So `'"icache"'` would work and would create a folder called `"icache"`, which I don't want. Linux can do that but Windows throws an error because of this.

The [load_env()](https://github.com/intake/intake/blob/69c6565e8bed77a61f7e71e697053cea9ca5fb4d/intake/config.py#L152) function does not handle any exception or error thrown by `ast.literal_eval()` and instead tries to assign a not defined variable to the config dictionary, which causes the error.

It's possible to work around this by not working with local environment variables but I think this should be something one can do. Or at least could do with version 1?

I'm open to work on this problem but would love some insight and guidance on where to start. Or if I'm using local environment variables the wrong way?

Best Johannes

jroettenbacher:
When local environment variables of the form `INTAKE_*="any-string"` are present on import, the load_env() function would return an `UnboundLocalError`.
This fixes that.

Fixes #855 

We could think about not warning the user about this behaviour. What do you think?
Best Johannes

### DIFF
--- a/intake/config.py
+++ b/intake/config.py
@@ -35,6 +35,7 @@ defaults = {
     "extra_imports": [],
     "import_block_list": [],
     "reader_avoid": [],
+    "raise_on_error": False,
 }
 
 
@@ -157,8 +158,18 @@ class Config(dict):
                 k2 = k[7:].lower()
                 try:
                     val = ast.literal_eval(v)
-                except ValueError:
-                    pass
+                except (ValueError, SyntaxError) as e:
+                    if self["raise_on_error"]:
+                        logger.info(
+                            f"Environment variable '{v}' cannot be parsed by ast.literal_eval()"
+                        )
+                        raise e
+                    logger.warning(
+                        f"Failed to parse environment variable '{k}' (value: '{v}') using ast.literal_eval()."
+                        f"\nParsing as string."
+                        f"\nError was: '{type(e).__name__}: {e}'"
+                    )
+                    val = str(v)  # make sure it is a string
                 self[k2] = val
 
 
--- a/intake/config.py
+++ b/intake/config.py
@@ -48,6 +48,14 @@ class Config(dict):
     """Intake's dict-like config system
 
     Instance ``intake.conf`` is globally used throughout the package
+
+    Attributes:
+        environment_conf_parse : str
+                                 "ignore" (default), "warn" or raise an "error"
+                                 when parsing local environment variables as strings.
+        #TODO: Add other config values
+
+
     """
 
     def __init__(self, filename=None, **kwargs):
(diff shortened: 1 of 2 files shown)


intake/config.py
```diff
@@ -35,6 +35,7 @@
     "extra_imports": [],
     "import_block_list": [],
     "reader_avoid": [],
+    "raise_on_error": False,
```
martindurant, line 38: environment_conf_parse_error with possible values "raise", "warn", "ignore" with the latter as default?

intake/config.py
```diff
@@ -157,8 +158,18 @@ def load_env(self):
...
+                        logger.info(
+                            f"Environment variable '{v}' cannot be parsed by ast.literal_eval()"
+                        )
+                        raise e
+                    logger.warning(
+                        f"Failed to parse environment variable '{k}' (value: '{v}') using ast.literal_eval()."
+                        f"\nParsing as string."
+                        f"\nError was: '{type(e).__name__}: {e}'"
+                    )
+                    val = str(v)  # make sure it is a string
```
martindurant, line 172: https://docs.python.org/3/library/os.html#os.environ says they must always be strings

martindurant:
How do these changes look to you? Makes sense?

I guess https://intake.readthedocs.io/en/latest/api2.html#intake.config.Config or elsewhere in the docs should explicitly specify what each config key does, and what values are allowed, and there should also be validation somewhere. Without pydantic...

jroettenbacher:
Hi, the changes make everything more clear. Thanks for that. I would add the default config values and their explanation to the [Config class](https://github.com/intake/intake/blob/69c6565e8bed77a61f7e71e697053cea9ca5fb4d/intake/config.py#L48).

intake/config.py
jroettenbacher: So should we remove the `str( )` around `v` ?

martindurant:
You are right, documenting right in the Config class makes a lot of sense.

PR #858, opened 2025-01-09
New UP type: named options

martindurant:
This is a useful multi-value UP type; but it does not solve the case of "I want the reader to choose between multiple possible data inputs", which probably requires changing the order in which arguments get processed.


martindurant:
I need to consider how best to auto-populate such instances via extract_parameters

PR #860, opened 2025-01-16
Enable PDF and EPUB formats on readthedocs

### LINKED ISSUE #851: can we get a read the docs PDF?
Often a read the docs page has an option in the lower right to download a PDF version of the docs. This is helpful to drop into an LLM assistant and then have the assistant tell me how to use an unfamiliar tool. I don't see that option at https://intake.readthedocs.io/en/latest/

Can the PDF be made available on the read the docs page?

https://docs.readthedocs.io/en/stable/downloadable-documentation.html

danyeaw:
Fixes #851 by enabling PDF and EPUB formats on RTD. Docs on this are also here:
https://docs.readthedocs.io/en/stable/guides/enable-offline-formats.html

They will appear on the RTD flyout in the bottom right hand corner.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,5 +1,7 @@
 version: 2
 
+formats: all
+
 build:
   os: ubuntu-22.04
   tools:


danyeaw:
Thanks @martindurant!

PR #861, opened 2025-01-21
Implement catalog mapper

### LINKED ISSUE #854: Loading into sql db catalog entry
I read on the SQLAlchemy catalog reader that the entries from the resulting catalog could be loaded by another reader implementing SQLQuery. What’s an example of this & how does that work? I’m asking because I would like to implement a DuckDB catalog reader that makes entries from the information_schema.tables & write to those tables with the results of other readers.

martindurant:
Also adds "PandasToDuck" converter, and allows reading from Duck via arrow rather than pandas objects.

Fixes #854 

@dprovder, usage:
```
import intake
cat = intake.Catalog()
cat["pand"] = intake.readers.datatypes.Parquet("out.parquet").to_reader(outtype="pandas:DataFrame")
mapper = intake.readers.transform.CatalogMapper(func="PandasToDuck", transform=True)
cat2 = mapper.run(cat, conn="out.duckdb", name_arg="table", read=True)
```
Explanations:
- transform=True implies that the the readers in the catalog already know about a transform called PandasToDuck, which will be true for any pandas reader
- calling mapper.run works if you already have a Catalog. If instead you had a reader producing a catalog (from elsewhere), you would do `reader.CatalogMapper(...)
- for the meaning of name_arg and read, see the new docstring

Having executed the above, you will have a new catalog with the one entry for a table in duck with the same name as in the original catalog, in this case "pand".

### DIFF
--- a/intake/readers/convert.py
+++ b/intake/readers/convert.py
@@ -89,8 +89,44 @@ class DuckToPandas(BaseConverter):
     instances = {"duckdb:DuckDBPyRelation": "pandas:DataFrame"}
     func = "duckdb:DuckDBPyConnection.df"
 
-    def run(self, x, *args, **kwargs):
-        return x.df()
+    def run(self, x, *args, with_arrow=True, **kwargs):
+        if with_arrow:
+            import pandas as pd
+
+            table = x.to_arrow_table()
+            data = {
+                col: pd.Series(pd.arrays.ArrowExtensionArray(val))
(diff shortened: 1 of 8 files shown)


dprovder:
I'm curious why you didn't extend the catalog class? so that we could do `cat.map(PandasToDuck)`
going to data day this weekend, I'll try to see if I could show what I mean in an an implementation

PR #866, opened 2025-03-05
Add test for cat_mapper

martindurant:


intake/readers/transform.py
dprovder, line 1: This looks good, removing the init & dependencies on func. Cleaner than my solution. I tried the whole func defaults to identity thing but couldn’t work that out

intake/readers/tests/test_workflows.py
dprovder, line 1: What’s the difference between reader.transform.CatalogMapper & just calling it on the reader directly?

dprovder (approved):
CatalogMapper should work in pipelines with these changes

martindurant:
Copying comment:
> The idea was to either save the parameters for the reader or the token in the table’s comments so as to identify changes when writing to duck from a catalog

referring to [this diff](https://github.com/martindurant/intake/pull/4/files#diff-14e51905c1509e385d0dc5ad93af28cab4726981b131da011ac418379d48bba1). This is separate functionality, should probably be a different PR.

PR #880, opened 2025-11-10
Use fsspec to open files from remote URL with glob for XArrayDatasetReader

### LINKED ISSUE #879: Opening xr.Dataset from remote url with glob not working
I'm trying to open some remote netCDF files that are in a tar archive, so I've set this up as follows:

```
from intake.readers import datatypes, XArrayDatasetReader

data_url = "tar://*.nc::github://openghg:example_data@/footprint/tac_footprint_inert_201607.tar.gz"
data = datatypes.HDF5(data_url)

reader = XArrayDatasetReader(data)
reader.read()
```

This raises the error: 

```
ValueError: cannot do wild-card matching for paths that are remote URLs unless engine='zarr' is specified. Got paths: tar://*.nc::github://openghg:example_data@/footprint/tac_footprint_inert_201607.tar.gz. Instead, supply paths as an explicit list of strings.
```

This problem was fixed in PR #848, but I think the case of a glob string is handled incorrectly: since the url is a string with a `*` in it, we end up in the case on line 1301 of readers.py. But then the check `is_fsspec_url(data.url[0])` checks if the first character of the url is `://`, which is false, so the files aren't parsed.

I think the check should be
```
if (isinstance(data.url, str) and is_fsspec_url(data.url)) or is_fsspec_url(data.url[0]):
    ...
```

Incidentally,

```
of = fsspec.open_files(data_url)

with of as f:
    ds = open_mfdataset(f)
```
works, so it that works for the other cases, maybe that is an easier way to proceed.

I can open a PR to fix this if you like.

brendan-m-murphy:
Closes #879

Fix for issue #879. I added a test to trigger the xarray error (`open_mfdataset` can't accept remote URLs with globs), then added a fix to catch the case where the data URL passed to `XArrayDatasetReader` is a string with remote URL and a glob.

The test doesn't contain any assert statements, I suppose it could check that the retrieved data matches the input data.

A more general way of handling the case where the data's URL is a sequence or a string with a glob might be
```
of = fsspec.open_files(data.url, **(data.storage_options or {}))

with of as f:
    return open_mfdataset(f)
```
but before trying/implementing this, I'd want to create tests for all of the cases this would affect.

P.S. Sorry this got pushed here instead of my fork. For some reason my text editor's git interface pushes upstream even if I ask for origin, and I don't contribute from forks often enough to remember this. The force push is a rebase to fix up some comments on the test.

### DIFF
--- a/scripts/ci/environment-pip.yml
+++ b/scripts/ci/environment-pip.yml
@@ -27,6 +27,7 @@ dependencies:
       - bokeh
       - dask
       - h5netcdf
+      - h5py < 3.15  # due to HDF error, see netcdf4 Issue #1438
       - intake
       - netcdf4
       - pip


brendan-m-murphy:
I'll make a local pip environment to see if I get the same error. We've had some trouble with netcdf4 (and h5netcdf/h5py) on OpenGHG's pip CI recently.

PR #888, opened 2026-05-07
Fix #887 getshell warnings

### LINKED ISSUE #887: getshell config logic seems off
Following #883, we've been having a bunch of these warnings: 
```python
/g/data/xp65/public/apps/med_conda/envs/analysis3-26.02/lib/python3.11/site-packages/intake/catalog/utils.py:173: UserWarning: Shell command not executed due to getshell=False
  warnings.warn("Shell command not executed due to getshell=False")
```

I did some digging and I'm not sure this logic looks totally right:

```python
ipdb++> ll
    144 def expand_defaults(default, client=False, getenv=True, getshell=False):
    145     """Compile env, client_env, shell and client_shell commands
    146
    147     Execution rules:
    148     - env() and shell() execute where the cat is loaded, if getenv and getshell
    149       are True, respectively
    150     - client_env() and client_shell() execute only if client is True and
    151       getenv/getshell are also True.
    152
    153     If both getenv and getshell are False, this method does nothing.
    154
    155     If the environment variable is missing or the shell command fails, the
    156     output is an empty string.
    157     """
    158     breakpoint()
    159     r = re.match(r"env\((.*),?(.*)\)", default)
    160     if r and not client and getenv:
    161         gs = r.groups()
    162         default = os.environ.get(gs[0], gs[1] if len(gs) > 1 else "")
    163     r = re.match(r"client_env\((.*)\)", default)
    164     if r and client and getenv:
    165         default = os.environ.get(r.groups()[0], "")
    166     r = re.match(r"shell\((.*)\)", default)
    167     if r and not client and getshell:
    168         try:
    169             cmd = shlex.split(r.groups()[0])
    170             default = subprocess.check_output(cmd).rstrip().decode("utf8")
    171         except (subprocess.CalledProcessError, OSError):
    172             default = ""
    173     else:
--> 174         warnings.warn("Shell command not executed due to getshell=False")
    175     r = re.match(r"client_shell\((.*)\)", default)
    176     if r and client and getshell:
    177         try:
    178             cmd = shlex.split(r.groups()[0])
    179             default = subprocess.check_output(cmd).rstrip().decode("utf8")
    180         except (subprocess.CalledProcessError, OSError):
    181             default = ""
    182     else:
    183         warnings.warn("Shell command not executed due to getshell=False")
    184     return default
    185

ipdb++> r
ipdb++> client
False
ipdb++> getshell
True
```

Still digging - when I tested my changes, I get 54 test failures vs.... 54 on master, so I assume my pytest config is borked for unrelated reasons. 

I'll open a PR when I have a fix

charles-turner-1:
Should fail - fix coming in a secondary commit.

This is a super minimal test failure & fix - the rest of the function is ~8 years old so I'd like to keep the behaviour as close to how it was as possible. 

Closes #887

### DIFF
--- a/intake/catalog/tests/test_utils.py
+++ b/intake/catalog/tests/test_utils.py
@@ -4,6 +4,8 @@
 #
 # The full license is in the LICENSE file, distributed with this software.
 # -----------------------------------------------------------------------------
+import warnings
+
 import pandas as pd
 import pytest
 
@@ -72,3 +74,14 @@ def test_coerce(value, dtype, expected):
     out = utils.coerce(dtype, value)
     assert out == expected
     assert type(out) == type(expected)
+
+
+def test_expand_defaults():
+    """
+    expand_defaults should not warn that getshell=False when it is True.
+    """
+    with warnings.catch_warnings():
+        warnings.filterwarnings("error", category=UserWarning)
+        utils.expand_defaults(default="v2000-01-01", client=False, getenv=True, getshell=True)
+
+    assert True
--- a/intake/catalog/utils.py
+++ b/intake/catalog/utils.py
@@ -169,7 +169,7 @@ def expand_defaults(default, client=False, getenv=True, getshell=False):
             default = subprocess.check_output(cmd).rstrip().decode("utf8")
         except (subprocess.CalledProcessError, OSError):
             default = ""
-    else:
+    elif not getshell:
         warnings.warn("Shell command not executed due to getshell=False")
     r = re.match(r"client_shell\((.*)\)", default)
     if r and client and getshell:
@@ -178,7 +178,7 @@ def expand_defaults(default, client=False, getenv=True, getshell=False):
             default = subprocess.check_output(cmd).rstrip().decode("utf8")
         except (subprocess.CalledProcessError, OSError):
             default = ""
-    else:
+    elif not getshell:
         warnings.warn("Shell command not executed due to getshell=False")
     return default
 


charles-turner-1:
Sorry, I'm a bit confused with the CI here - it only seems to test readers?

Failure with [#27915f7](https://github.com/intake/intake/pull/888/commits/27915f7f5ae7c3d43a3eb5a2ea5c0cce1fe31d06):
___
```python
def test_expand_defaults():
    """
    expand_defaults should not warn that getshell=False when it is True.
    """
    with warnings.catch_warnings():
        warnings.filterwarnings("error", category=UserWarning)
        utils.expand_defaults(default="v2000-01-01", client=False, getenv=True, getshell=True)

    assert True
```

raises: `FAILED intake/catalog/tests/test_utils.py::test_expand_defaults - UserWarning: Shell command not executed due to getshell=False`

charles-turner-1:
Following https://github.com/intake/intake/pull/888/commits/38e516e9d67cc865ee6f2313ab038222a185c97e
___
```bash
$ pytest intake/catalog/tests/test_utils.py                                                                                                                            
==================================================================================================================================================== test session starts =====================================================================================================================================================
platform darwin -- Python 3.12.5, pytest-8.3.2, pluggy-1.5.0
rootdir: /Users/u1166368/catalog/intake
configfile: pyproject.toml
plugins: env-1.1.4, cov-5.0.0, anyio-4.9.0, html-4.1.1, metadata-3.1.1, mypy-0.10.3, flake-detection-0.1.0, zarr-3.1.3, mock-3.14.0, xdist-3.6.1
collected 22 items

intake/catalog/tests/test_utils.py ......................                                                                                                                                                                                                                                                              [100%]

===================================================================================================================================================== 22 passed in 0.64s
```

martindurant:
Sorry, I didn't notice the new code here. The failure appears unrelated.

d70-t:
Sorry for digging this out again, but I've the feeling that the behaviour of the warnings is still broken.

What I would expect is: I only get a warning if the code actually wanted to execute a shell command, but was stopped to do so, because of `getshell=False`.

Currently, if there's no match in the regex (`r == None`) and `getshell == False`, I get a warning, even though there wouldn't be any need for running a shell at all.

